PR #3, opened 2024-08-20
Add new CacheStrategy & CacheMiddleware classes

jblz:
* Add `POST` to the request methods that we _can_ cache, but default to not caching them
* Add a method for queries to declare a TTL via `get_cache_ttl` method
* Add debug logging to `fetch` & `cache`
* Add mechanism to bypass cache via `remote_data_blocks_bypass_cache` filter

This isn't perfect yet, but it gives us a jumping-off point for fine-tuning our caching strategy.

|Cold Cache|Warm Cache|
|---|---|
|Screenshot 2024-08-21 at 9 32 47 AM|Screenshot 2024-08-21 at 9 35 25 AM|


inc/http-client/class-cache-strategy.php
```diff
@@ -0,0 +1,57 @@
...
+	private static function getRequestString( RequestInterface $request ) {
+		$uri        = $request->getUri();
+		$uri_string = $uri->getScheme() . '://' . $uri->getHost() . $uri->getPath();
+		$method     = $request->getMethod();
+		$body       = $request->getBody()->getContents();
+		return $method . ' ' . $uri_string . ' ' . $body;
+	}
+
+	protected function should_bypass_cache( RequestInterface $request ) {
+		$bypass = $request->getHeader( self::HEADER_BYPASS_CACHE );
```
jblz: We can inspect the request here to e.g. put tighter scrutiny on `POST` requests.

inc/http-client/class-cache-strategy.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace RemoteDataBlocks\HttpClient;
+
+use Kevinrob\GuzzleCache\Strategy\GreedyCacheStrategy;
+use Psr\Http\Message\RequestInterface;
+use RemoteDataBlocks\Logging\LoggerManager;
+
+class CacheStrategy extends GreedyCacheStrategy {
```
mhsdef: Worth s/`CacheStrategy`/`CustomCacheStrategy`?

Just thinking, in other context, like when instantiated, some sort of qualifier like that might convey more meaning. And `RemoteDataBlocksCacheStrategy` feels a tad long even for me (though I'm open to it!).

inc/http-client/class-cache-middleware.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace RemoteDataBlocks\HttpClient;
+
+use Kevinrob\GuzzleCache\CacheMiddleware as GuzzleCacheCacheMiddleware;
+
+class CacheMiddleware extends GuzzleCacheCacheMiddleware {
```
mhsdef: Same question as [below](https://github.com/Automattic/remote-data-blocks/pull/3/files#r1725700885) here. What are your feels on `CustomCacheMiddleware`? 🤔

mhsdef (approved):
LGTM, didn't test.

It would be nice to have unit tests on the cache strategy class. Maybe take Cursor for a spin on that if you haven't tried it yet? I suspect it'd do nicely with a class that's so direct as that one.

inc/http-client/class-cache-strategy.php
```diff
@@ -0,0 +1,57 @@
...
+	const HEADER_TTL          = 'X-RemoteDataBlocks-TTL';
+	const HEADER_BYPASS_CACHE = 'X-RemoteDataBlocks-Bypass-Cache';
+
+	private static function getRequestString( RequestInterface $request ) {
+		$uri        = $request->getUri();
+		$uri_string = $uri->getScheme() . '://' . $uri->getHost() . $uri->getPath();
+		$method     = $request->getMethod();
+		$body       = $request->getBody()->getContents();
+		return $method . ' ' . $uri_string . ' ' . $body;
```
mehmoodak: Is it safe to log the content of the `$body`?

inc/http-client/class-cache-strategy.php
```diff
@@ -0,0 +1,57 @@
...
+		}
+
+		return false;
+	}
+
+	public function fetch( RequestInterface $request ) {
+		$logger = LoggerManager::instance();
+
+		if ( $this->should_bypass_cache( $request ) ) {
+			$logger->debug( 'Bypassing cache: ' . self::getRequestString( $request ) );
```
mehmoodak: Should we have any prefix in debugging messages which will make the search easier?

mehmoodak:
+1 on @mhsdef suggestion of testing.

inc/http-client/class-cache-strategy.php
jblz: I think the PHP namespace will suffice for now to indicate these are bespoke & not from the composer lib.  If / when we have more than one implementation of the strategy / middleware classes, that's probably a good time to be more specific about the naming, IMO.

inc/http-client/class-cache-strategy.php
jblz: > Is it safe to log the content of the $body?

In production, we'd not want to do this for sure. Since we're just logging to the debug log, it's probably not all that critical. At least for now as we debug the middleware and strategy.

inc/http-client/class-cache-strategy.php
jblz: e.g. We discussed inspecting the GraphQL operation for mutations and setting those to not cache.

Additionally, we discussed using a method on e.g. the `QueryContext` that we leverage to bypass the cache.

inc/http-client/class-cache-strategy.php
jblz: Good idea.

jblz (commented):
+1 to tests for these.  I wanted to make sure folks agreed w/ the direction before writing them.  I'll see how feasible it is.

inc/http-client/class-cache-strategy.php
chriszarate: Just pass a string to `LoggerManager::instance()`, it accepts a namespace

jblz: @chriszarate Seems like the `Logger` constructor / `create` fn does, but not the static `LoggerManager::instance` fn:

https://github.com/Automattic/remote-data-blocks/blob/01eaa598eb983ec9aa9e98a586aa0aa4864750f6/inc/logging/logger-manager/logger-manager.php#L27-L33

...unless I'm missing something?

jblz: Any reason to not just use the `Logger::create` fn here?

chriszarate: Oh, right. I guess we already have a namespace and don't really need one? Or we want multi-level namespaces?

jblz: I think @mehmoodak's suggestion was for multi-level namespaces to be able to more-easily identify that the debug log came from e.g. the HTTP Client module (similar to how the [debug npm package](https://www.npmjs.com/package/debug) allows multi-level prefixing / filtering.

We could drill in via Query Monitor like this:

Screenshot 2024-08-23 at 5 19 40 PM

I think I'll file it to an issue for an enhancement outside this PR.

jblz: https://github.com/Automattic/remote-data-blocks/issues/12

composer.json
```diff
@@ -14,7 +14,8 @@
     "galbar/jsonpath": "^3.0",
     "guzzlehttp/guzzle": "^7.8",
     "kevinrob/guzzle-cache-middleware": "^5.1",
-    "psr/log": "^3.0"
+    "psr/log": "^3.0",
+    "webonyx/graphql-php": "^15.12"
```
chriszarate: It would be nice to avoid shipping this big vendor dependency, seems like our usage is very limited

tests/inc/http-client/HttpClientTest.php
```diff
@@ -11,6 +11,8 @@
 use GuzzleHttp\Exception\ConnectException;
 use GuzzleHttp\Psr7\Request;
 use GuzzleHttp\Exception\RequestException;
+use Kevinrob\GuzzleCache\Storage\VolatileRuntimeStorage;
```
chriszarate: this `Kevinrob` person has some good adjectives, A++

composer.json
jblz: +1.  replaced w/ a much smaller lib that splits out its parser:

https://github.com/graphpql/graphpinator-parser / https://packagist.org/packages/infinityloop-dev/graphpinator-parser

Part of:
https://github.com/graphpql/graphpinator

jblz: It's listed here: https://graphql.org/community/tools-and-libraries/?tags=php

inc/http-client/class-cache-strategy.php
jblz: I ended up prefixing these with `Rdb` -- feel free to push a change if anyone has a preference.

inc/HttpClient/RdbCacheMiddleware.php
```diff
@@ -0,0 +1,14 @@
+<?php declare(strict_types = 1);
+
+namespace RemoteDataBlocks\HttpClient;
+
+class RdbCacheMiddleware extends \Kevinrob\GuzzleCache\CacheMiddleware {
+	/**
+	 * @var array<string, bool>
+	 */
+	// phpcs:ignore WordPress.NamingConventions.ValidVariableName.PropertyNotSnakeCase, SlevomatCodingStandard.TypeHints.PropertyTypeHint.MissingNativeTypeHint
```
jblz, line 9: These are needed to comply with the parent class' declaration

jblz:
@chriszarate want to give this another once over?

inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -82,7 +82,12 @@ protected function get_raw_response_data( array $input_variables ): array|WP_Err
 			RequestOptions::JSON    => $body,
 		];
 
-		$this->http_client->init( $endpoint_base );
+		$client_options = [
+			// This is a custom option ignored by GuzzleHttp\Client, so it is prefixed with '__'.
+			'__default_cache_ttl' => $this->query_context->get_cache_ttl( $input_variables ),
```
chriszarate: Would be nice to stick this on a class as a constant, maybe on `HttpClient` where it is used.

chriszarate (approved):
Looks great, just had one suggestion (to improve my own code) :)

PR #17, opened 2024-08-26
Airtable Integration: Extract AirtableDatasource & leverage it for built-in & dynamic datasources

jblz:
Much of this came in via #39, so I've limited it to abstracting the Airtable config.

The examples pull in a `AirtableDatasource` and provide a table "variation" (for lack of a better name -- any ideas to improve the API?) instead of hard-coding the endpoint.

This lets us use a similar approach to dynamically register blocks for Airtable sources saved in the db, etc.


inc/editor/configuration-loader/configuration-loader.php
```diff
@@ -5,9 +5,14 @@
 defined( 'ABSPATH' ) || exit();
 
 use Error;
+use RemoteDataBlocks\Config\AirtableDatasource;
+use RemoteDataBlocks\Config\HttpDatasourceConfig;
 use RemoteDataBlocks\Config\QueryContext;
+use RemoteDataBlocks\Config\ShopifyDatasource;
 use RemoteDataBlocks\Logging\Logger;
 use RemoteDataBlocks\Logging\LoggerManager;
+use RemoteDataBlocks\REST\DatasourceCRUD;
```
jblz: This class should probably get moved out of the `REST` namespace

inc/config/dynamic-datasource-trait.php
```diff
@@ -0,0 +1,24 @@
...
+	private string $slug;
+	private string $uuid;
+
+	public function set_slug( string $slug ): void {
+		$this->slug = $slug;
+	}
+
+	public function get_slug(): string {
+		return $this->slug;
+	}
```
mhsdef: What is the purpose of the slug?

cc @shekharnwagh

shekharnwagh: This was meant to be the predictable unique identifier, so that we can reference the credentials in example code using this, and then users could tag the data source in the Settings Page with this slug. So the slug would act as a link between those 2 systems. Earlier we had `uuid` as unique identifier, but that is not predictable until user adds the data source from Setting Page. Open to other ideas if they seem easier, this was just the first thought that popped into my mind.

inc/editor/configuration-loader/configuration-loader.php
mhsdef: Agree.

inc/config/dynamic-datasource-trait.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace RemoteDataBlocks\Config;
+
+trait DynamicDatasource {
```
mhsdef: How do we all feel about traits?

jblz: This was mainly a shortcut since these values didn't make much sense in an `HttpDatasource` instance.

In a world with only Single Inheritance, traits can be handy, but sometimes confusing.

inc/integrations/airtable-integration/airtable-integration.php
```diff
@@ -0,0 +1,44 @@
...
+		$data_sources = DatasourceCRUD::get_data_sources( REMOTE_DATA_BLOCKS_AIRTABLE_SERVICE );
+
+		foreach ( $data_sources as $config ) {
+			// Transform data to our experimental format, which is all array based
+			$config = array_map(
+				function ( $value ) {
+					return is_object( $value ) ? (array) $value : $value;
+				},
+				(array) $config
+			);
```
mhsdef, line 26: this was a hack when i was bumbling about in that unmerged try, but, at this point (hopefully) unnecessary.

(though I still think we should array everywhere personally!)

mhsdef (approved):
Tests out for me. 👍

PR #25, opened 2024-09-01
Add GitHub File as HTML block

mhsdef:
This PR adds a GitHub File as HTML block. The main query, for now, requires a file extension hardcoded in registration to determine the types of files to list.

The file that is selected is fetched from GitHub as HTML and inserted into the current page via a pattern that binds the source to a paragraph block. Not ideal semantically, but, the best we have there in binding at the moment. The result renders nicely in spite of the `<p>` wrapping.


example/github/remote-data-blocks/inc/queries/class-github-get-file-as-html-query.php
```diff
@@ -0,0 +1,112 @@
...
+				'path' => '$.url',
+				'type' => 'string',
+			],
+		],
+	];
+
+	public function get_endpoint( array $input_variables ): string {
+		return sprintf(
+			'https://api.github.com/repos/%s/%s/contents/%s?ref=%s',
+			$this->get_datasource()->repo_owner,
```
chriszarate, line 61: Shouldn't this be an input variable?

mhsdef: Sounds like yes.

I'll tweak. I need to get my head screwed on better in this part. The way inputs and outputs pipe between each other is still pretty magic to me.

PR #28, opened 2024-09-02
add google sheets data source in settings page

shekharnwagh:
# Description

This PR adds a new type of data source in Settings Page - Google Sheets. It uses [service accounts to authenticate](https://developers.google.com/identity/protocols/oauth2/service-account) and fetch data from Google Sheets. This is the first data source which uses OAuth2.0 to authenticate.

The flow for adding a Google Sheet data source is as follows:

- Go the Google Cloud Console and create a new project (or select an existing one)
- Enable the Google Sheets and Drive API (required to list spreadsheets) for the project
- Create a new service account for the project
- Add a new key to the service account and download the JSON key file
- Grab the email address of the service account and grant that email address access to the spreadsheet you want to use
- Go to the Settings Page for Remote Data Blocks, click on "Add Data Source" and select the "Google Sheets"
- Enter the downloaded JSON key file contents in the "Credentials" field
- The service account credentials are used to generate an access token which is used to fetch the list of spreadsheets which the service account has access to and the list of sheets in the selected spreadsheet
- Select the spreadsheet and sheet you want to use and click on "Save" which will save the data source

Currently the saved data sources configured in the Settings Page are not hooked up with the registered blocks.

A separate PR will be made to add an example blocks for Google Sheets data source.


maxschmeling (approved):
Works great!

Maybe a few additional code comments that could be useful, but we can improve on that as we continue working on this.

PR #29, opened 2024-09-04
Try expanding interface-driven datasource configs

mhsdef:
## Why

I'm uneasy with current trending of complexity in the PHP core.

## What

Try this experimental twist to see if we can beat some of the beast back!

## How

We already had a `HttpDatasourceConfig` interface in the codebase. This PR attempts to shift some more core parts that direction, elevate a compositional approach, and lays some groundwork for additional explicit contracts (ie, interfaces, mostly, in PHP). 

See this (contrived) example of what that can mean in practice:

[Inheritance embraced](https://gist.github.com/mhsdef/690973352a6e3b8398d86de59e263a60)

vs

[Composition, Interfaces ](https://gist.github.com/mhsdef/ded798ed4d9dce647875b8212c5637a0)

The wins there may look small but when multiplied over real world scenarios hundreds or thousands of times--I promise, they add up. The mental overhead wins and testability wins make it worth it in my experience.


inc/editor/configuration-loader/configuration-loader.php
```diff
@@ -251,33 +245,4 @@ public static function unregister_all(): void {
...
-					$datasource = new ShopifyDatasource( $_source->token, $_source->store );
-					$datasource->set_uuid( $_source->uuid );
-					$datasource->set_slug( $_source->slug );
-					self::register_blocks_for_shopify_data_source( $datasource );
-					break;
-				default:
-					break;
-			}
-		}
-	}
```
mhsdef, line 270: I think we'll probably want to do this again, with full intent, for all the services we provide out of the box.

For now, I moved it back out to be with the rest of the code in `example` because the split locations were tough to mentally navigate.

example/airtable/elden-ring-map/register.php
```diff
@@ -2,7 +2,7 @@
 
 namespace RemoteDataBlocks\Example\Airtable\EldenRingMap;
 
-use RemoteDataBlocks\Config\AirtableDatasource;
+use RemoteDataBlocks\Config\HttpDatasource;
```
mhsdef: We can move back to concrete datasources per service (that implement the `HttpDatasourceInterface`, natch!) if we find we need to; but, so far configs + a generic `HttpDatasource` work for all cases.

inc/config/http-datasource/http-datasource.php
```diff
@@ -1,57 +1,77 @@
...
+	private array $config;
 
-defined( 'ABSPATH' ) || exit();
+	/**
+	 * Constructor for the HttpDatasource.
+	 *
+	 * @param array $config The configuration object for this HTTP datasource.
+	 */
+	public function __construct( array $config ) {
+		$this->config = $config;
```
mhsdef: TODO: add validation

example/rest-api/art-institute/inc/queries/class-art-institute-datasource.php
```diff
@@ -1,17 +0,0 @@
-<?php
-
-namespace RemoteDataBlocks\Example\ArtInstituteOfChicago;
-
-use RemoteDataBlocks\Config\HttpDatasource;
-
-class ArtInstituteOfChicagoDatasource extends HttpDatasource {
```
chriszarate, line 7: These examples are now drifting from simple exercises that help the implementor. This feels like a step backwards for example code. Or do we expect users to emulate our patterns of dynamic datasources?

example/airtable/elden-ring-map/inc/queries/class-airtable-elden-ring-list-locations-query.php
```diff
@@ -15,22 +15,27 @@ class AirtableEldenRingListLocationsQuery extends QueryContext {
...
-			'title' => [
+			'map_name' => [
 				'name' => 'Name',
 				'path' => '$.fields.Name',
 				'type' => 'string',
 			],
-			'x'     => [
+			'title'    => [
+				'name' => 'Name',
+				'path' => '$.fields.Name',
```
chriszarate: Duplicate?

example/rest-api/art-institute/inc/queries/class-art-institute-datasource.php
chriszarate: (This comment is more directed at the base PR, which mixes dynamic data sources into example code)

example/airtable/elden-ring-map/register.php
```diff
@@ -15,21 +15,35 @@
...
 		$logger->warning( sprintf( '%s is not defined, cannot register %s block', 'EXAMPLE_AIRTABLE_ELDEN_RING_ACCESS_TOKEN', $block_name ) );
 		return;
 	}
 
-	$elden_ring_datasource = new AirtableDatasource( $access_token, $base, $tables );
-	$list_locations_query  = new AirtableEldenRingListLocationsQuery( $elden_ring_datasource );
-	$list_maps_query       = new AirtableEldenRingListMapsQuery( $elden_ring_datasource );
+	$locations_config = [
+		'friendly_name'   => $block_name . ' (Locations)',
+		'uid'             => 'appqI3sJ9R2NcML8Y/tblc82R9msH4Yh6ZX',
```
chriszarate: I'm confused about `uid` vs `uuid` vs `slug`. Is `uid` used?

example/airtable/elden-ring-map/register.php
```diff
@@ -15,21 +15,35 @@
...
 		$logger = LoggerManager::instance();
 		$logger->warning( sprintf( '%s is not defined, cannot register %s block', 'EXAMPLE_AIRTABLE_ELDEN_RING_ACCESS_TOKEN', $block_name ) );
 		return;
 	}
 
-	$elden_ring_datasource = new AirtableDatasource( $access_token, $base, $tables );
-	$list_locations_query  = new AirtableEldenRingListLocationsQuery( $elden_ring_datasource );
-	$list_maps_query       = new AirtableEldenRingListMapsQuery( $elden_ring_datasource );
+	$locations_config = [
+		'friendly_name'   => $block_name . ' (Locations)',
```
chriszarate: Do we have a "regular name" that this is distinguishable from? "Display name"?

example/airtable/elden-ring-map/register.php
```diff
@@ -15,21 +15,35 @@
...
+			'Authorization' => "Bearer {$access_token}",
+			'Content-Type'  => 'application/json',
+		],
+	];
+
+	$maps_config = [
+		'friendly_name'   => $block_name . ' (Maps)',
+		'uid'             => 'appqI3sJ9R2NcML8Y/tblS3OYo8tZOg04CP',
+		'endpoint'        => 'https://api.airtable.com/v0/appqI3sJ9R2NcML8Y/tblS3OYo8tZOg04CP',
+		'request_headers' => [
```
chriszarate: I like the config approach. But our agreement was that v1 would be code-based and, after learning from additional use cases, we would tackle a more static config approach. I worry that we don't have enough examples under our belt to understand what limitations this brings, and we will bifurcate the happy path.

inc/config/http-datasource/http-datasource-interface.php
```diff
@@ -0,0 +1,85 @@
...
+
+	/**
+	 * Constructor for the HttpDatasource.
+	 *
+	 * @param array $config The configuration object for this HTTP datasource.
+	 */
+	public function __construct( array $config );
+
+	/**
+	 * Get a human-readable name for this datasource.
```
chriszarate: I'm torn about where these comments are most useful, in the interface or the implementation? Where are users more likely to look? Maybe both?

inc/config/http-datasource/http-datasource-interface.php
```diff
@@ -0,0 +1,85 @@
...
+	 * that can be used in user interfaces or for identification purposes.
+	 *
+	 * @return string The friendly name of the datasource.
+	 */
+	public function get_friendly_name(): string;
+
+	/**
+	 * Get a unique identifier for this datasource.
+	 *
+	 * This method should return a sha256 hash of a globally unique identifier
```
chriszarate: Doesn't seem like this returns a hash

inc/config/http-datasource/http-datasource-interface.php
```diff
@@ -0,0 +1,85 @@
...
+	public function get_endpoint(): string;
+
+	/**
+	 * Get the request headers. Override this method to provide authorization or
+	 * other custom request headers. Note that the query configuration can override
+	 * or extend these headers at request time.
+	 *
+	 * @return array Associative array of request headers.
+	 */
+	public function get_request_headers(): array;
```
chriszarate, line 45: This is probably where I'm most concerned about a config based approach. Lots of APIs require runtime-generated tokens.

inc/config/http-datasource/http-datasource.php
chriszarate: Just to state the alternate view, a code-based approach can benefit from simple type safety of interfaces.

inc/editor/configuration-loader/configuration-loader.php
chriszarate: Agree that it doesn't belong with example code, almost a blocker IMO

example/airtable/elden-ring-map/inc/queries/class-airtable-elden-ring-list-locations-query.php
mhsdef: We need to talk about this one soon.

example/airtable/elden-ring-map/register.php
mhsdef: Ignore the payload atm. This is a quick experiment on top of the other PR so the fields are quite malleable to say the least. I just wanted to get it up to discuss while keeping the code working.

example/airtable/elden-ring-map/register.php
mhsdef: sure, "Display name" is better.

inc/config/http-datasource/http-datasource-interface.php
mhsdef: it did til a last minute change (i realized it should be on write).

will update

inc/config/http-datasource/http-datasource-interface.php
mhsdef: How is this significantly different than before? 

In our concrete class (`HttpDatasource`), we do implement a default now; but it can still be overridden.

chriszarate: Agreed, it's not different. Just wondering aloud if a config-based approach can address cases where dynamism is needed.

example/airtable/elden-ring-map/register.php
mhsdef: FWIW, on another pass, I would take it this direction and probably stop here.

_Not_ pure json.

In the quest for data isolation from logic, I forgot about the perf implications. It's rough in PHP to spend so much time in json (or yaml) file loading on every request. So, associative arrays, in PHP, to leverage opcache. And we retain that executable flexibility.

mhsdef:
closing in favor of #39

PR #31, opened 2024-09-09
Add example child theme with block styling

chriszarate:
This PR adds an example child theme that provides styling of some of the example remote data blocks, primarily via `theme.json`. Here are the main improvements:

- Add example child theme of `twentytwentyfour`, with `theme.json` and `style.css` overrides.
- Improve label markup to use block settings / CSS variables and allow style overrides.
- Add additional block supports to allow users to add style overrides via the block editor UI.
- Add class names to bound blocks as targets for style overrides, e.g., `rdb-block-data-price` for a block bound to the `price` field.

### Airtable Event

Screenshot 2024-09-09 at 17 49 00

### Shopify Product

Screenshot 2024-09-09 at 17 48 23


.wp-env.json
```diff
@@ -10,5 +10,6 @@
     "./example",
     "https://downloads.wordpress.org/plugin/query-monitor.3.16.4.zip",
     "https://downloads.wordpress.org/plugin/redis-cache.2.5.3.zip"
-  ]
+  ],
+  "themes": [ "./example/theme" ]
```
mehmoodak, line 14: 1. Should we put this in `.wp-env.override.json` file?
2. Should we create a sample override file with name `.wp-env.override.sample.json` for easy overriding?

```json
{
  "themes": [ "./example/theme" ],
  "config": {
    "REMOTE_DATA_BLOCKS_EXAMPLE_AIRTABLE_ELDEN_RING_ACCESS_TOKEN": "",
    "REMOTE_DATA_BLOCKS_EXAMPLE_AIRTABLE_EVENTS_ACCESS_TOKEN": "",
    "REMOTE_DATA_BLOCKS_EXAMPLE_SHOPIFY_ACCESS_TOKEN": ""
  }
}
```

example/theme/style.css
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * Therefore, a Remote Data Block named "Shopify Product" containing a paragraph
+ * block bound to a field named `description` can be targeted with a selector:
+ *
+ * .wp-block-remote-data-blocks-shopify-product p.rdb-block-data-description {
+ *   /* styles here * /
+ * }
+ *
+ * However, note that these styles do not get applied in the editor. For simple
+ * styles, using `theme.json` and the `css` property is recommended.
```
mehmoodak: Should we use [enqueue_block_editor_assets](https://developer.wordpress.org/reference/hooks/enqueue_block_editor_assets/) action hook and have a separate `editor.css` file to apply styles on editor also?

Or maybe we can have a common styles file that targets both frontend and editor?

example/theme/theme.json
```diff
@@ -0,0 +1,106 @@
+{
+  "version": 3,
```
mehmoodak, line 3:
```suggestion
  "$schema": "https://schemas.wp.org/trunk/theme.json",
  "version": 3,
```

src/blocks/remote-data-container/block.json
```diff
@@ -12,8 +12,26 @@
     "remote-data-blocks/remoteData": "remoteData"
   },
   "supports": {
+    "background": {
+      "backgroundImage": true,
+      "backgroundSize": true
+    },
+    "className": true,
```
mehmoodak: Default value is already `true`.

https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#classname

src/blocks/remote-data-container/block.json
```diff
@@ -12,8 +12,26 @@
...
+    "color": {
+      "background": true
```
mehmoodak: Default value is already `true`

https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#color

src/utils/string.ts
```diff
@@ -1,5 +1,18 @@
-export const toTitleCase = ( str: string ): string => {
+// Provide functions to consistently generate class names.
```
mehmoodak: Can we please have a JSDoc comment here which will be helpful in intellisense.

mehmoodak (approved):

.wp-env.json
chriszarate: I purposefully put it in the default `.wp-env.json`, I think we want it present for all developers. 

The second idea is a good one.

mehmoodak (approved):

PR #32, opened 2024-09-10
Update tsconfig target to ES6

mehmoodak:
The minimum supported WP version of the plugin is `6.6` which uses Gutenberg editor having ES6 configured as `target` version ([src](https://github.com/WordPress/gutenberg/blob/wp/6.6/packages/eslint-plugin/configs/esnext.js#L12)) so updating the tsconfig `target` version in the plugin.


mhsdef:
nice, good eye

PR #33, opened 2024-09-10
add westeros houses example using google sheets

shekharnwagh:
# Description

This PR adds an example block for Google Sheets. Google Sheets like Airtable and unlike Shopify doesn't have a static schema for the data. So this example is just one example and the schema would be different for each Google Sheet. The example used here is a sheet containing the list of houses of Westeros with the columns as "Name", "Seat", "Region", "Words" and "Sigil". The 2 blocks added are:

- "Westeros House" to add a single house
- "Westeros Houses List" to add all houses

# Testing

## Google Sheets and Google API Access Setup
- To test this block, create a Google Sheet with above columns with the "Sigil" column having the URL of the sigil image. 
- [Create a project](https://developers.google.com/workspace/guides/create-project) in Google Cloud Platform. `resourcemanager.projects.create` permission is needed to create a new project. Skip this step if you already have a project setup in your organization in Google Cloud Platform.
- [Enable the Google Drive API](https://console.cloud.google.com/apis/library/drive.googleapis.com) for your project.
- [Enable the Google Sheets API](https://console.cloud.google.com/apis/library/sheets.googleapis.com) for your project.
- [Create a service account](https://cloud.google.com/iam/docs/service-accounts-create) for your project.
- [Create a key](https://cloud.google.com/iam/docs/keys-create-delete) for the service account. This will download a JSON key file. Keep this file safe as it will be used to authenticate the block.
- Convert the downloaded keys JSON to a JSON string, Base64 encode it and add it to the `REMOTE_DATA_BLOCKS_EXAMPLE_GOOGLE_SHEETS_WESTEROS_HOUSES_ACCESS_TOKEN"` key in the `.wp-env.override.json` file.

## Trying out the block
- Start the local WordPress environment with `npm run monolith`
- Go to the WordPress admin dashboard at `http://localhost:8888/wp-admin`
- Create a new post and try adding the newly added "Westeros House" block to the post.
- You can also try adding the "Westeros Houses List" block to the post.

## Screenshots

| Screenshot 2024-09-13 at 4 09 39 PM |
| :---: |
| Single "Westeros House" block in Editor |

| Screenshot 2024-09-13 at 4 09 49 PM |
| :---: |
| "Westeros Houses List" block in WordPress Frontend |


example/google-sheets/westeros-houses/inc/queries/class-get-westeros-houses-query.php
```diff
@@ -0,0 +1,81 @@
...
+				],
+			],
+			'type'      => 'id',
+		],
+	];
+
+	public array $output_variables = [
+		'is_collection' => false,
+		'mappings'      => [
+			'rowId'     => [
```
chriszarate: maybe should be snake_case for consistency

shekharnwagh: fixed in 7f9547a

mhsdef (approved):
Cool stuff! Didn't test, but, code makes sense. :shipit:

chriszarate (approved):
Tested, looks good. But I am concerned about all the fantasy content in these examples 😂

inc/integrations/google/auth/google-auth/google-auth.php
```diff
@@ -47,12 +64,38 @@ public static function generate_token_from_service_account_key(
...
-		return self::get_token_using_jwt( $jwt, $token_uri );
+		$token = self::get_token_using_jwt( $jwt, $token_uri );
+
+		if ( ! is_wp_error( $token ) ) {
+			wp_cache_set(
+				$cache_key,
+				$token,
+				'oauth-tokens',
+				3000, // 50 minutes
+			);
```
shekharnwagh, line 86: @chriszarate @mhsdef I am wondering if we should encrypt this token before writing to cache ? We do encrypt the data source config when we store it in options table, but that is persistent as opposed to this which has TTL of 50mins for now.

mhsdef: Ideally, yeah. We could do what we're doing with the DB option if there's a hook that fires in the cache set. I didn't see one, in brief perusing, but maybe I missed it.

Otherwise, could invoke our encryption code manually to handle.

PR #34, opened 2024-09-10
Add IGDB datasource example

maxschmeling:


example/rest-api/igdb/inc/queries/class-igdb-list-games-query.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace RemoteDataBlocks\Example\ArtInstituteOfChicago;
+
+use RemoteDataBlocks\Config\HttpDatasourceConfig;
+use RemoteDataBlocks\Config\QueryContext;
+
+class IGDBListGamesQuery extends QueryContext {
```
mhsdef, line 8: just a heads up, gonna need to swap this to `extends HttpQueryContext` now post #40

PR #36, opened 2024-09-10
Documentation pass numero 2

mhsdef:
Freshening things up a bit. Nothing in stone.


mhsdef:
these are mostly placehoders, relative to what i personally would want to read, quickly, as a dev.

needs moar pictures. (and better copy than this cruft)

PR #39, opened 2024-09-12
Expand interface-driven datasource configs

mhsdef:
## Why

It is not a silver bullet, but, programming to interfaces and favoring composition over inheritance can provide tighter abstractions, looser couplings, and useful flexibility.

See this (contrived) example of what that can mean in practice:

[Inheritance embraced](https://gist.github.com/mhsdef/690973352a6e3b8398d86de59e263a60)

vs

[Composition, Interfaces ](https://gist.github.com/mhsdef/ded798ed4d9dce647875b8212c5637a0)

The wins there may look small but when multiplied over real world scenarios hundreds or thousands of times--I promise, they add up. The mental overhead wins and testability wins make it worth it in my experience.

## How

We already had a `HttpDatasourceConfig` interface in the codebase. This PR attempts to shift some more core parts that direction, elevate a compositional approach, and lays some groundwork for additional explicit contracts (ie, interfaces, mostly, in PHP).


inc/integrations/airtable-integration/airtable-integration.php
```diff
@@ -0,0 +1,40 @@
...
+				function ( $value ) {
+					return is_object( $value ) ? (array) $value : $value;
+				},
+				(array) $config
+			);
+			self::register_blocks_for_shopify_data_source( $config );
+		}
+	}
+
+	private static function register_blocks_for_shopify_data_source( array $config ): void {
```
chriszarate:
```suggestion
	private static function register_blocks_for_airtable_data_source( array $config ): void {
```

inc/integrations/airtable-integration/airtable-integration.php
```diff
@@ -0,0 +1,40 @@
...
+		$airtable_datasource = new AirtableDatasource( $config['access_token'], $config['base'], $config['tables'] );
+
+		$block_name = 'Airtable (' . $airtable_datasource->get_display_name() . ')';
+
+		// TODO: Implement query UI
```
chriszarate: what does this mean?

inc/integrations/airtable-integration/airtable-integration.php
```diff
@@ -0,0 +1,40 @@
...
+				(array) $config
+			);
+			self::register_blocks_for_shopify_data_source( $config );
+		}
+	}
+
+	private static function register_blocks_for_shopify_data_source( array $config ): void {
+		$airtable_datasource = new AirtableDatasource( $config['access_token'], $config['base'], $config['tables'] );
+
+		$block_name = 'Airtable (' . $airtable_datasource->get_display_name() . ')';
```
chriszarate: This doesn't make sense to me. Won't you want to register a different block for each table? Tables can have wildly different schema and might represent different object types, with their own distinct nouns. (E.g., "events" vs "speakers").

remote-data-blocks.php
```diff
@@ -22,6 +22,11 @@
 
 define( 'REMOTE_DATA_BLOCKS__REST_NAMESPACE', 'remote-data-blocks/v1' );
 
+// Datasource services
+define( 'REMOTE_DATA_BLOCKS_AIRTABLE_SERVICE', 'airtable' );
+define( 'REMOTE_DATA_BLOCKS_GITHUB_SERVICE', 'github' );
```
chriszarate, line 27: unused

PR #40, opened 2024-09-12
Split QueryContext into base and http variants

mhsdef:
`s/QueryContext/HttpQueryContext`

Dial in the contracts a little more specifically and be consistent with other file structure.


inc/editor/configuration-loader/configuration-loader.php
```diff
@@ -208,7 +208,7 @@ private static function register_selector( string $block_title, string $type, Qu
 		);
 	}
 
-	public static function register_query( string $block_title, QueryContext $query ): void {
+	public static function register_query( string $block_title, QueryContextInterface $query ): void {
```
mhsdef, line 211: This is the practical why--this and increasing mental model accuracies. I don't see any blockers to the core code being able to handle anything that conforms to that contract.

chriszarate (approved):
awesome, good encapsulation of the http-specific implementation details

Issue #43, opened 2024-09-12
Error when activating plugin: Cannot use reserved class name 'RemoteDataBlocks\REST\true'


arunsathiya:
Following README instructions, I tried activating this plugin on a dev site, but I get the following error upon clicking `Activate` button:

```php
Fatal error: Cannot use 'RemoteDataBlocks\REST\true' as class name as it is reserved in /var/www/html/wp-content/plugins/remote-data-blocks/inc/rest/datasource-crud/datasource-crud.php on line 26
```

jblz:
Hi, @arunsathiya.

Thanks so much for giving the Remote Data Blocks plugin a try! Please note that it’s still in early stages of development and not yet suitable for production use, but we’re really glad you’ve taken it for a spin and uncovered this issue.

It looks like that error was caused by the return type declaration here:

https://github.com/Automattic/remote-data-blocks/blob/73b1e0ea7c9e292a1cb173530ea6587ea27cdc17/inc/rest/datasource-crud/datasource-crud.php#L26

Per https://www.php.net/manual/en/language.types.declarations.php, support for the literal type true was only added in PHP 8.2.

We’ll update the declaration to allow this to work in 8.1 as well and add some testing to catch this sort of thing in the future (#44).

In the meantime, please let us know if you give it a try in 8.2+ and have any feedback.

arunsathiya:
Thanks for sharing that information, @jblz—great to see it being improved. I could successfully install on PHP 8.2+, and I configured Airtable as a source under `Settings > Remote Data Blocks`, but after this step, I cannot yet see `Remote Data Blocks` or `Airtable` in the page/post editor.

jblz:
@arunsathiya I appreciate you following up.

The UI for configuring data sources is not yet hooked into the block registration piece. Some of that work has been shared in #17, but it's not quite ready yet.

For now, the data sources have to be hard-coded in a companion plugin.  Here are a couple of examples:

https://github.com/Automattic/remote-data-blocks/tree/8526923e0852a0b18c409d2ce4c523c2bd2b5fb1/example/airtable

Cheers.

maxschmeling:
@arunsathiya really appreciate the early engagement.

Just so you're aware, we've just gotten [our documentation site](https://remotedatablocks.com) online, but there's not much content yet (and we're working on a real design). Through the rest of September you should start seeing more content show up there and the API and workflows will solidify a bit as part of that process.

[...]

PR #45, opened 2024-09-12
Add DatasourceCrud tests

mhsdef:
See #43 and #44 as background.

We need to be covering more of our code asap.


mhsdef:
Repro'd the report failure. 👍 

```
PHP Fatal error:  Cannot use 'RemoteDataBlocks\REST\true' as class name as it is reserved in /home/runner/work/remote-data-blocks/remote-data-blocks/inc/rest/datasource-crud/datasource-crud.php on line 26
Script ./vendor/bin/phpunit handling the test event returned with error code 255
```

PR #49, opened 2024-09-13
Try removing slugs

mhsdef:
Slugs and UUIDs seem to overlap, generally, from what I've been able to experience. I think if we can streamline to one "primary key" in our data structure it will eliminate some confusion.

Here I've tried to remove most code while swapping in a looser `display_name` field to cover the display concerns the slug had been fulfilling at times in various UI.

## How To Test

Unit tests passing. Check your favorite integrations to make sure I haven't borked them!


mhsdef:
Punting again, for now.

PR #50, opened 2024-09-13
Move to PSR-4

### LINKED ISSUE #38: Explore moving autoloading to PSR-4
https://www.php-fig.org/psr/psr-4/

mhsdef:
Fixes #38.

Check out `inc` and see if it makes sense logically to you.

## How To Test

Unit tests passing. Smoke test around your favorite stuff if you want for extra credit.


remote-data-blocks.php
```diff
@@ -28,25 +29,24 @@
...
-Editor\FieldShortcode::init();
-Editor\QueryOverrides::init();
-Editor\PatternEditor::init();
+Editor\AdminNotices\AdminNotices::init();
+Editor\DataBinding\BlockBindings::init();
+Editor\DataBinding\FieldShortcode::init();
+Editor\DataBinding\QueryOverrides::init();
+Editor\BlockManagement\BlockRegistration::init();
+Editor\BlockManagement\ConfigurationLoader::init();
+Editor\PatternEditor\PatternEditor::init();
```
jblz, line 41: This is the only part I don't love (the repetition).

What do you think about moving the PatternEditor class up a namespace level?

remote-data-blocks.php
```diff
@@ -28,25 +29,24 @@
...
 
 // Load Settings Page
-PluginSettings::init();
+PluginSettings\PluginSettings::init();
```
jblz, line 44: Same here

jblz (approved):
Holy overhaul, Batman! <3

remote-data-blocks.php
mhsdef: Yeah, I agree with you from the perspective _inside_ the file. My hesitation would be from the other direction.

Do you think the folders make sense from the perspective of the file structure, relative to peers?

jblz: I only think a dir / folder is worth it if we expect numerous related things to live together in it.  In the case of `PatternEditor` & `PluginSettings`, they're the only files in them, so I'd just move them up a level and lose their folders.  No idea if that's spec compliant, though 🤷

mhsdef: I'm tracking with you. And I agree aesthetically.

For those particular files, I feel like that leads to situations that imply an architectural importance that we probably don't intended.

For example, take `PluginSettings.php`. As is in a folder, it sits here, with rough "peers" which feels logically consistent to me relative to the application's architecture:

```
|- Config
|- Editor
|- HttpClient
|- ...
|- PluginSettings
|- ...
```

If we pull it up, the lone file directly under the root implies, to me, that _This Is Particularly Important Start Here:_

```
|- Config
|- Editor
|- HttpClient
|- ...
|- REST
|- WpDbStorage
|- PluginSettings.php
```

Does that make sense? Maybe it's just me.

jblz: Yeah, that makes sense. Tradeoffs and all that.

PR #53, opened 2024-09-14
Add block identifier to binding args in patterns

chriszarate:
Fix: In the block editor, we were adding the remote data block identifier (currently the block name) to the binding args so that the binding could be identified even if it was separated from its parent container / context. However, we were not adding the block identifier in registered patterns (despite the fact that they are always separated!).

1. This PR addresses that, helping us to make connections between patterns and the remote data blocks that they may be associated with.
2. Bugfix enabled by 1: We now pass the block identifier into additional utility functions so that we only operate on bindings that are relevant to the current remote data block.
3. Improvement enabled by 1: We rename the property used to store the block identifier from `name` to `block`. The former is very vague and used in multiple places for other purposes. The latter is more specific. We may want to move from a block name identifier to something immutable, but for now block name is all we have available and it unblocks other progress.

### DIFF
--- a/inc/editor/block-bindings/block-bindings.php
+++ b/inc/editor/block-bindings/block-bindings.php
@@ -214,7 +214,7 @@ class BlockBindings {
 		$field_name = $source_args['field'];
 		$index      = $source_args['index'] ?? 0; // Index is only set for loop queries.
 
-		if ( isset( $source_args['name'] ) && $source_args['name'] !== $block_name ) {
+		if ( isset( $source_args['block'] ) && $source_args['block'] !== $block_name ) {
 			self::log_error( 'Block binding belongs to a different remote data block', $block_name, $field_name );
 			return null;
 		}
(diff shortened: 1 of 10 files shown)


mhsdef:
@chriszarate heads up, this change broke stuff. I'm not sure if it's actual breakage or if implementations just need some sort of update to get back on track.

Check out the previous commit (08099cdf6b21bbd3fa4a8d3d79b6a3f90ac6b8e0) and check the Github block pattern select or the Shopify product pattern select there vs this commit.

chriszarate:
> this change broke stuff. I'm not sure if it's actual breakage or if implementations just need some sort of update to get back on track.

It most likely broke any bindings (or patterns with bindings) that existed prior to this change. Are you able to replicate with a fresh binding?

mhsdef:
you mean...by creating a new block?

yeah, `trunk`:

Screenshot 2024-09-23 at 8 48 42 AM


At 08099cdf6b21bbd3fa4a8d3d79b6a3f90ac6b8e0:
Screenshot 2024-09-23 at 8 51 43 AM

chriszarate:
Ah ok, perfect, thanks for detail. Will fix.

PR #54, opened 2024-09-15
Allow association of patterns with remote data blocks

chriszarate:
Render a select box in pattern editor to associate patterns with remote data blocks. This allows patterns (synced and unsynced) to be included in the pattern chooser and for the block binding controls to be rendered in the pattern editor.

Screenshot 2024-09-14 at 20 14 43

Screenshot 2024-09-14 at 20 15 43


mhsdef:
I can get to the dialog and set a pattern association as shown. But, it's not showing when I go to select with the block. 

Any tips for testing?

chriszarate:
> Any tips for testing?

After making the association, it does require you to set a binding for at least one of the blocks in the pattern. This should be made clear somewhere, likely the pattern selection modal.

mhsdef:
woof, sorry for the noise, first time using codecov and this is what it's doing out of the box.

I'll try the pattern association some more. And see if I can dial down this bot.

mhsdef:
I'll dial or rip that thing out tomorrow.

https://docs.codecov.com/docs/common-recipe-list

It's def not meant to be blocking. Or quite so noisy. Just looking for holistic picture.

mhsdef:
Nice. Got it wired together. Pain in a lot of places, but, mostly that's on core for how half baked these pieces we _need_ are right now. 

Good step. 👍

mhsdef (approved):

PR #66, opened 2024-09-16
Use CompatibleHttpDatasource to bridge UI-managed and user-defined data sources

chriszarate:
This change removes the `remote_data_blocks_data_sources` filter, which encourages users who commit data sources via code to mix them into UI-managed data sources. This already causes issues with array vs. object representation, and exposes and locks us into internal implementation details. (We may still want to offer a filter for UI-managed data sources, but likely need to do it in a place where those changes take actual effect, instead of just a display context.)

Instead, we can allow data sources to declare themselves compatible and provide methods that implement that compatibility. Right now that is limited to simply displaying them on the settings screen, but we can expand that if desired.

Note: Over time, we probably want to make changes to the object structure of UI-managed data sources to align them more closely with (if not directly implement) `HttpDatasource`.


inc/Config/Datasource/CompatibleHttpDatasource.php
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * HttpDatasource class
+ *
+ * Implements the HttpDatasourceInterface to define a generic HTTP datasource.
+ *
+ * @package remote-data-blocks
+ * @since 0.1.0
+ */
+abstract class CompatibleHttpDatasource extends HttpDatasource {
```
mhsdef, line 13:
```suggestion
abstract class HardcodeHttpDatasource extends HttpDatasource {
```

Whatcha think? Not elegant, but, to the point. 

"Compatible" doesn't give me much to hang my head on mentally.

mhsdef (approved):
> Over time, we probably want to make changes to the object structure of UI-managed data sources to align them more closely with (if not directly implement) HttpDatasource.

Strong agree. I'd like to see us:

- Move toward structural alignments as you note there sooner than later.
- Switch everything configy to consistent array or object representations so we can stop the guesswork, transformations, and bugs. (I vote arrays since they are dead simple and IMO shift across execution environments with less drama.)
- Further teach `ConfigurationLoader` about hardcoded, wpdb, and (eventually) wpvip integrated datasources such that our app logic can count on that class as the single source of truth from there onward and, for most cases, treat them all the same.

So instead of:

```
// Merge the UI-managed and user-defined sources.
$crud_sources         = DatasourceCrud::get_data_sources();
$user_defined_sources = ConfigurationLoader::get_compatible_data_sources();
```

...in (soon) multiple places. We can just straight:

```
ConfigurationLoader::get_data_sources();
```

...because we'll have aligned, merged, etc earlier in the call stack.

inc/Config/Datasource/CompatibleHttpDatasource.php
chriszarate: I hear that, "compatible" is sort of inherent in our use of interfaces, so it's not a good descriptor. Let's get some wider feedback and I'll make the edit.

PR #67, opened 2024-09-16
Fix wp.components.TextControl is deprecated notice

mhsdef:
Our tests have exposed this:

```
stderr | tests/src/blocks/remote-data-container/components/modals/input-modal.test.tsx > InputModal > opens the modal when the button is clicked
Bottom margin styles for wp.components.TextControl is deprecated since version 6.7 and will be removed in version 7.0. Note: Set the `__nextHasNoMarginBottom` prop to true to start opting into the new styles, which will become the default in a future version.
```


chriszarate (approved):
Nice was wondering about that!

PR #68, opened 2024-09-16
Split ConfigurationLoader into ConfigStore and ConfigRegistry

chriszarate:
A hopefully uncontroversial naming change that paces the way for global access functions. `ConfigurationLoader` was a poor name (guilty!) and was doing two very different jobs. This splits the class and updates all references. No functional changes.

A follow-on PR will introduce access functions and update docs.


maxschmeling:
If I'm understanding the purpose correctly on quick glance, isn't it more of a `ConfigRegistrar` than `ConfigRegistry`?

Not sure I like the `Registrar` name either though. But `Registry` and `Store` seem like synonyms.

chriszarate:
Happy to change it, the main thing is having the store / access separate from the entrypoints to register things.

chriszarate:
@maxschmeling FWIW I was following WP patterns for this, which has lots of `Registry` classes with `register_*` methods:

https://github.com/search?q=repo:WordPress/wordpress-develop%20registry&type=code

maxschmeling:
Might be a slight difference from some of those in the link because in those cases, they actually are registering with internal state. But having the name match up makes sense. And it's probably a pedantic argument and probably doesn't matter.

LGTM 🚢

mhsdef (approved):
@maxschmeling not pedantic, this stuff matters a ton over time. I have seen so much complexity build in large codebases _just_ from people confused over wtf this abstraction meant vs that abstraction (and giving up a rolling their own third thing).

This feels like a step in the right direction as long as we aren't overly attached to exactly what it is here. Eg, I think, ultimately, we may need to break down query vs block registration concerns.

:shipit:

chriszarate:
> Eg, I think, ultimately, we may need to break down query vs block registration concerns.

Indeed

PR #69, opened 2024-09-16
Add access functions

chriszarate:
Add access functions in the global scope and document them.

- I removed the existing `use function` statements in example code. They were used inconsistently, they are optional, and it seemed better to keep example code simpler.
- I did not create an access function for `ConfigRegistry::register_query()` because I think the use case is very shaky. The example code that uses it is likely slated for removal.
- I was somewhat torn about whether to update `ConfigRegistryTest` to use these access functions, but did so in 03829f4a2381abcfdb3cc1f22c0cc2ff451ef2e0. Thoughts?


mhsdef:
So, what I think you have going here...is the nascent "contract" between us and the customers. 

That's great. I think we should lean into that. This is the entry point. Our external API.

Should we simply call it `functions.php` in keeping with that?

> I was somewhat torn about whether to update ConfigRegistryTest to use these access functions, but did so in https://github.com/Automattic/remote-data-blocks/commit/03829f4a2381abcfdb3cc1f22c0cc2ff451ef2e0. Thoughts?

Yep, I see what you mean.

I lean toward we do what you did but move and rename the test -- `FunctionsTest` or whatever. We want to test the "contract", what customers give us, and the expect results. We don't need/want to test every implementation angle. I guess, in jargon speak, favor integration tests over unit tests?

chriszarate:
I thought `functions.php` was a theme thing, but as long as it doesn't cause conflicts, wfm

mhsdef:
[Not exclusively so](https://github.com/WordPress/WordPress/blob/747bda52eb923cdc671c38f647a0e267faf994f6/wp-includes/functions.php), but, you're right that the common understanding is associated with theming.

Do you think prefixing it like `rdb-functions.php` would be better?

chriszarate:
> Do you think prefixing it like `rdb-functions.php` would be better?

We can leave it as `functions.php` and see if it causes confusion, guessing it will be just fine.

mhsdef (approved):

PR #75, opened 2024-09-19
Add static example API

chriszarate:
This implements the task of creating a "static" toy API that is bundled with the plugin. This allows new users and explorers to immediately begin using the plugin meaningfully without any configuration or provisioning credentials.

This API mirrors the "Event Planning" template from Airtable, but it is easily swappable with another API.

This API is not implemented an HTTP API, mainly because it complicates local development: routing a request from inside a container to a knowable URL is non-trivial and requires configuration (which we want to avoid). Instead the queries implement a custom `QueryRunner` in order to serve data from a static JSON file.

Along the way, we have made some improvements to the `QueryRunner` that will benefit other implementors of a non-HTTP API:

- We separate the `execute` method (which handles field mapping) from the actual HTTP request, which more flexibly allows for custom transports.
- We prevent `ResponseInterface` from leaking into `QueryContextInterface`. Fixing this mistake allows queries to be truly agnostic towards their transport.

### DIFF
--- a/example/remote-data-blocks-example-code.php
+++ b/example/remote-data-blocks-example-code.php
@@ -35,7 +35,6 @@ function load_only_if_parent_plugin_is_active() {
 	}
 
 	require_once __DIR__ . '/airtable/elden-ring-map/register.php';
-	require_once __DIR__ . '/airtable/events/register.php';
 	require_once __DIR__ . '/github/remote-data-blocks/register.php';
 	require_once __DIR__ . '/rest-api/art-institute/register.php';
 	require_once __DIR__ . '/rest-api/zip-code/register.php';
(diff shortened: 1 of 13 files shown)


mhsdef:
@chriszarate heads up on this one too -- this one broke the Airtable event list for some reason.

I think it's specific to that integration as Shekhar's Westeros list with Sheets still seems to work.

example/remote-data-blocks-example-code.php
```diff
@@ -35,7 +35,6 @@ function load_only_if_parent_plugin_is_active() {
 	}
 
 	require_once __DIR__ . '/airtable/elden-ring-map/register.php';
-	require_once __DIR__ . '/airtable/events/register.php';
```
mhsdef, line 38: oh, yea, that'd do it. Was that intentional?

chriszarate: Yes, intentional. Because in this PR:

> This API mirrors the "Event Planning" template from Airtable, but it is easily swappable with another API.

PR #78, opened 2024-09-19
Add WordPress Playground blueprint

chriszarate:
This adds a working blueprint for WordPress Playground. This is not an alternative dev environment, since Playground still does not support cURL transport, object cache, etc. But it is a useful way to explore the capabilities of the plugin.

Try it with this link, which points to the `blueprint.json` in this PR:

https://playground.wordpress.net/?blueprint-url=https://raw.githubusercontent.com/Automattic/remote-data-blocks/8110ff9b3ed36823935bcfccabf0510c9a665051/blueprint.json

It works via #75, which ships a static example API that does not rely on HTTP transport. A single Remote Data Block is available, "Conference Event":

https://github.com/user-attachments/assets/f69787d0-304b-4c7f-98ef-53153486a2f0


blueprint.json
```diff
@@ -0,0 +1,32 @@
...
+  "steps": [
+    {
+      "step": "installPlugin",
+      "options": {
+        "activate": true
+      },
+      "pluginZipFile": {
+        "caption": "Installing Remote Data Blocks",
+        "resource": "url",
+        "url": "https://github-proxy.com/proxy/?repo=Automattic/remote-data-blocks&release=ae7f58e&asset=remote-data-blocks.zip"
```
mhsdef, line 35:
```suggestion
        "url": "https://github-proxy.com/proxy/?repo=Automattic/remote-data-blocks&release=latest&asset=remote-data-blocks.zip"
```

chriszarate: I wish, doesn't work

mhsdef:
maybe worth tossing a step to set the site title to something non-generic? (in case of multiple open windows, for example)

```
{
      "step": "setSiteOptions",
      "options": {
        "blogname": "Remote Data Blocks"
      }
}
```

blueprint.json
mhsdef: Hmm... thought @stoph was using that somewhere.

mehmoodak: It's working for me, able to download the zip file with `latest` in the url.

chriszarate: @mehmoodak But is the downloaded file non-empty or unzippable? I think @stoph was working on a fix.

mehmoodak: It's empty 😓

jarekmorawski:
Thanks for making this happen! I can't seem to get the resources to load; it's stuck at `Loading...`. Is that expected?

image

mehmoodak:
@jarekmorawski is this happening on a particular remote block or all remote blocks?

chriszarate:
I synced quickly with Jarek, the issue happens when editing content via the Site Editor as opposed to the regular post editor. Will investigate

PR #80, opened 2024-09-19
Map Example: Fix in editor and enable on frontend w/ server-fetched / rendered coordinates

jblz:
Airtable Map blocks now work in the editor and frontend. Frontend data is populated from the remote server (or WP cache, if enabled).


example/airtable/elden-ring-map/src/blocks/elden-ring-map/view.js
```diff
@@ -1,3 +1,26 @@
...
+import { createRoot } from '@wordpress/element';
+
+import { Map } from './map';
+
+domReady( () => {
+	const elements = [
+		...document.getElementsByClassName( 'wp-block-remote-data-blocks-elden-ring-map' ),
+	];
+
+	elements.forEach( element => {
```
jblz, line 12: I'm pretty sure this will run multiple times if there are more than one map block in a doc.  It will have to be tied to the individual element once I have more time to look at it.

example/airtable/elden-ring-map/src/blocks/elden-ring-map/view.js
```diff
@@ -1,3 +1,26 @@
...
+			const parsed = JSON.parse( context );
+			coordinates = parsed.coordinates ?? [];
+		} catch ( error ) {}
+
+		if ( ! coordinates.length ) {
+			return;
+		}
+
+		const root = createRoot( element );
+		root.render( <Map coordinates={ coordinates } /> );
```
jblz, line 25: I tried to use the Interactivity API for this, but since it's loaded as a (preact) module, I didn't have access to a react root to put the existing Map component stuff.  It will take some refactoring to do it that way.

src/blocks/remote-data-container/components/item-list/item-list.tsx
```diff
@@ -32,13 +32,22 @@ export function ItemList( props: ItemListProps ) {
...
-		const blocks =
-			pattern?.blocks.map( block => cloneBlockWithAttributes( block, result, props.blockName ) ) ??
-			[];
-
-		return (
-			<ItemPreview key={ index } blocks={ blocks } onSelect={ () => props.onSelect( result ) } />
-		);
-	} );
+	return (
+		<ul>
```
jblz, line 35: This is kinda unrelated, but I noticed that we're printing `<li>` elements outside of an `<ul>`, `<ol>`, or `<menu>` element which is [out of spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-li-element).

example/airtable/elden-ring-map/inc/interactivity-store/interactivity-store.php
```diff
@@ -0,0 +1,37 @@
...
+
+defined( 'ABSPATH' ) || exit();
+
+use RemoteDataBlocks\Editor\DataBinding\BlockBindings;
+use WP_Block;
+
+class InteractivityStore {
+	public static function get_map_interactive_context( WP_Block $block ): array {
+		$query_context = [
+			'blockName'  => 'remote-data-blocks/elden-ring-location', // Why doesn't this match: $block->name,
```
chriszarate: because you are calling this from the Elden Ring Map block

example/airtable/elden-ring-map/inc/interactivity-store/interactivity-store.php
```diff
@@ -0,0 +1,37 @@
...
+			'queryInput' => $block->context['remote-data-blocks/remoteData']['queryInput'],
```
chriszarate:
```suggestion
			'queryInput' => $block->context[ BlockBindings::$context_name ]['queryInput'],
```

src/blocks/remote-data-container/components/item-list/item-list.tsx
chriszarate: Thanks, this was copied from core, but probably missing a container component that supplies the list element.

PR #81, opened 2024-09-20
Extend ExampleApiDataSource from CodedHttpDatasource

shekharnwagh:
This fixes the plugin settings page crashing with error -

```
Fatal error: Uncaught Error: RemoteDataBlocks\Config\QueryContext\HttpQueryContext::__construct(): Argument #1 ($datasource) must be of type RemoteDataBlocks\Config\Datasource\HttpDatasource, RemoteDataBlocks\ExampleApi\Queries\ExampleApiDataSource given, called in /var/www/html/wp-content/plugins/remote-data-blocks/inc/ExampleApi/ExampleApi.php on line 34
in /var/www/html/wp-content/plugins/remote-data-blocks/inc/Config/QueryContext/HttpQueryContext.php on line 64
```

Extending the `ExampleApiDataSource` from `CodedHttpDatasource` fixes the type issue thrown when the `ExampleApiDataSource` instance is passed to query classes, for Example API block. This also has the upside of showing the Example Data Source in the setting page.

### DIFF
--- a/inc/ExampleApi/Queries/ExampleApiDataSource.php
+++ b/inc/ExampleApi/Queries/ExampleApiDataSource.php
@@ -2,13 +2,13 @@
 
 namespace RemoteDataBlocks\ExampleApi\Queries;
 
-use RemoteDataBlocks\Config\Datasource\DatasourceInterface;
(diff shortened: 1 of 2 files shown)


chriszarate (approved):
Thanks!

PR #82, opened 2024-09-20
Wire up Validator and Sanitizer

mhsdef:
This PR wires up `Validator`s and `Sanitizer`s introduced in #83.

The main things it does are:

- Reusable validation/sanitization whether data is via REST, hardcode, or remote fetch (pre-binding).
- Firm up our data structures.

## How To Test

Unit tests passing. Smoke test your favorite integrations.


example/airtable/elden-ring-map/inc/queries/class-airtable-elden-ring-list-locations-query.php
```diff
@@ -39,6 +39,6 @@ class AirtableEldenRingListLocationsQuery extends HttpQueryContext {
 	];
 
 	public function get_endpoint( array $input_variables ): string {
-		return $this->get_datasource()->get_endpoint( 'locations' ) . '?filterByFormula=FIND%28%27' . $input_variables['map_name'] . '%27%2C%20%7BMap%7D%29%3E0';
+		return $this->get_datasource()->get_endpoint() . '/tblc82R9msH4Yh6ZX?filterByFormula=FIND%28%27' . $input_variables['map_name'] . '%27%2C%20%7BMap%7D%29%3E0';
```
maxschmeling, line 42: Can we pull the table ID from the config? It would be ideal if getting the table ID here worked the same way with a hard coded data source as it does with a data source provided by the settings page.

inc/Validation/Validator.php
```diff
@@ -25,6 +25,10 @@ public function validate( array $data ): bool|WP_Error {
 	}
 
 	private function validate_schema( array $schema, $data ): bool|WP_Error {
```
maxschmeling, line 27: Should we use an off the shelf library for this type of thing?

mhsdef: I looked a few things:
- Straight reuse of WordPress REST validation/sanitization - https://github.com/Automattic/remote-data-blocks/pull/83
- Concocting something directly JSON Schema 4-based.
- `Respect/Validation` - https://github.com/Respect/Validation
- Symfony Validation - https://symfony.com/doc/current/validation.html
- And glanced at a few others like Laravel validation, `rakit/validation`, and `vlucas/valitron`.

If you have something in mind, I'm certainly open to it. I just ended up here--a WP REST-like [which is JSON Schema 3/4ish itself]--after weighing various pros/cons.

example/airtable/elden-ring-map/inc/queries/class-airtable-elden-ring-list-locations-query.php
mhsdef: I feel pretty confident we can make smooth user experiences across both ways so I wouldn't fret relative to that angle. We will have excellent flexibility in UI/UX if we align our data constructs well down low with what they represent.

To me, a fundamental question here is what exactly _is_ an "Airtable data source"? What is the primary entity customers are connecting up to?

Is it the base? Is it the base+table?

maxschmeling: I think that's still somewhat of an open question and something we need feedback on. It requires more flexibility from our plugin, but I think ideally it can be all of the above. If the data source specifies a specific table, then it's essentially locked. If it doesn't, then the query config / UI has to specify the table for each case. So that could be in the query code, or perhaps we could build UI to support it in Gutenberg. But that last use case is down the road a bit.

inc/Validation/Validator.php
maxschmeling: I'm ok with this, but if we find ourselves needing more and more from it, it might be worth more exploration.

inc/Integrations/Google/Sheets/GoogleSheetsDatasource.php
```diff
@@ -46,31 +48,22 @@ class GoogleSheetsDatasource extends HttpDatasource {
 					'universe_domain'             => [ 'type' => 'string' ],
 				],
 			],
+			'display_name'   => [ 'type' => 'string' ],
```
shekharnwagh, line 51: @mhsdef Is there a reason we made `display_name` mandatory here, but are not requiring it for other data sources ?

src/data-sources/airtable/AirtableSettings.tsx
shekharnwagh, line 1: @mhsdef Why was the `table` field removed from the Airtable Source Config ?

inc/Integrations/Google/Sheets/GoogleSheetsDatasource.php
mhsdef: probably accidental that it alone is that way. I'll adjust it to optional too.

I think, ultimately, we'll want to require on the backend. Maybe able to prefill it (or the slug?) on the front for better UX.

src/data-sources/airtable/AirtableSettings.tsx
mhsdef: Reverted this in #105.

PR #89, opened 2024-09-23
Automatically add block arg to user-registered patterns

chriszarate:
Fixes a bug introduced by https://github.com/Automattic/remote-data-blocks/pull/53 (h/t @mhsdef). We now detect pattern compatibility by inspecting the block arg of the block binding. Rather than ask users to bake this arg into their supplied patterns (which is probably error-prone), we can inject it for them when they call `register_remote_data_block_pattern`.

This PR also removes the unused `patterns` key of the registered block config.


mhsdef (approved):
Coolio, it works!

Screenshot 2024-09-23 at 11 13 19 AM

And tests, woo. Ship it.

PR #91, opened 2024-09-24
Update QueryContextInterface and HttpQueryContext to define variables using function

chriszarate:
Defining input and output variables via a property has been flagged by multiple people as confusing, limiting, and inconsistent. This was especially apparent when working on query docs in https://github.com/Automattic/remote-data-blocks/pull/90.

It's confusing mainly because the property names (`$input_variables` and `$output_variables`) do not communicate that they are _definitions_ and conflict with identically named arguments that contain the runtime _values_.

It's limiting because implementors often wanted to reach for code to help them define variables, resulting in awkward constructor use ([example](https://github.com/Automattic/remote-data-blocks/blob/65fae40eeb696bfbd68ddd1c958df74094a00b9c/example/rest-api/art-institute/inc/queries/class-art-institute-search-query.php#L20)).

It's inconsistent because every other property is defined via a method.

The diff is a bit scary but is mostly whitespace. There are no actual functional changes, just moving from this:

```php
class MyQuery extends HttpQueryContext {
  public array $input_variables = [ /* definition */ ];
}
```

... to this:

```php
class MyQuery extends HttpQueryContext {
  public function define_input_variables(): array {
    return [ /* definition, unchanged */ ];
  }
}
```


mhsdef:
This kinda space is where I intensely miss Elixir. Or anything with a good pipe operator. But, yeah, Elixir because it's ergonomics are TEH BEST. ❤️ 😄 

```
input_variables
 |> list_query1(datasource)
 |> Enum.random()
 |> get_query(datasource)
 |> apply_pattern()
```

mhsdef (approved):
sorry! I still feel too fuzzy about the way the core of that interacts to give deep feedback.

It seems fine if you feel good about it; and, yea, deduping the overlap feels like a win.

`s/define_*_variables/get_*_variables_definition`? or `get_*_variables_schema`?

(I feel like the "define" part has already happened, when the person wrote it, so "getting" it feels more natural. "definition" is more human-y speak but we do have "schema" as a thing now elsewhere and our audience is devs and it too is pretty well known.)

PR #92, opened 2024-09-24
Remove `get_metadata` from `QueryContextInterface`

chriszarate:
Remove `get_metadata` from `QueryContextInterface` and implement it as a filter instead. Use cases are currently limited to field shortcodes and it is completely unused in example code. Moving to `QueryRunner` also gives implementors another control mechanism via custom QueryRunner, if they choose. Limiting the surface area of `QueryContextInterface` helps with docs.

No change in functionality, just moving.


mhsdef (approved):
Sounds good to me.

I'll note that I think we should (continue) to be thoughtful of tradeoffs with action / filter api and object / interface implementations.

It's a quite delicate balance between community norms and logical integrity. In general, I think it's good we've had relatively few hooks, but also sometimes it might be right as you note here. 🙃

PR #96, opened 2024-09-24
Move our PHP to declare(strict_types = 1)

mhsdef:
**tl;dr:** this makes type hinting work like you actually think it does.

In PHP, `declare(strict_types=1);` is a declaration that enables strict typing for a specific PHP file. This declaration affects how type checking is performed for function arguments and return values within that file. 

Without strict typing (default behavior):

1. PHP performs type juggling (automatic type conversion) when passing arguments to functions or returning values.
2. Allows for more flexible coding but can lead to unexpected results or hidden bugs.
3. Type hints are treated as suggestions rather than strict requirements.

With strict typing, we gain:

1. Better type safety: Strict typing provides stronger type checking, reducing the risk of type-related errors. With strict typing, type mismatches result in a `TypeError`, while without it, PHP attempts to convert types silently.
2. Code clarity: Strict typing requires function signatures more explicit about expected types.

### DIFF
--- a/bin/wp-env/decoupled/mu-plugins/home-url-filter.php
+++ b/bin/wp-env/decoupled/mu-plugins/home-url-filter.php
@@ -1,4 +1,7 @@
 <?php
+
+declare(strict_types = 1);
+
 add_filter('home_url', function ( $url, $path ) {
 	if ( have_posts() && is_singular() ) {
 		return site_url( $path );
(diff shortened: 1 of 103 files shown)


mhsdef:
As example of it doing what you'd think it was already doing... a couple of things it caught: 5d9614ac0c1d8cca2387f3f811a198ed3a1fc62a

PR #98, opened 2024-09-24
Update `withBlockBinding` filter strategy to override core

chriszarate:
Recent changes to Gutenberg in `trunk` broke our method of rendering remote data in the block editor. Core is preparing a public API that will allow custom block bindings to more easily do what we are doing manually. In its current state, however, it does not allow non-core code to participate and instead falls back to the binding label ("Remote Field Binding", in our case).

By adjusting our filter strategy, we can override this core behavior in a way that works for current and future releases.

- Move from `editor.BlockEdit` to `blocks.registerBlockType`.
- Use a higher filter priority (lower number) than Core.

This PR also updates the binding label for consistency (visible in sidebar block settings).

### DIFF
--- a/inc/Editor/DataBinding/BlockBindings.php
+++ b/inc/Editor/DataBinding/BlockBindings.php
@@ -24,7 +24,7 @@ class BlockBindings {
 	 */
 	public static function register_block_bindings(): void {
 		register_block_bindings_source( self::$binding_source, [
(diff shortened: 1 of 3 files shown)


jblz (approved):

maxschmeling (commented):
This is awesome! Glad we can make it work consistently across versions!

PR #102, opened 2024-09-24
Allow "inner blocks" block pattern registration option

chriszarate:
This provides a fix for the missing map pattern raised by @jblz after https://github.com/Automattic/remote-data-blocks/pull/89, but goes further to address the use case of bypassing the pattern selection flow with a designated pattern.

The fix for restoring the pattern to the pattern selection modal is fairly narrow:

- [Include blockTypes in supportedPatterns logic](https://github.com/Automattic/remote-data-blocks/commit/895fb4ccb70c15b2297febe0ab1586bb291c3c01)

Beyond that, this PR does three things:

- We are already registering a "default" pattern that displays the data in a simple format. This allows us to be sure that there is always a pattern available (e.g., when we need to display items in a list for selection). Formalizing this a bit in the localized block config allows us to identify the default pattern without a kluge-y assumption (based on the format of the pattern name).
  - [Update BlockPatterns to always register a default pattern and return pattern name](https://github.com/Automattic/remote-data-blocks/commit/6405850f5f0636ba38dd61730e29460f4ede2783)
  - [Provide default pattern in localized remote data block config](https://github.com/Automattic/remote-data-blocks/commit/0cde3538b2d28e89f48a6b5196b9c2f083a5bfc4)
- Looking at example usage of `register_remote_data_block_pattern`, it's clear that pattern title should be a first-class argument. Updating the function signature makes the ergonomics better.
  - [Update register_remote_data_block_pattern signature to reflect usage](https://github.com/Automattic/remote-data-blocks/commit/58c8bddd4d30dbc269b161018df95db7496aba42)
- Finally, we can recognize a `role` option passed to `register_remote_data_block_pattern` that allows us to designate patterns for a specific use case. A `role` of `inner_blocks` means that the pattern should always be used as the inner blocks of a remote data block, thereby bypassing the pattern selection flow. (Please suggest better names! I avoided `default` here since "default pattern" already means something else.)
  - [Recognize role=inner_blocks pattern option](https://github.com/Automattic/remote-data-blocks/commit/e0ee8fcafc69c42b5b210d3d9a853a2c0ae826bb)
  - [Use config-provided patterns in usePatterns](https://github.com/Automattic/remote-data-blocks/commit/80e36bf27eeca6d7b9dd3ae7a9101c6da3dad2ee)

### DIFF
--- a/docs/extending.md
+++ b/docs/extending.md
@@ -98,7 +98,7 @@ function register_your_block_pattern() {
     $block_name    = 'Your Custom Block';
     $block_pattern = file_get_contents( **DIR** . '/your-pattern.html' );
 
-    register_remote_data_block_pattern( $block_name, 'your-namespace/your-pattern', $block_pattern );
+    register_remote_data_block_pattern( $block_name, 'Pattern Title', $block_pattern );
 }
 add_action( 'init', 'YourNamespace\\register_your_block_pattern' );
 ```
--- a/example/airtable/elden-ring-map/register.php
+++ b/example/airtable/elden-ring-map/register.php
@@ -26,7 +26,7 @@ function register_airtable_elden_ring_map_block() {
 	register_remote_data_list_query( $block_name, $list_maps_query );
 
 	$block_pattern = file_get_contents( __DIR__ . '/inc/patterns/map-pattern.html' );
-	register_remote_data_block_pattern( $block_name, 'Elden Ring Map', $block_pattern );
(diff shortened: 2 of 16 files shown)


jblz (approved):
Yer a wizard, Zarate.

chriszarate:
> Please suggest better names! I avoided default here since "default pattern" already means something else.

Merging to unblock folks, but we can easily revisit the naming in a simple update.

PR #110, opened 2024-09-25
Add launch.json

chriszarate:
Had this sitting locally forever, keep forgetting to PR


.vscode/launch.json
```diff
@@ -0,0 +1,14 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Listen for Xdebug",
+      "type": "php",
+      "request": "launch",
+      "port": 9003,
+      "pathMappings": {
+        "/var/www/html/wp-content/plugins/remote-data-blocks": "${workspaceFolder}/"
```
jblz:
```suggestion
        "/var/www/html/wp-content/plugins/remote-data-blocks": "${workspaceFolder}/",
        "/var/www/html/wp-content/plugins/example": "${workspaceFolder}/example"
```

This makes it work for debugging examples as well.

.vscode/launch.json
```diff
@@ -0,0 +1,14 @@
...
+      }
+    }
```
jblz, line 20:
```suggestion
    },
    {
      "type": "chrome",
      "request": "launch",
      "name": "Launch Chrome against localhost",
      "url": "http://localhost:8888/",
      "webRoot": "${workspaceFolder}"
    }
```

While we're at it, I use this one regularly to debug the frontend (I like to keep a Chrome profile per project).

jblz (dismissed):
thanks for PR'ing

PR #111, opened 2024-09-26
add fields selection in airtable settings form

shekharnwagh:
# Description

This PR adds field selection to the Airtable data source settings page. The list of fields is shown as checkboxes, and the user can select the fields to be used in the remote data block. This input is shown once a table is selected.

The fields are currently stored in the data source settings option under the `tables` field where it lives under the key `output_query_mappings` along side the `id` and `name` of the table. An example of updated airtable datasource settings is -

```json
{
    "uuid": "",
    "service": "airtable",
    "access_token": "personal_access_token",
    "base": {
        "id": "base_id",
        "name": "Base Name"
    },
    "tables": [
        {
            "id": "table_id",
            "name": "Table Name",
            "output_query_mappings": [
                {
                    "name": "Field Name 1"
                },
                {
                    "name": "Field Name 2"
                }
            ]
        }
    ],
    "slug": "airtable-slug"
}
```

| Screenshot 2024-09-26 at 5 18 01 PM |
| :---: |
| Airtable data source settings page with field selection |

# Testing

Try creating and updating a new airtable data source with this new field selection input. Verify that the fields are correctly saved and retrieved from the database.


inc/Integrations/Airtable/AirtableDatasource.php
```diff
@@ -34,11 +34,20 @@ class AirtableDatasource extends HttpDatasource {
...
 						],
+						'output_query_mappings' => [
+							'type'  => 'array',
+							'items' => [
+								'type'       => 'object',
+								'properties' => [
+									'name' => [ 'type' => 'string' ],
+								],
+							],
+						],
```
mhsdef, line 50: I'm ok with us sticking this here if it helps make it work for next week.

But...pre-warning, I think it is gonna need to come out to a new flow involving loading into the query abstractions after that. We can have that discussion then though.

inc/Sanitization/Sanitizer.php
```diff
@@ -72,6 +72,9 @@ private function sanitize_config( array $config, array $schema ): array {
 					if ( is_array( $value ) ) {
 						if ( isset( $field_schema['items'] ) ) {
 							$sanitized[ $key ] = array_map(function ( $item ) use ( $field_schema ) {
+								if ( is_array( $item ) && isset( $field_schema['items']['properties'] ) ) {
+									return $this->sanitize_config( $item, $field_schema['items']['properties'] );
+								}
```
mhsdef, line 77: thanks!

tests/inc/Sanitization/SanitizerTest.php
```diff
@@ -110,6 +110,55 @@ public function test_sanitize_nested_array() {
 		$this->assertSame( $expected, $result );
 	}
 
+	public function test_sanitize_nested_array_of_objects() {
```
mhsdef, line 113: And tests for it! ❤️

mhsdef (approved):

inc/Integrations/Airtable/AirtableDatasource.php
shekharnwagh: > I think it is gonna need to come out to a new flow involving loading into the query abstractions

Yep, only added it here as stopgap until we do that.

PR #114, opened 2024-09-26
Do not auto-align assignments to the right-most level

jblz:
We don't want to enforce all subsequent assignments to have the same spacing before the assignment operator.
Doing so means a single change to one line can result in a large changeset that's more difficult to review.

Case in point: reintroducing the `map_name` key to the associative array here:
https://github.com/Automattic/remote-data-blocks/pull/80/commits/3c272bce662c8974313359e181427db1ebf62f5f

...resulted in a number of unrelated changes that needed to be applied here:
https://github.com/Automattic/remote-data-blocks/pull/80/commits/579313a01b99902f2be348ff351e7be962b224f4

In the context of a pull request, merge commit, etc. etc., this obscures the intent behind the change and makes it harder to spot bugs.


mehmoodak (approved):

mehmoodak:
Should we do something similar for PHPDocs?

image

maxschmeling:
@jblz can we rerun this and generate a fresh PR?

jblz:
@maxschmeling done.  whattya think?

jblz:
@maxschmeling I bumped this again.  Here's what I did:

* check out this branch
* `git reset HEAD~1`
* `git checkout -- .`
* `git pull origin trunk`
* `nvm i; npm i; composer install` (for good measure)
* `composer run phpcs-fix`
* commit & force push the result

I'll let someone else get this landed if it's wanted.

PR #118, opened 2024-09-27
Add query context loading from persisted data

mhsdef:
We need a way to load query contexts from persistence in lieu of #111.

This change lays some tentative groundwork there for the future and then hooks it up.

## How to Test

1. Add an Airtable datasource via settings.
2. Add some fields.
3. Add blocks.


inc/Config/QueryContext/HttpQueryContext.php
```diff
@@ -30,7 +90,8 @@ class HttpQueryContext implements QueryContextInterface, HttpQueryContextInterfa
 	public function __construct(
 		private HttpDatasource $datasource,
 		public array $input_schema = [],
-		public array $output_schema = []
+		public array $output_schema = [],
+		private array $config = [],
```
chriszarate, line 94: shouldn't config replace all of these args?

inc/Config/QueryContext/HttpQueryContext.php
```diff
@@ -89,15 +150,16 @@ public function get_datasource(): HttpDatasource {
 	 * Override this method to specify a custom endpoint for this query.
 	 */
 	public function get_endpoint( array $input_variables ): string {
-		return $this->get_datasource()->get_endpoint();
+		$endpoint = $this->config['endpoint'] ?? $this->get_datasource()->get_endpoint();
```
chriszarate, line 153: not a big fan of this logic gate. there should be a singular source

mhsdef: I certainly agree. I see all of this as an interim state caught between our original understanding, "all code", and the growth of the persisted pathes.

I'm just trying to keep it all floating for next week right now.

inc/Config/QueryContext/HttpQueryContext.php
```diff
@@ -180,4 +242,49 @@ public function process_response( string $raw_response_data, array $input_variab
...
+		];
+	}
+
+	/**
+	 * Replace placeholders in the endpoint url with matching input variables.
+	 * 
+	 * The placeholders are expected to be in the Ruby on Rails/Express.js/Phoenix/etc
+	 * placeholder format of :placeholder, eg, `/path/:variable_name`.
+	 */
+	protected function replace_endpoint_placeholders( string $endpoint, array $input_variables ): string {
```
chriszarate, line 280: do we really need this? (maybe, but probably not yet). if we do need it i would prefer we delegate to a library or wordpress function or this is just gonna be a huge headache

inc/Config/QueryContext/HttpQueryContext.php
mhsdef: Yes, but thinking release day + 1.

inc/REST/DatasourceController.php
```diff
@@ -110,15 +111,28 @@ public function create_item( $request ) {
...
-		$code_configured_data_sources = ConfigStore::get_datasources_displayable();
-		$ui_configured_data_sources   = DatasourceCrud::get_data_sources_list();
-		return rest_ensure_response( array_merge( $code_configured_data_sources, $ui_configured_data_sources ) );
+		$ui_configured_data_sources          = DatasourceCrud::get_data_sources_list();
+		$data_sources_from_registered_blocks = ConfigStore::get_datasources_displayable();
+		$merged_data_sources                 = array_merge(
+			$ui_configured_data_sources,
+			$data_sources_from_registered_blocks
+		);
+		$unique_data_sources                 = Utils::remove_duplicates_by_key( $merged_data_sources, 'slug' );
```
chriszarate, line 134: This is temporary right? Eventually we want to work from a central store? (If so can we mark as temporary?

mhsdef: I'm chatting with Shekhar about this right now.

inc/Utils/Utils.php
```diff
@@ -0,0 +1,30 @@
+<?php declare(strict_types = 1);
+
+namespace RemoteDataBlocks\Utils;
+
+class Utils {
+	/**
+	 * Removes duplicates values from an array of arrays bases on the specified key.
```
chriszarate, line 7: Just for clarity, because I can see a bug in my future
```suggestion
	 * Removes duplicates values from an array of arrays bases on the specified key. The first array member with the duplicate value is retained.
```

inc/Config/QueryContext/HttpQueryContext.php
mhsdef: At two working days from release? Yes.

Again, I am approaching this as bridging two paradigms that are not yet aligned not something etched in stone.

inc/Config/QueryContext/HttpQueryContext.php
```diff
@@ -180,4 +242,49 @@ public function process_response( string $raw_response_data, array $input_variab
...
+		$validated = $validator->validate( $config );
+
+		if ( is_wp_error( $validated ) ) {
+			return $validated;
+		}
+
+		return new self( $config['datasource'], $config['input_schema'], $config['output_schema'], $config );
+	}
+
+	public function to_array(): array {
```
chriszarate, line 266: What is the use case for this method? Why doesn't this just return the private $config? And should this be `final`?

mhsdef: It's part of `ArraySerializableInterface`. We could split into serializable/unserializable if we find having both together is a poor assumption, but, currently it's that way because it's a fairly common arrangement and most abstractions can probably conjure some sort of output even if they don't use it much.

Sure, it could return `$config`. I just stuck things that would never be privileged in any way in there for now because there's no immediate use case and I had bigger fish to fry.

inc/Config/QueryContext/HttpQueryContext.php
```diff
@@ -180,4 +242,49 @@ public function process_response( string $raw_response_data, array $input_variab
 	final public function is_response_data_collection(): bool {
 		return $this->output_schema['is_collection'] ?? false;
 	}
+
+	// @todo: consider splitting the datasource injection out from query context so we don't have to tie a query
+	// to a datasource when instantiating. instead, we can just require applying queries to datasources in query
+	// runner execution. ie: $query_runner->execute( $query, $datasource );
```
chriszarate, line 248: I don't think this bodes well for type safety, especially for implementors who provide custom datasources (because they have additional reusable properties they want to inject)

inc/Config/QueryContext/HttpQueryContext.php
chriszarate: Can we just exempt datasource properties from config for now? It would keep the surface area smaller and help us move a bit more carefully

inc/Config/QueryContext/HttpQueryContext.php
mhsdef: Yeah, I'm not saying that's the only way or the best way. I just wanted to make a note because I'm feeling the tension in the current coupling and that thought came to mind.

inc/Config/QueryContext/HttpQueryContext.php
mhsdef: I'm not sure I understand what you mean. I needed this one, `endpoint`, specifically when attempting to transform what we have in the DB into something that executes.

inc/Integrations/Airtable/AirtableDatasource.php
```diff
@@ -96,4 +97,69 @@ public function to_ui_display(): array {
...
+				'type' => $mapping['type'] ?? 'string',
+			];
+		}
+
+		return HttpQueryContext::from_array([
+			'datasource'    => $this,
+			'input_schema'  => [],
+			'output_schema' => $output_schema,
+			'query_name'    => $this->config['tables'][0]['name'],
+			'endpoint'      => $this->get_endpoint() . '/' . $this->config['tables'][0]['id'],
```
chriszarate, line 162: I get it. It would be nice if we could represent 90+% of queries in config. But I don't think we know enough about usage yet. We should just make an `AirtableQuery` class

inc/REST/DatasourceController.php
mhsdef: > Eventually we want to work from a central store? 

Definitely. I think that lift happens right now because it seems like it should be small. Eg: move upstream, read from DB on `ConfigStore` `init`, merge there, keyed, so it is authoritative and naturally deduplicative.

inc/Config/QueryContext/HttpQueryContext.php
chriszarate: We can create an `AirtableGetItemQuery` (keeping the static config changes in this PR, they're great) to implement endpoint and image_url. I don't think the approach to endpoint and image_url in this PR is a shippable solution.

inc/Integrations/Airtable/AirtableDatasource.php
mhsdef: I see that differently. This is working code, the deadline is here, and it's reasonably easy to `rm -rf` it.

inc/Config/QueryContext/HttpQueryContext.php
chriszarate: As mentioned above, I don't think this is a requirement for release

inc/Utils/Utils.php
shekharnwagh: Maybe merging the arrays would be better ? That would still mean one of them would override the other if they have different values for the same key. But that seems unlikely.

mhsdef:
Take two in #124.

PR #123, opened 2024-09-27
Add DatasourceCrud::get_by_slug

mhsdef:
We've determined we want to support this in some form and, for release next week, this--leaning on `slug`s--is the clearest, simplest path.


mhsdef:
No idea how it coopted my title on squash merge with the last commit message on this one...  :|

PR #124, opened 2024-09-27
Make Airtable loading from DB work

mhsdef:
Same idea as #118 but use concrete classes for the airtable queries.

## How to Test

1. Add an Airtable datasource via settings.
2. Add some fields.
3. Add blocks.


inc/Integrations/Airtable/AirtableDataSource.php
```diff
@@ -96,4 +96,67 @@ public function to_ui_display(): array {
 			'uuid'    => $this->config['uuid'] ?? null,
 		];
 	}
+
+	public function ___temp_get_query(): AirtableGetItemQuery|\WP_Error {
```
maxschmeling, line 100: Is the naming here to indicate this isn't how we want this API to look in the future?

mhsdef: Yep.

The query data this needs is currently being stored inside the data source data inside the `remote_data_blocks_config` options value and I think, AFAIK, we are all in agreement we should split all that apart as soon a we can manage.

maxschmeling (approved):
This seems to work well. The only issue I'm seeing which we can look at later is for fields that are from other tables, it doesn't seem to bind those correctly. In this case, it's the `Locations` field on the `Maps` table:
image

PR #125, opened 2024-09-28
Improve local development docs and start script

chriszarate:
- [Improve start and stop commands](https://github.com/Automattic/remote-data-blocks/commit/3292fa3b7eb798a9a9852502929e32ebbab22dd4) (no functional changes)
  - Better messaging
  - More consistent cleanup
  - Fix issues flagged by [Shellcheck](https://www.shellcheck.net/)
  - Fix race condition that sometimes occurs by creating object cache first
- [Update example API label to match documentation](https://github.com/Automattic/remote-data-blocks/commit/d7f4e8df8bc502e86c17b677105a0a71bd3c86fc)
- [Improve local development documentation](https://github.com/Automattic/remote-data-blocks/commit/aff662ae8e5226c349644946e9b0525589c326c1)
- [Add local playground blueprint](https://github.com/Automattic/remote-data-blocks/commit/562f905ae25ec6a66c72d70025d648f0db1deaad)
  - As documented with strong wording, this is not intended to be a development environment. But it is extremely useful to quickly validate changes.
  - My primary motivation, though, is to help users sidestep a footgun. Pointing `@wp-now` to `blueprint.json` can actually result in the current folder being wiped out. It's how I lost a lot of uncommitted progress on docs earlier this week. :/

Question: Is "monolith" a confusing term since we aren't shipping with documented decoupled capabilities? Should we make it the default for `npm run start`?


jblz (approved):

jblz:
> Question: Is "monolith" a confusing term

IMO, yes.

> Should we make it the default for npm run start?

IMO, also yes.

maxschmeling (approved):

maxschmeling:
> > Question: Is "monolith" a confusing term
> 
> IMO, yes.
> 
> > Should we make it the default for npm run start?
> 
> IMO, also yes.

Yes, I think defaulting to monolith would make sense.

I'm not sure if it's confusing or not. I do think it probably makes sense to just assume that our default environment is "monolith" and not necessarily talk about it as "equal" to a decoupled environment.

PR #126, opened 2024-09-30
Add integration template workflow doc

mhsdef:
Walk through creating a zip code block via an integration template.

I'm open to terminology other than "integration template" if anyone has a better idea for something that:

- has a query
- has registration code
- doesn't have a data source (initially)

I do think this thing needs a name of some sort though.


docs/workflows/zip-code-integration-template.md
```diff
@@ -0,0 +1,116 @@
+# Integration templates
+
+Developers can [extend](index.md) the plugin to meet uses cases that require custom logic. Often, the full details are known up front to the developer, in which case they can simply write a complete implementation.
+
+However, sometimes the developer may not know the final data source details.
+
+Integration templates bridge the gap between developers and WordPress instance admins when implementing custom data fetching solutions. These templates support scenarios where developers need to create custom blocks that interact with third-party APIs, but may not have full access to authentication details managed by other teams.
```
chriszarate: There are other (probably more likely?) scenarios here, for example:

1. Variable data sources, concrete queries. I commit my queries to code, but I define different data sources in the UI for my production vs staging vs local environments. 
2. Supported data sources, unsupported queries. I like using UI to define data sources, so I do so. However, RDB doesn't yet support generating queries automatically for my data source, so I must commit code.
   - Or, I require code-level customization to my queries that requires committing queries to code.

chriszarate (approved):

chriszarate:
> I'm open to terminology other than "integration template"

I'm bad at naming things but this term was a little confusing to me—mostly because we don't use "integration" anywhere else. I have been thinking of it as a "mixed UI and code path." What we're describing is a way to link code and UI. Is there a way to encode that in the name?

mhsdef:
> What we're describing is a way to link code and UI. Is there a way to encode that in the name?

That is what it is attempting to do. Perhaps poorly, I'm not particularly attached to it.

> we don't use "integration" anywhere else

We do have `Integrations` in the code. The thinking there was that they are a follow-on from what we are calling these things on the VIP side so we could/should follow that precedent.

maxschmeling:
I don't really view it as "mixing" all that much. I just view the data source config in the settings page as "fancy environment variables" more or less.

docs/workflows/zip-code-integration-template.md
maxschmeling: Yeah, I think we want to frame this as "You write your code, then provide your variables through the settings page" and the alternative of creating your own very custom data source is kind of "hidden" in "Advanced Usage" or something only when you really need to use a very specific configuration setup.

I don't think providing connection details in the settings page feels like a "mixed" or "separated" approach... it's just making a few key details into variables.

maxschmeling: So I guess what I'm saying is it might be better to surface this doc as the "primary" flow and label the "fully code based" one where you provide your own credentials in the code as the "advanced usage" or secondary flow.

Does that make sense?

docs/workflows/index.md
```diff
@@ -4,4 +4,5 @@
 
 ## Code-based workflows
 
-- [Create an Airtable remote data block with code](workflows/airtable-with-code.md)
+- [Create an Airtable integration with code](workflows/airtable-with-code.md)
+- [Create a Zip Code integration template with code](workflows/zip-code-integration-template.md)
```
maxschmeling:
```suggestion
- [Create a Zip Code integration template with code](workflows/zip-code-integration-template.md)
- Advanced usage: [Create an Airtable integration with code](workflows/airtable-with-code.md)
```

maxschmeling: Something like this is what I was thinking... thoughts?

I'm open to the idea that we shouldn't frame anything as Standard / Advanced right now since we don't really know the customer usage yet... but that's what I think of them as.

chriszarate: I like it. Also totally fine 👍 🚢  with any docs PR, we will go through plenty of revisions and reorgs later, just need words at this point

mhsdef:
> I don't really view it as "mixing" all that much. I just view the data source config in the settings page as "fancy environment variables" more or less.

Hmm. Interesting framing. I can see that angle.

PR #128, opened 2024-09-30
fix google sheet data source add API

shekharnwagh:
# Description

- Fixes the Google Sheets data source save operation was failing due to the `GoogleSheetsDataSource` schema not being in sync with the schema used in data sources settings form.
- Filter out empty meta values from before being displayed in the settings list view. This can happen due to the data sources coming from the registered block for whom the data sources are configured in code.

### DIFF
--- a/inc/Integrations/Google/Sheets/GoogleSheetsDataSource.php
+++ b/inc/Integrations/Google/Sheets/GoogleSheetsDataSource.php
@@ -13,7 +13,7 @@ class GoogleSheetsDataSource extends HttpDataSource {
 	protected const SERVICE_SCHEMA = [
 		'type'       => 'object',
 		'properties' => [       
-			'credentials'    => [
+			'credentials'  => [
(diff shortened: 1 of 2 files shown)


mhsdef:
> Filter out empty meta values from before being displayed in the settings list view. This can happen due to the data sources coming from the registered block for whom the data sources are configured in code.

What's an example of this?

mhsdef (approved):

shekharnwagh:
> > Filter out empty meta values from before being displayed in the settings list view. This can happen due to the data sources coming from the registered block for whom the data sources are configured in code.
> 
> What's an example of this?

When we create the `GoogleSheetsDataSource` object from code config - https://github.com/Automattic/remote-data-blocks/blob/b72e67f1a6f78fe36b1fd5c41207c1b4975b1d6a/inc/Integrations/Google/Sheets/GoogleSheetsDataSource.php#L95-L110

PR #129, opened 2024-09-30
fix google sheet data source get_endpoint failing

shekharnwagh:
The get_endpoint was failing due to spreadsheet_id being incorrectly accessed from config.


mhsdef (approved):

mhsdef:
Still LGTM! And thanks for the tests! 

I do think it'd be a good discussion for all of us sometime on the amount of static classes we have going. In my experience, PHP wants to PHP so the happy path is instantiations and some amount of OOP.

But that's beyond this and for another day. 🚢

PR #133, opened 2024-10-01
Specify WordPress and PHP versions in .wp-env.json

chriszarate:
I think I'm going to advocate to close this PR (that I just opened). 😂  Riding `trunk` has its benefits and lets us see the breakings on the horizon.

### DIFF
--- a/.wp-env.json
+++ b/.wp-env.json
@@ -1,4 +1,6 @@
 {
+  "core": "WordPress/WordPress#6.6.2",
+  "phpVersion": "8.2",
   "config": {
     "WP_DEBUG_DISPLAY": true,
     "WP_DEVELOPMENT_MODE": "plugin",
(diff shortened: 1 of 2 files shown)


mhsdef:
Yeah, I think us being out in front is probably a good thing.

It defaults to the latest production version btw. If we want `trunk` specifically, I think we need to do the `#trunk` thing.

chriszarate:
> It defaults to the latest production version btw. If we want trunk specifically, I think we need to do the #trunk thing.

Thanks for the clarification. So do we want `trunk`? We don't have e2e tests (yet?) so we don't have another mechanism to test new versions.

maxschmeling:
I think staying on the latest released version is best for now. We definitely need to be figuring out how to test and maintain proper version support once we release 1.0

PR #145, opened 2024-10-01
Fix Shopify logo

chriszarate:
Screenshot 2024-10-01 at 16 00 23


maxschmeling (approved):

maxschmeling:
I don't love the alignment, but that's how it was... this PR does its job. 👍 thanks

PR #146, opened 2024-10-02
Adds pre-commit hook

mehmoodak:
Run linting commands before every commit and only on changed files.


chriszarate:
Does `pre-commit` run on every commit when rebasing? Would `pre-push` be better?

mehmoodak:
> Does pre-commit run on every commit when rebasing?

Tested it by rewording one of the commit during rebase and the `pre-commit` hook didn't run.

If that's the case then I think `pre-commit` is better than `pre-push`.

**Edit:** It's running while doing `edit` during rebase. Looking into this.

mehmoodak:
@chriszarate  added the logic to avoid running `pre-commit` hook during rebase. Please take a look now, thanks.

jblz:
> Would pre-push be better?

In general, I find pre-push to be a better devex (at the expense of some interim commits that do not adhere to our coding standards).

Is there another advantage to pre-commit that we gain that I'm not thinking of?

mehmoodak:
> Is there another advantage to pre-commit that we gain that I'm not thinking of?

I think it's better to run these lints/static analysis on every commit. I was concerned about speed, so I've ensured that we only run linting commands on staged files.

If there's a strong preference for running it on `pre-push` instead, let me know and I will change it otherwise we can also revisit the decision later.

chriszarate:
> I think it's better to run these lints/static analysis on every commit.

Why? What benefits will this provide that `pre-push` will not? I haven't used git hooks that much.

mehmoodak:
I prefer to fix issues during each commit and keep each commit clean, rather than doing it at the push level.

I've tried to make the `pre-commit` hook as non-intrusive as possible, but if you think `pre-push` is a better approach, I'm open to switching.

chriszarate:
> I prefer to fix issues during each commit and keep each commit clean

There are situations (like when refactoring or moving files in isolated commits) where that feels like a burden, but I guess we have `--no-verify` available.

I'm willing to give it a shot and we can always change it.

chriszarate (approved):

maxschmeling:
I do tend to prefer it on `pre-push` as well, but like @chriszarate said, I'm up for giving it a try like this and adjusting down the road if we decide to.

PR #147, opened 2024-10-03
Add data source repository

mhsdef:
I gave this some thought early last week, and, IMO our data source interplay shakes out like this abstractly:

- Data Source Providers (code, wpdb/ui, and eventually VIP Integrations)
- Data Source Writers (wpdb/ui only)
- Data Source Repository

That last one, the repository, is the single place to go to read/write from that we seek. It should encapsulate the underlying complexity so nothing higher up needs to think about it.

Our data needs, being multi-sourced, are complex enough that I suggest we try _light_ usage of the [repository pattern](https://doeken.org/blog/repository-pattern). We feel like a good candidates in that:

- we have multiple sources that we have to juggle
- we aren't using an ORM so we have to do _something_ to contain the complexity

And the pattern, done well, makes data flow and testing a breeze.


mhsdef:
So, init would be in the main plugin file, something like:

```
Data\DataSourceRepository::init(
  [
    Config\ConfigStore::class,
    DataSourceCrud::class
    // eventually VipIntegrationGoop:class or whatever
  ],
  [
    DataSourceCrud::class
  ]
)
```

There is a bit of an impedence mismatch at the moment between our static class usages and that this class may need instantiation to use in reality. FWIW, suggest we consider moving that way. PHP is happier path with instantiations and dependency injections in most cases.

inc/Data/DataSourceRepository/DataSourceRepositoryInterface.php
```diff
@@ -0,0 +1,37 @@
+<?php declare(strict_types = 1);
+
+namespace RemoteDataBlocks\Data;
+
+use RemoteDataBlocks\Config\Datasource\DatasourceInterface;
+
+defined( 'ABSPATH' ) || exit();
+
+interface DataSourceRepositoryInterface {
```
mhsdef, line 9: Inject this instead of a concrete `DataSourceRepository` into all the things that need it. Controllers, whatever. Then in testing you just pass a test class that implements the interface and you're golden.

PR #151, opened 2024-10-07
Improve local development scripts

chriszarate:
A collection of improvements to the local development scripts. Feedback and suggested changes welcome!

- As noted in https://github.com/Automattic/remote-data-blocks/pull/125, the term "monolith" can be confusing and is better assumed as the default. The command to start a local environment is now `npm run dev` or `npm run dev:decoupled`.
- Enable Xdebug by default.
- Instead of maintaining two different `.wp-env.json` files, we can consolidate to one and manage any differences (options, plugins) using WP-CLI.
- Print URL and credentials in the console. Open WordPress automatically if user has set `WP_ENV_OPEN_BROWSER`.
- Better messaging overall in console.
- Organize the scripts into Bash functions for readability and reuse.
- Set Webpack output to minimal. It will still print detailed warnings and errors when present, otherwise gets out of your way.
- Set `WP_REDIS_DISABLE_BANNERS=true` to suppress Redis Object Cache Pro nags.
- Add JSON schema ref for `.wp-env.json`.

Script output now looks like this:

Screenshot 2024-10-07 at 11 53 09


shekharnwagh (approved):
Thanks for improving the dev scripts. I tried the start and destroy commands—they worked!

jblz (approved):
Groovy

PR #155, opened 2024-10-09
Adds Tracks Analytics on backend

mehmoodak:
### Description

Adds Tracks Analytics by using the Telemetry library available in MU Plugins which will only work if site is hosted on VIP Platform or if the analytics is enabled via filter.

### How to test?

1. Create a site using `vip dev-env` which will have the MU Plugins.
1. Move `remote-data-blocks` plugin to the `plugins` folder.
1. Add filter in `plugin-loader.php` to enable tracking.

    ```php
    add_filter( 'remote_data_blocks_enable_tracks_analytics', '__return_true' );
    ```
1. Comment [these lines](https://github.com/Automattic/remote-data-blocks/blob/78c5cd5a7fdb345b5825932551ca8fc738ac6927/inc/Analytics/TracksAnalytics.php#L32C1-L35C4) to allow tracking on `local` env.
1. Trigger track events, verify that the events are logged on Live View without any error.


tests/inc/stubs.php
```diff
@@ -17,7 +17,15 @@ function esc_html( string $text ): string {
 	return $text;
 }
 
-function register_block_pattern( string $name, array $options ): void {
+function register_block_pattern( string $_name, array $_options ): void {
```
mehmoodak: psalm was complaining about unused parameters so prefixed them with `_`

remote-data-blocks.php
```diff
@@ -34,16 +36,19 @@
 Editor\BlockManagement\ConfigRegistry::init();
 Editor\PatternEditor\PatternEditor::init();
 
-// Example API
+// Analytics.
+$GLOBALS['rdb_tracks'] = new TracksAnalytics();
```
mehmoodak: - Avoided static methods on class because of issues during mocking.
- Used global here to avoid instantiation before recording every event.

inc/Config/DataSource/HttpDataSource.php
```diff
@@ -36,10 +36,17 @@ public function get_image_url(): ?string {
 		return null;
 	}
 
+	/**
+	 * Get the service name.
+	 */
+	public function get_service(): ?string {
+		return isset( $this->config['service'] ) ? $this->config['service'] : null;
```
chriszarate: fyi PHP has nullish coalescing since 7.0
```suggestion
		return $this->config['service'] ?? null;
```

psalm.xml
```diff
@@ -20,7 +20,10 @@
         <PossiblyUnusedMethod errorLevel="suppress"/>
     </issueHandlers>
     <stubs>
+        <file name="vendor/php-stubs/wordpress-globals/wordpress-globals.php"/>
         <file name="vendor/php-stubs/wordpress-stubs/wordpress-stubs.php"/>
+        <file name="vendor/php-stubs/wordpress-stubs/wordpress-stubs.php"/>
```
chriszarate: duplicate
```suggestion
```

remote-data-blocks.php
chriszarate: We should solve these issues instead of introducing a global

mehmoodak: Issue resolved. Thanks for your help.

inc/Analytics/EnvironmentConfig.php
```diff
@@ -0,0 +1,81 @@
...
+	 * Tracks analytics core properties.
+	 * 
+	 * This is set by the Tracks library available in MU Plugins.
+	 *
+	 * @var array<string, mixed>
+	 */
+	private array $tracks_core_props = [];
+
+	public function __construct() {
+		if ( function_exists( 'Automattic\VIP\Telemetry\Tracks\get_tracks_core_properties' ) ) {
```
chriszarate, line 26: Does this work? I would expect that backslashes need to be escaped
```suggestion
		if ( function_exists( 'Automattic\\VIP\\Telemetry\\Tracks\\get_tracks_core_properties' ) ) {
```

inc/Analytics/EnvironmentConfig.php
```diff
@@ -0,0 +1,81 @@
...
+			} );
+		}
+	}
+
+	public function is_enabled_via_filter(): bool {
+		return apply_filters( 'remote_data_blocks_enable_tracks_analytics', false ) ?? false;
+	}
+
+	public function get_tracks_lib_class(): ?string {
+		if ( ! class_exists( 'Automattic\VIP\Telemetry\Tracks' ) ) {
```
chriszarate, line 39: ?
```suggestion
		if ( ! class_exists( 'Automattic\\VIP\\Telemetry\\Tracks' ) ) {
```

inc/Analytics/EnvironmentConfig.php
```diff
@@ -0,0 +1,81 @@
...
+
+	public function is_local_env(): bool {
+		if ( ! isset( $this->tracks_core_props['vipgo_env'] ) ) {
+			return false;
+		}
+
+		return 'local' === $this->tracks_core_props['vipgo_env'];
+	}
+
+	public function is_remote_data_blocks_plugin( string|null $plugin_path ): bool {
```
chriszarate, line 62: ideally these methods are `private` and unit tests target the public methods

inc/Analytics/TracksAnalytics.php
```diff
@@ -0,0 +1,170 @@
...
+		self::record_event( 'remotedatablocks_plugin_toggle', [ 'action' => 'deactivate' ] );
+	}
+
+	/**
+	 * Track usage of Remote Data Blocks.
+	 *
+	 * @param int      $post_id Post ID.
+	 * @param \WP_Post $post Post object.
+	 */
+	public static function track_remote_data_blocks_usage( int $post_id, object $post ): void {
```
chriszarate, line 97:
```suggestion
	public static function track_remote_data_blocks_usage( int $post_id, \WP_Post $post ): void {
```

inc/Analytics/EnvironmentConfig.php
mehmoodak: Yes it works, seems like we don't need escaping in this case (using same syntax in MU Plugins also).

inc/Analytics/TracksAnalytics.php
mehmoodak: The `object` type is used intentionally here, as using `WP_Post` directly in tests would require a tricky approach.

inc/Analytics/EnvironmentConfig.php
mehmoodak: Sorry didn't get you. We do require these methods as `public` so that they can be accessed in `TracksAnalytics` class.

tests/inc/Analytics/EnvironmentConfigTest.php
```diff
@@ -0,0 +1,77 @@
...
+
+	public function testIsWpvipSiteReturnsFalse(): void {
+		$obj = new EnvironmentConfig();
+
+		$this->assertEquals( false, $obj->is_wpvip_site() );
+	}
+
+	public function testIsWpvipSiteReturnsTrue(): void {
+		$obj = new EnvironmentConfig();
+		set_private_property( EnvironmentConfig::class, $obj, 'tracks_core_props', [ 'hosting_provider' => 'wpvip' ] );
```
maxschmeling, line 34: Do we absolutely need these private property methods?

Can we mock/stub out `get_tracks_core_properties` instead of setting `tracks_core_props` directly?

maxschmeling (approved):
My only issue is with the private member stuff in the tests. If you think it's really necessary, I'm ok with it. But it seems like there could be a better way.

tests/inc/Analytics/EnvironmentConfigTest.php
mehmoodak: We're assigning `track_core_props` inside the `init` action hook, so using `set_private_property` seems okay given that we're not running tests within the WP environment.

PR #163, opened 2024-10-17
Fix image block binding

maxschmeling:
At some point the `alt` and `url` attributes for our `core/image` block bindings got renamed to `image_alt` and `image_url`.  These aren't the correct bindings, so binding the image block wasn't working. This fixes that.

### DIFF
--- a/inc/Integrations/Airtable/AirtableDataSource.php
+++ b/inc/Integrations/Airtable/AirtableDataSource.php
@@ -45,7 +45,10 @@ class AirtableDataSource extends HttpDataSource {
 								'type'       => 'object',
 								'properties' => [
 									'name' => [ 'type' => 'string' ],
-									'type' => [ 'type' => 'string', 'required' => false ],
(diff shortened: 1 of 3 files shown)


inc/Integrations/Airtable/AirtableDataSource.php
```diff
@@ -117,7 +121,7 @@ public function ___temp_get_query(): AirtableGetItemQuery|\WP_Error {
 		];
 
 		foreach ( $this->config['tables'][0]['output_query_mappings'] as $mapping ) {
-			$output_schema['mappings'][ strtolower( $mapping['name'] ) ] = [
+			$output_schema['mappings'][ $mapping['name'] ] = [
```
chriszarate:
```suggestion
			$output_schema['mappings'][ ucfirst( $mapping['name'] ) ] = [
```

chriszarate (approved):

PR #185, opened 2024-10-30
Add Tracks Analytics on frontend

mehmoodak:
### Description

Add Tracks Analytics by using the `@automattic/calypso-analytics` package. With each track event, the following properties will be tracked:

- Properties related to VIP Platform (e.g. org_id, env_id).
- Properties related to Remote Data Blocks (e.g. plugin_version).
- Event specific properties.

**Note:** This tracking will only work if site is hosted on VIP Platform or if the analytics is enabled via filter.

### How to test?

1. Create a site using `vip dev-env` which will have the MU Plugins.
1. Move `remote-data-blocks` plugin to the `plugins` folder.
1. Add filter in `plugin-loader.php` to enable tracking.

    ```php
    add_filter( 'remote_data_blocks_enable_tracks_analytics', '__return_true' );
    ```
1. Checkout [this wp-calypso PR](https://github.com/Automattic/wp-calypso/pull/96007), `build` and `link` the `@automattic/calypso-analytics` package.
1. To allow tracking on local env comment check on [PHP](https://github.com/Automattic/remote-data-blocks/blob/1574d6f729327eda7a10ecfa686c58a3252f8187/inc/Analytics/TracksAnalytics.php#L30C1-L34C4) and [JS](https://github.com/Automattic/remote-data-blocks/blob/1574d6f729327eda7a10ecfa686c58a3252f8187/src/blocks/remote-data-container/utils/tracks.ts#L15C1-L19C3) side.
1. Trigger track events, use "Live View" to verify that the events are getting logged.


types/globals.d.ts
```diff
@@ -1,6 +1,22 @@
...
 	var REMOTE_DATA_BLOCKS: LocalizedBlockData | undefined;
 	var REMOTE_DATA_BLOCKS_SETTINGS: LocalizedSettingsData | undefined;
+
+	/**
+	 * Generic Tracks properties sent from MU Plugins.
+	 */
+	var VIP_TRACKS_BASE_PROPS: VipTracksBaseProps | undefined;
+}
+
+export interface VipTracksBaseProps {
```
chriszarate: Put this in a separate file like `tracks.d.ts`

src/blocks/remote-data-container/utils/tracks.ts
```diff
@@ -0,0 +1,26 @@
+import { recordTracksEvent } from '@automattic/calypso-analytics';
+
+import { getTracksBaseProps } from '@/utils/localized-block-data';
+
+/**
+ * Send a tracks event with the given name and properties.
+ */
+export function sendTracksEvent( eventName: string, props: Record< string, unknown > ): void {
+	const vipBaseProps = window.VIP_TRACKS_BASE_PROPS;
```
chriszarate: Is a global necessary? Can the Tracks library expose a way to register these base props?

.husky/pre-commit
```diff
@@ -1,6 +1,7 @@
 if [ -d "$(git rev-parse --git-path rebase-merge)" ] || [ -d "$(git rev-parse --git-path rebase-apply)" ]; then
-  echo "Rebase in progress. Skipping pre-commit hook.\n";
-  exit 0;
+  echo "Rebase in progress. Skipping pre-commit hook.\n"
+  exit 0
 fi
 
 lint-staged
+npm run check-types # Run tsc on the whole project, as running it only on staged files may result in import or type errors.
```
chriszarate: Probably better to think of type checking as something different from "linting" and remove it from scripts with that name.

src/blocks/remote-data-container/utils/tracks.ts
mehmoodak: This global is used to pass information from PHP to the frontend. I did some search about a better way to pass this information but couldn't find anything.

inc/Editor/BlockManagement/BlockRegistration.php
```diff
@@ -77,6 +78,7 @@ public static function register_blocks(): void {
 				'availableBindings' => $available_bindings,
 				'loop' => $config['loop'],
 				'name' => $block_name,
+				'dataSource' => ConfigStore::get_data_source( $block_name ),
```
chriszarate: Is this trying to describe the type of data source, like Airtable or Shopify? Maybe we can name it `dataSourceType`

inc/REST/RemoteDataController.php
```diff
@@ -77,6 +77,7 @@ public static function execute_query( WP_REST_Request $request ): array|WP_Error
 		return array_merge(
 			[
 				'block_name' => $block_name,
+				'data_source' => ConfigStore::get_data_source( $block_name ),
```
chriszarate: same here

src/blocks/remote-data-container/components/BlockBindingControls.tsx
```diff
@@ -68,6 +84,11 @@ export function BlockBindingControls( props: BlockBindingControlsProps ) {
 			? Object.entries( availableBindings ).find( ( [ key ] ) => key === contentField )?.[ 1 ]?.name
 			: undefined;
 		updateBinding( 'content', { ...contentArgs, field: contentField, label } );
+		sendTracksEvent( 'remotedatablocks_remote_data_container_actions', {
+			action: 'show_label',
```
chriszarate: `toggle_label`?

src/blocks/remote-data-container/components/BlockBindingControls.tsx
```diff
@@ -68,6 +84,11 @@ export function BlockBindingControls( props: BlockBindingControlsProps ) {
...
+			data_source: getBlockDataSource( remoteDataName ),
+			value: showLabel,
```
chriszarate, line 90: could probably be boolean

src/blocks/remote-data-container/components/BlockBindingControls.tsx
chriszarate: or `action: showLabel ? 'show_label' : 'hide_label'`

src/blocks/remote-data-container/components/field-shortcode/FieldShortcodeButton.tsx
```diff
@@ -76,11 +79,20 @@ export function FieldShortcodeButton( props: WPFormatEditProps ) {
 	function onSelectField( data: FieldSelection, fieldValue: string ) {
 		updateOrInsertField( data, fieldValue );
 		onClose();
+		sendTracksEvent( 'remotedatablocks_field_shortcode', {
+			action: data.action,
+			data_source: getBlockDataSource( data.remoteData?.blockName ),
```
chriszarate: seems like looking up the data source type should be delegated to `sendTracksEvent`

src/blocks/remote-data-container/components/modals/InputModal.tsx
```diff
@@ -25,9 +28,14 @@ export function InputModal( props: InputModalProps ) {
 		setInputState( { ...inputState, [ field ]: value } );
 	}
 
-	function wrappedOnSelect(): void {
+	function onSaveClick(): void {
```
chriszarate:
```suggestion
	function onSelectItem(): void {
```

src/blocks/remote-data-container/filters/withBlockBinding.tsx
```diff
@@ -66,6 +66,7 @@ function BoundBlockEdit( props: BoundBlockEditProps ) {
 						attributes={ attributes }
 						availableBindings={ availableBindings }
 						blockName={ blockName }
+						remoteDataName={ remoteDataName }
```
chriszarate, line 69: Instead of prop drilling the container block name, I think we should just prop drill the `sendTracksEvent` function with the data_source_type baked in

src/blocks/remote-data-container/utils/tracks.ts
chriszarate: I understand, I just don't think mu-plugins should be responsible for this

mehmoodak: Agreed. The main reason of passing this information from MU Plugins is to have a consistent implementation across all of VIP.

> Can the Tracks library expose a way to register these base props?

By registering you mean?

src/blocks/remote-data-container/components/field-shortcode/FieldShortcodeButton.tsx
mehmoodak: We don't have the `data_source` prop on all events, so I didn't delegate it to `sendTracksEvent`.

src/blocks/remote-data-container/utils/tracks.ts
chriszarate: > The main reason of passing this information from MU Plugins is to have a consistent implementation across all of VIP.

I am only objecting to the JS global as the method of passing the base props. Currently we are injecting it into the `<head>` which is not ideal. Providing it as a global also provides little value, since each plugin still needs custom code (and types) to access the global and inject it. I would rather avoid the JS global and script in `<head>`, have the plugins load the base props via PHP, and then inject it into their own global (e.g., `REMOTE_DATA_BLOCKS_SETTINGS` for this plugin). Alternatively plugins could integrate via a REST endpoint or a custom library.

> By registering you mean?

Another approach would be to render the base props as meta tag(s) and have the client-side code read them (once) and include them on each event.

src/blocks/remote-data-container/components/field-shortcode/FieldShortcodeButton.tsx
chriszarate: Can it just be undefined in those cases?

types/globals.d.ts
mehmoodak: https://github.com/Automattic/remote-data-blocks/pull/185/commits/be36ae58a7d2ef1d97a1d4591a620de7ac137190

src/blocks/remote-data-container/utils/tracks.ts
mehmoodak: > and then inject it into their own global (e.g., REMOTE_DATA_BLOCKS_SETTINGS for this plugin)

I will implement it in `REMOTE_DATA_BLOCKS_SETTINGS` const.

> Alternatively plugins could integrate via a REST endpoint or a custom library.

I have initially suggested it in P2 and Mo feedback on this is that using the REST API is more expensive then making the calls directly to Tracks from the client.

> Another approach would be to render the base props as meta tag(s) and have the client-side code read them (once) and include them on each event.

It's more parsing, which I think is better to avoid.

inc/Editor/BlockManagement/BlockRegistration.php
mehmoodak: https://github.com/Automattic/remote-data-blocks/pull/185/commits/a3e42f9527602ce6ac3f10a2fb44e0ea57e4c442

inc/REST/RemoteDataController.php
mehmoodak: Removed it. I was previously passing the `data_source` prop, but now I'm using the block name to retrieve the data source instead.

src/blocks/remote-data-container/components/BlockBindingControls.tsx
mehmoodak: https://github.com/Automattic/remote-data-blocks/pull/185/commits/bf3289a1e22f6cff137c9504d48c7d29e8a877d9

src/blocks/remote-data-container/components/field-shortcode/FieldShortcodeButton.tsx
mehmoodak: It can be `undefined`, but we still need to pass `blockName`, so I think it's better to pass `data_source`, which is the value we ultimately send to `Tracks` anyway.

By the way, I've updated the types for these Tracks Events. (https://github.com/Automattic/remote-data-blocks/pull/185/commits/a14dfce570d4075348f37c5246b26fc89b93785c)

src/blocks/remote-data-container/components/modals/InputModal.tsx
mehmoodak: https://github.com/Automattic/remote-data-blocks/pull/185/commits/cd8bfc1a8d90312e75aa033dc1dd8cc20a08084c

src/blocks/remote-data-container/filters/withBlockBinding.tsx
mehmoodak: We'll need prop drilling in either case, to me passing `blockName` seems simpler than passing the function.

.husky/pre-commit
mehmoodak: https://github.com/Automattic/remote-data-blocks/pull/185/commits/7ce75c6689953bc0eb331798e4bd2bc6a7c39717

src/blocks/remote-data-container/utils/tracks.ts
mehmoodak: > I will implement it in `REMOTE_DATA_BLOCKS` const.

Done. https://github.com/Automattic/remote-data-blocks/pull/185/commits/44a3a22af29ccd0e0eead8489809e8bfc9667723

tsconfig.json
```diff
@@ -18,5 +18,5 @@
     "target": "es6",
     "typeRoots": [ "./types", "./node_modules/@types" ]
   },
-  "include": [ "./types/*.d.ts", "./example/**/*", "./src/**/*", "./tests/**/*" ]
+  "include": [ "./**/*.d.ts", "./example/**/*", "./src/**/*", "./tests/**/*" ]
```
maxschmeling: Was this needed for something specific?

maxschmeling (approved):
Looks like a really good start. Some minor pedantic things I might do a little different, but  not worth nit picking. I like it.

src/blocks/remote-data-container/components/BlockBindingControls.tsx
```diff
@@ -51,11 +55,23 @@ export function BlockBindingControls( props: BlockBindingControlsProps ) {
 	function updateFieldBinding( target: string, field: string ): void {
 		if ( ! field ) {
 			removeBinding( target );
+			sendTracksEvent( 'remotedatablocks_remote_data_container_actions', {
+				action: 'remove_binding',
+				data_source_type: getBlockDataSourceType( remoteDataName ),
```
maxschmeling, line 60: Should this also have a `remote_data_field` property?

tsconfig.json
mehmoodak: Reverted. I was trying to fix some TS issue but this change is not needed now.

src/blocks/remote-data-container/components/BlockBindingControls.tsx
mehmoodak: For `remove_binding` action we don't get `field` value and that's why I haven't included `remote_data_field` in track props. Let me know if you think we should include it, I will find an alternative.

PR #191, opened 2024-11-07
github markdown block link fixes

shekharnwagh:
# Description

> [!IMPORTANT]  
> Depends on PR #198 

This PR fixes the broken internal links to other markdown docs when using "GitHub File As HTML" block.

## Changes

- Remove unused patterns for "GitHub File As HTML" block
- Allow slashes in overrides in url path when there is just one input variable. Earlier the overrides were delimited by slashes (`/`). For example in url `page_slug/$first/$second` `$first` and `$second` are assigned to first and second input variables respectively. This was a blocker in using the GitHub Doc Markdown file path as url overrides since the file path contains slashes. Now if there is just one input variable, all the url overrides are treated as just one input variable. For example `page_slug/$dir/$path/$file` will assign `$dir/$path/$file` to the input variable.
- If input variables are missing file extension then its added to allow having url overrides without the `.md` suffix
- Fix the internal links in the GitHub File As HTML block by replacing all the href attributes with local links ending in `.md`.

# Testing

- Start the local server using by following the instructions in the [Docs](https://github.com/Automattic/remote-data-blocks/blob/trunk/docs/local-development.md).
- Navigate to the page created in "GitHub File As HTML" example at the url `http://localhost:8888/gh`.
- Edit the page to add a "GitHub File As HTML" block and set the url overrides to `/gh/${file_path}`.
- Pass the file path of the markdown file in url like `http://localhost:8888/gh/${file_path}` (e.g. `http://localhost:8888/gh/docs/workflows/google-sheets-with-code`) and verify that the docs are rendered correctly and the internal links are working.


example/github/remote-data-blocks/inc/queries/class-github-get-file-as-html-query.php
shekharnwagh, line 1: @maxschmeling What do you think about extracting the query classes to shared code in `inc/Integrations/Github` directory ? GitHub would have a static response format like Shopify so this is a good candidate for this, but the current implementation is also somewhat specific to markdown.

shekharnwagh: The logic to make it work without the `.md` in the path could go here - https://github.com/Automattic/remote-data-blocks/blob/bec543b65bdbb6fe68591a5a107dcdbdae451e75/example/github/remote-data-blocks/inc/queries/class-github-get-file-as-html-query.php#L70 in this file where we could append the `.md` if it is absent in the path in the input variables. We would need to make it generic though as this query is generic for GitHub and specific to just a markdown file.

maxschmeling: I don't *think* we should move this to the integrations folder yet, though it might make sense to do that at some point. I think I'd want to do more with it rather than something that is narrowly scoped to what we're using it for.

maxschmeling (commented):
I'm wondering if this parser logic belongs in the plugin or if it should go in custom configuration in the site code. I just wonder if it's generic enough. Thoughts?

example/github/remote-data-blocks/inc/queries/class-github-get-file-as-html-query.php
shekharnwagh: I've moved the markdown parsing/url update logic here in the query file and removed the GitHubParser file from integrations DIR.

maxschmeling (approved):

PR #195, opened 2024-11-13
Adjust data source slug input

brookewp:
The data source slug can only contain lowercase alphanumeric characters and hyphens, so currently it isn't possible to enter anything else into the input field, including capital letters. 

We already use the method `toLowerCase` in the `sanitizeDataSourceSlug` function, so we can adjust the regex to be case insensitive to allow for capitals to be entered before sanitizing. 

This PR also moves the logic from the input's `onChange` to `onBlur`, so there is visual feedback showing capitals are being entered, but when leaving the field, it'll be adjusted to lower case: 

slug

In the future, we may want to consider adding some warning to let the user know these characters are invalid for this field.

### DIFF
--- a/src/data-sources/components/SlugInput.tsx
+++ b/src/data-sources/components/SlugInput.tsx
@@ -2,7 +2,7 @@ import { TextControl } from '@wordpress/components';
 import { useDebounce } from '@wordpress/compose';
 import { useMemo } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
-import React from 'react';
+import React, { useState } from 'react';
 
 import { useDataSources } from '@/data-sources/hooks/useDataSources';
 import { sanitizeDataSourceSlug } from '@/data-sources/utils';
@@ -15,12 +15,14 @@ interface SlugInputProps {
 
 export const SlugInput: React.FC< SlugInputProps > = ( { slug, onChange, uuid } ) => {
 	const { slugConflicts, checkSlugConflict } = useDataSources( false );
+	const [ newSlug, setNewSlug ] = useState( slug );
 
 	// eslint-disable-next-line @typescript-eslint/no-misused-promises
 	const debouncedCheckSlugConflict = useDebounce( checkSlugConflict, 500 );
-	const onSlugChange = ( newSlug: string | undefined ): void => {
+	const onSlugUpdate = () => {
 		const sanitizedSlug = sanitizeDataSourceSlug( newSlug ?? '' );
-		if ( sanitizedSlug !== slug ) {
+		if ( sanitizedSlug !== newSlug ) {
+			setNewSlug( sanitizedSlug );
 			onChange( sanitizedSlug );
 			void debouncedCheckSlugConflict( sanitizedSlug, uuid ?? '' );
 		}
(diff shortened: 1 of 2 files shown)


src/data-sources/components/SlugInput.tsx
```diff
@@ -15,12 +15,14 @@ interface SlugInputProps {
 
 export const SlugInput: React.FC< SlugInputProps > = ( { slug, onChange, uuid } ) => {
 	const { slugConflicts, checkSlugConflict } = useDataSources( false );
+	const [ newSlug, setNewSlug ] = useState( slug );
 
 	// eslint-disable-next-line @typescript-eslint/no-misused-promises
 	const debouncedCheckSlugConflict = useDebounce( checkSlugConflict, 500 );
-	const onSlugChange = ( newSlug: string | undefined ): void => {
+	const onSlugUpdate = () => {
```
maxschmeling, line 22: Theoretically we could make this an uncontrolled text input and use the parameter to this function to get `e.target.value` instead of having to update state on every `onChange` event. Really no need for the `newSlug` `useState` at all if we do it that way which reduces rerenders.

```suggestion
	const onSlugUpdate: React.ComponentProps< typeof TextControl >[ 'onBlur' ] = e => {
```

However, for some reason, the `TextControl` component we're using from `@wordpress/components` forces you to have an `onChange` and a `value` prop... so they seem to really want you to use it as a controlled component.

So maybe just ignore all of this. I'd be interested to know why the `value` and `onChange` props are required though.

maxschmeling (approved):

maxschmeling (changes requested):
This definitely improves the UX of typing into the field, but we should go ahead and improve the sanitization to be a little more robust.

With the help of AI, something like this:
```
function slugify(input: string) {
    return input
        .toString() // Ensure input is a string
        .toLowerCase() // Convert to lowercase
        .trim() // Trim leading and trailing spaces
        .replace(/\s+/g, '-') // Replace spaces with hyphens
        .replace(/[^a-z0-9-]/g, '') // Remove invalid characters
        .replace(/--+/g, '-') // Replace multiple hyphens with a single hyphen
        .replace(/^-+|-+$/g, ''); // Trim leading and trailing hyphens
}
```

brookewp:
>This definitely improves the UX of typing into the field, but we should go ahead and improve the sanitization to be a little more robust.
>With the help of AI, something like this:

Good point! Replacing spaces with hyphens is a bonus UX improvement too 😄 Implemented that in: [`f64f545` ](https://github.com/Automattic/remote-data-blocks/pull/195/commits/f64f5452c6116bcc1a3678305d12d200074bff38)

maxschmeling (approved):
Looks great. I'm not going to request the change, but I do wonder if we should rename the sanitize function name because it's really not just sanitization, but it's a transformation into slug form (so `slugify` makes more sense to me)... but it might just be pedantic and unnecessary.

brookewp:
>I'm not going to request the change, but I do wonder if we should rename the sanitize function name because it's really not just sanitization, but it's a transformation into slug form (so slugify makes more sense to me)... but it might just be pedantic and unnecessary.

I have no objections. 🙂 I updated that [here](https://github.com/Automattic/remote-data-blocks/pull/195/commits/8290fef8856357a38d7fc55d881537253ce10ead) and added jsdocs to the function to hopefully help clarify it further

PR #196, opened 2024-11-13
Fix naming and casing for existing Airtable field types

brookewp:
I noticed when testing that some table fields weren't available to select but they were visible in the help text: 

Screenshot 2024-11-13 at 1 49 05 PM

This was due to differences in the naming of the field types from Airtable. There are still ones that won't show since we haven't included all types yet. For now, I just wanted to update the ones not showing that we had intended to include. 

Screenshot 2024-11-13 at 1 48 41 PM


maxschmeling (approved):
Awesome. @stoph and I were talking this morning about why that was limited and that we need to get support for all the types in there. I didn't realize that `singleLineText` was just typod.

Thanks! 🚢

PR #198, opened 2024-11-14
add optional generate field to input_schema

shekharnwagh:
# Description

This PR introduces a new optional 'generate' function field to the query input schema. This allows for dynamic query input generation which would be required for modifying the query input. One of the use cases for this is the PR #191 where we need to append the `".md"` suffix to the github file path if not provided. 

## Changes:
- **Enhanced Query Input Handling:**
  - Added support for `generate` functions in query input schemas to facilitate dynamic query input generation. This function receives the input values array as input and returns a new value which will be used as the value for the input field.
  - Implemented `transform_query_input_with_generators` method to apply these transformations before executing queries.
  - Modified `get_query_input` method to incorporate query input transformations after applying overrides.

- **Updated Query Execution Logic:**
  - Updated `execute_query` method to use the new `get_query_input` method, ensuring that query inputs are correctly transformed and overridden before execution.

- **Unit Tests Added**
  - Added unit tests for `BlockBindings` class in `BlockBindingsTest.php` covering:
    - Execution of queries with no configuration.
    - Execution of queries returning results.
    - Application of query input overrides.
    - Query input transformations with single and multiple inputs.
    - Combined transformations and overrides.

### DIFF
--- a/inc/Config/QueryContext/HttpQueryContext.php
+++ b/inc/Config/QueryContext/HttpQueryContext.php
@@ -37,6 +37,10 @@ class HttpQueryContext implements QueryContextInterface, HttpQueryContextInterfa
 							'type' => 'array',
 							'required' => false,
 						],
+						'generate' => [
+							'type' => 'function',
+							'required' => false,
+						],
 					],
 				],
 			],
--- a/inc/Config/QueryContext/HttpQueryContext.php
+++ b/inc/Config/QueryContext/HttpQueryContext.php
@@ -37,7 +37,7 @@ class HttpQueryContext implements QueryContextInterface, HttpQueryContextInterfa
 							'type' => 'array',
 							'required' => false,
 						],
-						'generate' => [
+						'transform' => [
 							'type' => 'function',
 							'required' => false,
 						],
--- a/tests/inc/Editor/DataBinding/BlockBindingsTest.php
+++ b/tests/inc/Editor/DataBinding/BlockBindingsTest.php
@@ -220,7 +220,7 @@ class BlockBindingsTest extends TestCase {
 			'test_input_field' => [
 				'name' => 'Test Input Field',
 				'type' => 'string',
-				'generate' => function ( array $data ): string {
+				'transform' => function ( array $data ): string {
 					return $data['test_input_field'] . ' transformed';
 				},
 			],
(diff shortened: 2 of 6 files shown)


inc/Config/QueryContext/HttpQueryContext.php
```diff
@@ -37,6 +37,10 @@ class HttpQueryContext implements QueryContextInterface, HttpQueryContextInterfa
 							'type' => 'array',
 							'required' => false,
 						],
+						'generate' => [
```
shekharnwagh: Thought about naming this as `transform` or `filter`, but ended up naming it as `generate` to match with name for similar field in the `output_schema` - https://github.com/Automattic/remote-data-blocks/blob/fa91a56f6b36a3dc72390ef7bd704a862cf2b613/inc/Config/QueryContext/HttpQueryContext.php#L61-L64

Wouldn't mind using another name if there is an good reason to do it.

tests/inc/Editor/DataBinding/BlockBindingsTest.php
```diff
@@ -0,0 +1,380 @@
...
+
+	protected function tearDown(): void {
+		parent::tearDown();
+		Mockery::close();
+		MockWordPressFunctions::reset();
+	}
+
+	/**
+	 * @runInSeparateProcess
+	 */
```
shekharnwagh, line 69: Had to add the [`@runInSeparateProcess`](https://docs.phpunit.de/en/10.5/annotations.html#runinseparateprocess) decorator to prevent static method mocks for the `ConfigStore` class from causing side effects in other tests. Although using dependency injection instead of static methods could be a better approach, it would require significant changes due to the heavy use of static methods in this codebase. Open to suggestions for better ways to handle static methods in tests.

shekharnwagh:
Used GitHub Copilot to generate the PR Description - Everything except the first paragraph. Let me what do you think of them.

maxschmeling (approved):
I debated about the `generate` name in my head a pretty good amount. I think a name like `transform` would make more sense here because, the way it's written now, you can't really "generate" new stuff... you can just return a different value that what was there. I like the consistency with the output mappings, but there you're actually generating an entirely new field that otherwise wouldn't be in the config.

Though I get that, depending on the use case and how you implement it, the difference might not actually be that big...

So I don't care too much either way, but I would vote for changing it. But I'll leave it up to you.

shekharnwagh:
> I think a name like transform would make more sense here because, the way it's written now, you can't really "generate" new stuff...

@maxschmeling This makes sense and when writing it `transform` felt more natural and better explained what the function was doing. Updated the name for the function.

PR #201, opened 2024-11-16
Fix global track props on Backend Tracks Events

mehmoodak:
- Updated Tracks prop keys from `vipgo_env` to `vip_env` and `vipgo_org` to `vip_org`.
- During testing, I've noticed that global props of tracks aren't getting logged due to a race condition - track props are calculated before WP is fully loaded.


inc/Analytics/EnvironmentConfig.php
```diff
@@ -13,23 +13,6 @@
...
-	 * 
-	 * This is set by the Tracks library available in MU Plugins.
-	 *
-	 * @var array<string, mixed>
-	 */
-	private array $tracks_core_props = [];
-
-	public function __construct() {
-		if ( function_exists( 'Automattic\VIP\Telemetry\Tracks\get_tracks_core_properties' ) ) {
-			add_action( 'init', function (): void {
```
mehmoodak, line 27: Instead of assigning variables inside the action hook and then using them, I've refactored the code to retrieve the necessary values directly. This approach avoids potential issues that can arise from calculating values based on an empty array.

maxschmeling (approved):

PR #203, opened 2024-11-18
Settings UI updates 

brookewp:
## Description

Some small changes to the design of the settings page, mainly in the header:

Screenshot 2024-11-18 at 2 33 12 PM

- replaced the `Dropdown` with the `DropdownMenu` component for improved a11y. Before, it wasn't possible to tab/use arrow keys to navigate within the popover as expected. Focus available on submenu now: 
Screenshot 2024-11-18 at 2 33 19 PM

- replaced the deletion confirmation `Modal` with the `ConfirmDialog` component

Screenshot 2024-11-18 at 2 39 29 PM

There is also a back button in the header on the edit/add source pages. As a result, there are a few visual changes and a new form component added. The form will be handled in a follow-up, but it was added here so the header didn't look odd in the meantime:
Screenshot 2024-11-18 at 1 36 04 PM
After with form component: 
Screenshot 2024-11-18 at 2 34 26 PM

## Testing Instructions

- On the settings page, tab to the 'Add' menu button and use tab/arrow keys to navigate and select subitems 
- Click on the delete icon to access confirmation dialog. Click 'cancel' and then open it again and click 'Confirm' to verify the behaviour is the same as before
- Edit/add source pages should behave the same as before, and have a back button in the header

### DIFF
--- a/src/data-sources/AddDataSourceDropdown.tsx
+++ b/src/data-sources/AddDataSourceDropdown.tsx
@@ -1,0 +1,75 @@
+import { DropdownMenu, MenuGroup, MenuItem } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { chevronDown } from '@wordpress/icons';
+
+import { SUPPORTED_SERVICES_LABELS } from './constants';
+import { useSettingsContext } from '@/settings/hooks/useSettingsNav';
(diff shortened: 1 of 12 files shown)


shekharnwagh:
When trying to add a new google sheets data source it errors out since the `slug` field is sent as empty to the API.

Screenshot 2024-11-19 at 3 15 40 PM

PR #207, opened 2024-11-20
Improve "expected error" types

chriszarate:
- In special circumstances where we are faced with insufficient library types that we cannot augment, we should still work to preserve type safety. Instead of ignoring type errors and effectively relaxing to an `any` type, we can cordon the problematic entity and reexport it with strict typing. In this PR, we do this for `registerBlockBindingsSource`.
- Miscellaneous fixes:
  - Remove redundant `export` keywords in type declarations.
  - Move `NullableKeys` utility type to its own file.

### DIFF
--- a/src/types/expected-errors/registerBlockBindingsSource.ts
+++ b/src/types/expected-errors/registerBlockBindingsSource.ts
@@ -1,0 +1,36 @@
+// @ts-expect-error Temporary registerBlockBindingsSource type error workaround for WordPress 6.7
+import { registerBlockBindingsSource as originalRegisterBlockBindingsSource } from '@wordpress/blocks';
+
+import type {
+	BlockEditorStoreActions,
+	BlockEditorStoreSelectors,
+	BlockEditorStoreDescriptor,
+} from '@wordpress/block-editor';
+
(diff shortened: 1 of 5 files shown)


src/types/expected-errors/registerBlockBindingsSource.ts
```diff
@@ -0,0 +1,36 @@
...
+import { registerBlockBindingsSource as originalRegisterBlockBindingsSource } from '@wordpress/blocks';
+
+import type {
+	BlockEditorStoreActions,
+	BlockEditorStoreSelectors,
+	BlockEditorStoreDescriptor,
+} from '@wordpress/block-editor';
+
+interface GetValuesPayload< Context, Values > {
+	bindings: Values;
```
chriszarate, line 11: this probably isn't right, but not sure what it should be. we can improve these once we start to use them

alecgeatches (approved):
Great solution to the problem for now! I like that you've isolated the type-ignores to a file and wrapped with better types.

maxschmeling (approved):
This is a good solution for now. I'd like for us to also try to get these types into the DefinitelyTyped repo so we don't have to have it in our codebase.

PR #212, opened 2024-11-22
refactor github markdown links parsing

shekharnwagh:
# Description

Updates the markdown link processing generating absolute URLs prefixed with the current page slug, ensuring proper navigation within the documentation hierarchy. This is an improvement over the previous approach of generating relative URLs. This handles the cases when the current file_path in url overrides is empty and thus not equal to the current file_path in github.

# Testing

1. In WP Admin, start editing already existing draft with page title starting with "GitHub File As HTML"

2. Add a "GitHub File As HTML" block and configure it:
   - Select a nested markdown file (e.g. `docs/index.md`) 
   - Enable URL Overrides and select the `gh/{file_path}` pattern

3. Publish the page and flush permalinks:
   - Go to Settings > Permalinks
   - Click "Save Changes"

4. Verify navigation works:
   - Visit `{your-site-url}/gh` and test the main block link
   - Click through internal documentation links to verify they resolve correctly
   - Directly access `{your-site-url}/docs/index` to confirm proper URL structure
   - Verify anchor links within pages work as expected

### DIFF
--- a/docs/workflows/index.md
+++ b/docs/workflows/index.md
@@ -4,5 +4,6 @@
 
 ## Code-based workflows
 
-- [Create an Airtable integration with code](workflows/airtable-with-code.md)
-- [Create a Zip Code integration with code contract](workflows/zip-code-with-contract.md)
+- [Create an Airtable integration with code](airtable-with-code.md)
+- [Create a Google Sheets integration with code](google-sheets-with-code.md)
+- [Create a Zip Code integration with code contract](zip-code-with-contract.md)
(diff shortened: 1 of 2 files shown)


maxschmeling:
I wonder if a better approach here might be to take a function parameter to the constructor of `GitHubGetFileAsHtmlQuery` that tells it what to do with the full path when encountered in the content. That function would be able to just "hard code" the slug and make this a little more straight forward and configurable, right?

shekharnwagh:
> I wonder if a better approach here might be to take a function parameter to the constructor of `GitHubGetFileAsHtmlQuery` that tells it what to do with the full path when encountered in the content. That function would be able to just "hard code" the slug and make this a little more straight forward and configurable, right?

@maxschmeling Do you mean somehow taking the `page_slug` as constructor argument instead of taking it from global `$post` variable ?

The problem was not just when it encountered full paths but also when it got relative paths if the anything other than a markdown file in repo root was configured as default file for the page.

maxschmeling (approved):
Ah, ok, I wasn't thinking about it correctly. Thanks!

Looks good!

Issue #12, opened 2024-08-23
Add Support for Multi-Level Logging Components / Namespaces


jblz:
...to be able to more-easily identify that the debug log came from e.g. the HTTP Client module

We could drill in via Query Monitor like this:

![Screenshot 2024-08-23 at 5 19 40 PM](https://github.com/user-attachments/assets/e1e890c5-f61b-4221-b0e0-c9ed54710ec5)

chriszarate:
The "component" in the screenshot is automatically computed by Query Monitor and is intended to represent the plugin, mu-plugin, or theme that triggered the message. It is [filterable](https://github.com/johnbillion/query-monitor/blob/9b9a31c77ac05820e2c6d61b46672b731ad43d82/classes/Util.php#L288), but that filter does not have access to the log message or its context.

I've made some related improvements in https://github.com/Automattic/remote-data-blocks/pull/154

PR #11, opened 2024-08-23
settings page updates

shekharnwagh:
# Description

The PR includes changes to the data storage model for Data Sources, particularly for Airtable integration, and User Interface improvements for the settings page.

### Data Source Storage Model

A new unique field `slug` is added to the data sources. This field is user configurable as opposed to the `uuid` which is auto generated. This allows adding data sources in code and referencing them by the slug. Secrets for the data sources can be added in the settings page and can be linked to the data source by the slug.

The `base` and `table` fields are updated to store both the `id` and `name` for the Airtable base and table. This allows for displaying the name in the UI and using the id in the API calls. The updated storage model for Airtable is -

```json
{
	"id": "rec1234567890",
	"slug": "test-airtable-source",
	"name": "Test Airtable Source",
	"type": "airtable",
	"base": {
		"id": "rec1234567890",
		"name": "Test Airtable Base"
	},
	"table": {
		"id": "rec1234567890",
		"name": "Test Airtable Table"
	}
}
```

### Settings Page UI Improvements

- Replaced the simple list view with a table view for the data sources with the slug, service, meta (to display data source specific information like base and table for Airtable and shop name for Shopify) and actions (add/edit/delete) columns.
- Added spacing between the multiple buttons in the settings list view and form view (add/edit form).
- Added a cancel button to the add/edit form, to allow going back to the list view without unnecessarily saving the form.
- Added slug field to the add/edit form to allow adding data sources by slug.
- Consistently use `help` attribute for input elements to show extra information for the input like errors, hints, additional data fetched remotely after validation input.

| Screenshot 2024-08-26 at 4 37 32 PM |
| :---: |
| Data Sources List View |

| Screenshot 2024-08-26 at 4 37 50 PM |
| :---: |
| Data Sources Form View |

PR #220, opened 2024-11-27
Fix token decoding during `wp-env` startup.

alecgeatches:
This PR fixes an bug that can cause `wp-env` startup to fail when the `remote_data_blocks_config` option has been set previously and the WordPress version managed by `wp-env` changes. To reproduce:

1. Set up `.wp-env.override.json` with config values.
1. Run `wp-env start`.
1. Change the version of WordPress in `.wp-env.json`, e.g. `"core": "WordPress/WordPress#6.7"`
1. Run `wp-env start` again.
1. See token decoding errors in `wp-env start`, and see that the site failed to start:

    ```
    $ wp-env start --xdebug
    ✖ Error while running docker compose command.
    WordPress is already installed.

    # ...

    Container 4def32faa5199ca2ba3ba70ddee94dde-tests-mysql-1  Running
    Container 4def32faa5199ca2ba3ba70ddee94dde-tests-wordpress-1  Running
    [25-Nov-2024 17:07:12 UTC] PHP Fatal error:  Uncaught TypeError: json_decode(): Argument #1 ($json) must be of type string, bool given in /var/www/html/wp-content/plugins/remote-data-blocks/inc/PluginSettings/PluginSettings.php:162
    Stack trace:
    #0 /var/www/html/wp-content/plugins/remote-data-blocks/inc/PluginSettings/PluginSettings.php(162): json_decode(false, true)
    #1 /var/www/html/wp-includes/class-wp-hook.php(326): RemoteDataBlocks\PluginSettings\PluginSettings::decrypt_option('...')
    #2 /var/www/html/wp-includes/plugin.php(205): WP_Hook->apply_filters('...', Array)
    #3 /var/www/html/wp-includes/option.php(247): apply_filters('option_remote_d...', '...', 'remote_data_blo...')
    #4 /var/www/html/wp-content/plugins/remote-data-blocks/inc/WpdbStorage/DataSourceCrud.php(69): get_option('remote_data_blo...', Array)
    #5 /var/www/html/wp-content/plugins/remote-data-blocks/inc/WpdbStorage/DataSourceCrud.php(73): RemoteDataBlocks\WpdbStorage\DataSourceCrud::get_config()
    #6 /var/www/html/wp-content/plugins/remote-data-blocks/inc/Integrations/Airtable/AirtableIntegration.php(10): RemoteDataBlocks\WpdbStorage\DataSourceCrud::get_data_sources('airtable')
    #7 /var/www/html/wp-content/plugins/remote-data-blocks/remote-data-blocks.php(47): RemoteDataBlocks\Integrations\Airtable\AirtableIntegration::init()
    #8 /var/www/html/wp-settings.php(526): include_once('/var/www/html/w...')
    #9 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(1375): require('/var/www/html/w...')
    #10 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(1294): WP_CLI\Runner->load_wordpress()
    #11 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Bootstrap/LaunchRunner.php(28): WP_CLI\Runner->start()
    #12 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/bootstrap.php(83): WP_CLI\Bootstrap\LaunchRunner->process(Object(WP_CLI\Bootstrap\BootstrapState))
    #13 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/wp-cli.php(32): WP_CLI\bootstrap()
    #14 phar:///usr/local/bin/wp/php/boot-phar.php(20): include('phar:///usr/loc...')
    #15 /usr/local/bin/wp(4): include('phar:///usr/loc...')
    #16 {main}
      thrown in /var/www/html/wp-content/plugins/remote-data-blocks/inc/PluginSettings/PluginSettings.php on line 162
    Error: There has been a critical error on this website.Learn more about troubleshooting WordPress. There has been a critical error on this website.
    ```

---

The fix involved uncovering a few layers of stack traces related to `wp-env` initialization, load order, and type checks. I'll add some more context about the changes in the code below.


inc/Integrations/Airtable/AirtableIntegration.php
```diff
@@ -7,6 +7,10 @@
 
 class AirtableIntegration {
 	public static function init(): void {
+		add_action( 'init', [ __CLASS__, 'register_blocks' ] );
```
alecgeatches: This moves block registration to the `init` action rather than in `::init()`. The latter is called as the plugin files are loaded, but not necessarily all of WordPress. This doesn't fix any particular issue right now, but in general it's good to wait until the WordPress `init` action has fired before doing any data manipulations, as all of WordPress (and all plugin files) may not have been loaded yet.

Also changed below in `ShopifyIntegration`.

inc/PluginSettings/PluginSettings.php
```diff
@@ -154,19 +155,36 @@ public static function encrypt_option( array $new_value, string|array|bool $old_
...
-	public static function decrypt_option( string $value ): array|null {
+	public static function decrypt_option( string $value ): array {
 		$decryptor = new \RemoteDataBlocks\WpdbStorage\DataEncryption();
+		$is_error = false;
 
 		try {
 			$decrypted = $decryptor->decrypt( $value );
-			return json_decode( $decrypted, true );
+
+			if ( false === $decrypted ) {
```
alecgeatches, line 165: This fixes the original stack trace in the issue description. `$decryptor->decrypt()` may return `false`, which causes the original error when `json_decode()` is called:

```
PHP Fatal error:  Uncaught TypeError: json_decode(): Argument #1 ($json) must be of type string, bool given in /var/www/html/wp-content/plugins/remote-data-blocks/inc/PluginSettings/PluginSettings.php:162
Stack trace:
#0 /var/www/html/wp-content/plugins/remote-data-blocks/inc/PluginSettings/PluginSettings.php(162): json_decode(false, true)
```

alecgeatches: From testing, I think the reason this fails is that `LOGGED_IN_KEY` and `LOGGED_IN_SALT` have different values during `wp-env` initial setup and once the site is actually running. I tried adding `error_log()`s [in the relevant code](https://github.com/Automattic/remote-data-blocks/blob/d73c623aa6124e86af2e8f82e93187bbb350e4da/inc/WpdbStorage/DataEncryption.php#L114) where we use these values and both values changed during initialization of `wp-env` after a version change and during normal website usage.

Because the values are initially encrypted with the site's real `LOGGED_IN_` keys, this process only fails during `wp-env` initialization to a new WordPress version.

inc/PluginSettings/PluginSettings.php
```diff
@@ -154,19 +155,36 @@ public static function encrypt_option( array $new_value, string|array|bool $old_
...
+			self::show_decryption_error();
+			return [];
+		} else {
+			return json_decode( $decrypted, true );
+		}
+	}
+
+	private static function show_decryption_error(): void {
+		// Check that we have add_settings_error() available. This can be unavailable during wp-env startup.
+		if ( is_admin() ) {
```
alecgeatches, line 182: Even after the fixes above, I was getting this stack trace for using the `add_settings_error()` function:

```
PHP Fatal error:  Uncaught Error: Call to undefined function RemoteDataBlocks\PluginSettings\add_settings_error() in /var/www/html/wp-content/plugins/remote-data-blocks/inc/PluginSettings/PluginSettings.php:176
Stack trace:
#0 /var/www/html/wp-content/plugins/remote-data-blocks/inc/PluginSettings/PluginSettings.php(168): RemoteDataBlocks\PluginSettings\PluginSettings::add_decrypt_error()
#1 /var/www/html/wp-includes/class-wp-hook.php(326): RemoteDataBlocks\PluginSettings\PluginSettings::decrypt_option('r7xSgvl3faIL5qW...')
#2 /var/www/html/wp-includes/plugin.php(205): WP_Hook->apply_filters('r7xSgvl3faIL5qW...', Array)
#3 /var/www/html/wp-includes/option.php(247): apply_filters('option_remote_d...', 'r7xSgvl3faIL5qW...', 'remote_data_blo...')
#4 /var/www/html/wp-content/plugins/remote-data-blocks/inc/WpdbStorage/DataSourceCrud.php(69): get_option('remote_data_blo...', Array)
#5 /var/www/html/wp-content/plugins/remote-data-blocks/inc/WpdbStorage/DataSourceCrud.php(75): RemoteDataBlocks\WpdbStorage\DataSourceCrud::get_config()
#6 /var/www/html/wp-content/plugins/remote-data-blocks/inc/Integrations/Airtable/AirtableIntegration.php(10): RemoteDataBlocks\WpdbStorage\DataSourceCrud::get_data_sources('airtable')
#7 /var/www/html/wp-content/plugins/remote-data-blocks/remote-data-blocks.php(47): RemoteDataBlocks\Integrations\Airtable\AirtableIntegration::init()
#8 /var/www/html/wp-settings.php(526): include_once('/var/www/html/w...')
#9 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(1375): require('/var/www/html/w...')
#10 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Runner.php(1294): WP_CLI\Runner->load_wordpress()
#11 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/WP_CLI/Bootstrap/LaunchRunner.php(28): WP_CLI\Runner->start()
#12 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/bootstrap.php(83): WP_CLI\Bootstrap\LaunchRunner->process(Object(WP_CLI\Bootstrap\BootstrapState))
#13 phar:///usr/local/bin/wp/vendor/wp-cli/wp-cli/php/wp-cli.php(32): WP_CLI\bootstrap()
#14 phar:///usr/local/bin/wp/php/boot-phar.php(20): include('phar:///usr/loc...')
#15 /usr/local/bin/wp(4): include('phar:///usr/loc...')
#16 {main}
  thrown in /var/www/html/wp-content/plugins/remote-data-blocks/inc/PluginSettings/PluginSettings.php on line 176
Error: There has been a critical error on this website.Learn more about troubleshooting WordPress. There has been a critical error on this website.
```

The root cause is a bit harder to determine. I *think* WP_CLI is doing something strange with load order, as [`wp-settings.php` is loaded first](https://github.com/wp-cli/wp-cli/blob/0187f2b4ce7e83f31bf4beecd5ad4f11ddf8b57b/php/WP_CLI/Runner.php#L1373) (which fires the `init` action) [before admin functions](https://github.com/wp-cli/wp-cli/blob/0187f2b4ce7e83f31bf4beecd5ad4f11ddf8b57b/php/WP_CLI/Runner.php#L1383) are loaded into memory. This does look similar to WordPress' native load order, but I'm only seeing this error during initialization through WP_CLI. I found some similar-sounding load order concerns in WP_CLI issues like https://github.com/wp-cli/wp-cli/issues/6010.

We can get around this error by ensuring `is_admin()` is true, which means that `add_settings_error()` will be defined and actually do something.

inc/PluginSettings/PluginSettings.php
```diff
@@ -154,19 +155,36 @@ public static function encrypt_option( array $new_value, string|array|bool $old_
 		}
 	}
 
-	public static function decrypt_option( string $value ): array|null {
+	public static function decrypt_option( string $value ): array {
```
alecgeatches, line 158: `decrypt_option()` was also changed to return an empty array `[]` instead of `null` on error. Returning `null` causes a runtime type failure in the parent call to [`get_config()`](https://github.com/Automattic/remote-data-blocks/blob/d73c623aa6124e86af2e8f82e93187bbb350e4da/inc/WpdbStorage/DataSourceCrud.php#L68), which expects to return an `array`.

inc/Integrations/Airtable/AirtableIntegration.php
maxschmeling: I think we should move this out to `remote-data-blocks.php` and call `AirtableIntegration::init()` from inside a callback there instead of in the `init` function here. I think once this function is called you should be able to expect the blocks to be registered.

maxschmeling (changes requested):

inc/Integrations/Airtable/AirtableIntegration.php
alecgeatches: I'm happy to change this if you still disagree, but this is the general pattern we use for almost every `::init()` from `remote-data-blocks.php`:

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/Editor/AdminNotices/AdminNotices.php#L32-L35

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/Editor/DataBinding/BlockBindings.php#L17-L20

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/Editor/DataBinding/FieldShortcode.php#L8-L10

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/Editor/DataBinding/QueryOverrides.php#L10-L12

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/Editor/BlockManagement/BlockRegistration.php#L23-L26

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/Editor/PatternEditor/PatternEditor.php#L10-L13

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/ExampleApi/ExampleApi.php#L14-L16

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/PluginSettings/PluginSettings.php#L15-L21

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/REST/RemoteDataController.php#L16-L18

---

The only other classes we don't hook into the WordPress lifecycle are in our logging and Tracks setup:

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/Editor/BlockManagement/ConfigRegistry.php#L21-L24

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/Analytics/TracksAnalytics.php#L28-L46

I could see why we may want these working prior to WordPress startup for metrics and error logging. However, even the logger internally calls this constructor during `::init()`:

https://github.com/Automattic/remote-data-blocks/blob/b49bf7b31d00ccab8ef6aaf07f59cdb00477687e/inc/Logging/Logger.php#L65

This is an anti-pattern because only code that's loaded into memory before our plugin can register to use the `wpcomvip_log_level` hook. Anything loaded after won't be able to modify this hook, because it's already been called. This may not matter for this particular filter, but this is a potential footgun on all code interacting with WordPress features before `init` or `wp_loaded`.

In summary, most of our classes already wait until the `init` hook (or other lifecycle actions) to initialize class data internally. We have the option to initialize classes before that, but it's easy to include mistakes that don't work well with other WordPress code. Unless we really need global access to a class prior to `init` and are careful about which WordPress features we use, we'll eventually run into file load-order, uninitialized global state, and other weird initialization bugs. I think the best bet is to rely on lifecycle actions and priority for this rather than carefully coding around load-order bugs unless we have a great reason. I know you're not arguing that point, but I wanted to expand on why using filters as part of class instantiation is a reasonable pattern.

Doing this internal to the class (instead of in `remote-data-blocks.php`) just delegates that responsibility to the class as we already do in most cases. If we want to ensure that blocks are initialized by the time the default `init` action is called, we could also use a lower priority.

chriszarate:
Running things that might require WordPress initialization on `init` makes sense, but we have the potential for race conditions, since `BlockRegistration` also runs on `init`. So we might need to reintroduce a custom hook that runs after init.

alecgeatches:
@chriszarate @maxschmeling We can also run our important block registration on `init` with a lower priority (like `5`) if we want to ensure it's available before `init`, or as Chris suggested above we can run it on a custom hook that runs after `init`. Any thoughts? I think it's best if we can work around WordPress' lifecycle per-component, in whatever organization makes most sense.

alecgeatches:
Actually, it already appears we have `BlockRegistration` running with a higher priority of `50`:

https://github.com/Automattic/remote-data-blocks/blob/fdf9549f96e44f299696ef9793eadf19b5aa1f96/inc/Editor/BlockManagement/BlockRegistration.php#L23-L24

This makes sense to me as we'd expect blocks to register during `init` (priority `10`) and anything that may have a race condition can run with a higher priority after like this. Are there other race condition issues we should worry about?

chriszarate (approved):
Fine with this approach, but needs to be done with more classes now

inc/Integrations/Airtable/AirtableIntegration.php
chriszarate: Add explicit arity and priority args

alecgeatches:
@maxschmeling We've had some discussion above (https://github.com/Automattic/remote-data-blocks/pull/220#discussion_r1866179776 + more) since your prior review. Could you take a look at the current state and let me know if this is good to go? Thank you!

maxschmeling:
> @maxschmeling We've had some discussion above ([#220 (comment)](https://github.com/Automattic/remote-data-blocks/pull/220#discussion_r1866179776) + more) since your prior review. Could you take a look at the current state and let me know if this is good to go? Thank you!

Looks great! *Very* thorough rebuttal 😁

maxschmeling (approved):

PR #221, opened 2024-11-27
Make QueryRunner an implementation detail and provide `execute` on public interface

chriszarate:
Make QueryRunner an implementation detail and provide `execute` on public interface


maxschmeling:
I like this approach. More straight forward for the most common use.

maxschmeling (approved):

chriszarate:
Incorporated in #243

PR #229, opened 2024-12-03
Change data source slugs to an overridable auto-UUID

brookewp:
This replaces the slug with an editable display name and overridable UUID. 

After creating a data source, there is an option to edit the UUID in Edit settings via a settings menu (a temp UI solution which will likely change with upcoming UI updates). 

For now, this is how it looks: 

https://github.com/user-attachments/assets/dd4f67db-94fb-4a6f-97dc-43982d4e8d3b

I'm not certain on the best way to handle the examples/custom sources. Do we want UUIDs manually added here (which is what I've done for now) or is there a preferred alternative?


maxschmeling:
I haven't had a chance to review this thoroughly yet, but I pushed formatting fixes and also did notice a couple of places that reference `slug` still in comments and tests that probably need updated. There are some places where `slug` is still supposed to be there, but at least a couple I noticed where I don't think it should.

I'll finish reviewing more tomorrow, but this generally looks good!

brookewp:
> I haven't had a chance to review this thoroughly yet, but I pushed formatting fixes and also did notice a couple of places that reference slug still in comments and tests that probably need updated. There are some places where slug is still supposed to be there, but at least a couple I noticed where I don't think it should.

Thanks @maxschmeling! I'm sure I missed a few, but I think some of it is related to the question in my PR description: 

>I'm not certain on the best way to handle the examples/custom sources. Do we want UUIDs manually added here (which is what I've done for now) or is there a preferred alternative?

So before updating the docs/examples/comments, I wanted to check in on thoughts here.

chriszarate:
> Do we want UUIDs manually added here (which is what I've done for now) or is there a preferred alternative?

I think UUIDs should be created by us and immutable. Editing a UUID seems to be a vehicle for "agreement" between UI and code across environments. But it's just as easily a footgun for that agreement to be accidentally disrupted by a human.

I would rather see us provide an import / export of config that preserves UUIDs but allows configs to move between environments.

maxschmeling:
> > Do we want UUIDs manually added here (which is what I've done for now) or is there a preferred alternative?
> 
> I think UUIDs should be created by us and immutable. Editing a UUID seems to be a vehicle for "agreement" between UI and code across environments. But it's just as easily a footgun for that agreement to be accidentally disrupted by a human.
> 
> I would rather see us provide an import / export of config that preserves UUIDs but allows configs to move between environments.

@chriszarate and I talked about this some directly and agreed that the UUID being editable makes less sense than it did with a slug. We might want to move toward making the UUID completely immutable and require using environment variables or some other runtime configuration to specify the environment specific UUID for a data source.

We can remove the UUID edit box in a future PR though if we decide to commit to that approach.

maxschmeling (approved):

PR #234, opened 2024-12-06
add support for all airtable field types

shekharnwagh:
### Description

> [!NOTE]
> JIRA: [CAFE-1068](https://vipjira.atlassian.net/browse/CAFE-1068)

- **Support for string arrays**: Added handling when array values are received for `string` type where it concatenates them with comma delimiter and thus returning string values.
- **Renaming `price` to `currency`**: The `price` field has been renamed to `currency` and adds support for a new field in mapping options `prefix` which allows for displaying the currency symbol or prefix.
- **Currency Prefix/Symbol Support**: Added support for currency prefixes or symbols, enhancing the display options for monetary values.
- **Comprehensive Airtable Type Support**: Added support for all the Airtable fields documented here: https://airtable.com/developers/web/api/field-model. The Airtables types are mapped to plugin's output mapping types using the following mapping:

    | Airtable Field Type    | Mapped Type    | Special Handling Notes                           |
    |------------------------|----------------|--------------------------------------------------|
    | singleLineText         | string         | Direct string conversion                         |
    | multilineText          | string         | Direct string conversion                         |
    | email                  | string         | Direct string conversion                         |
    | phoneNumber            | string         | Direct string conversion                         |
    | richText               | string         | Direct string conversion                         |
    | barcode                | string         | Direct string conversion                         |
    | singleSelect           | string         | Direct string conversion                         |
    | date                   | string         | Direct string conversion                         |
    | dateTime               | string         | Direct string conversion                         |
    | lastModifiedTime       | string         | Direct string conversion                         |
    | createdTime            | string         | Direct string conversion                         |
    | multipleSelects        | string_array   | Array of strings joined with commas             |
    | multipleRecordLinks    | string_array   | Array of record IDs using [*].id                |
    | rollup                 | string         | Direct string conversion                         |
    | externalSyncSource     | string         | Direct string conversion                         |
    | number                 | number         | Cast to float                                    |
    | autoNumber             | number         | Cast to float                                    |
    | rating                 | number         | Cast to float                                    |
    | duration               | number         | Cast to float                                    |
    | count                  | number         | Cast to float                                    |
    | percent                | number         | Cast to float                                    |
    | createdBy              | string         | Extracts user's name using .name                |
    | lastModifiedBy         | string         | Extracts user's name using .name                |
    | singleCollaborator     | string         | Extracts user's name using .name                |
    | multipleCollaborator   | string_array   | Array of user names using [*].name              |
    | url                    | button_url     | Direct string conversion                         |
    | button                 | button_url     | Direct string conversion                         |
    | currency               | currency       | Uses field's currency symbol as prefix           |
    | checkbox               | boolean        | Cast to boolean                                  |
    | multipleAttachments    | image_url      | Takes first attachment's URL using [0].url       |
    | formula                | string         | Uses result type if available, else string       |
    | lookup                 | string         | Uses result type if available, else string       |

### Testing

To test these changes:

1. Start the local development server using instructions given [here](https://github.com/Automattic/remote-data-blocks/blob/trunk/docs/local-development.md)
2. Open the WP-Admin and go to Settings > Remote Data Blocks and add new Airtable data source
3. Verify that all the fields are supported and the output is as expected.
4. Try testing by adding various types of fields and verify that Airtable fields are mapped to the correct types.


inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -251,8 +254,12 @@ protected function get_field_value( array|string $field_value, string $default_v
...
 				return $field_value_single;
 
-			case 'price':
-				return sprintf( '$%s', number_format( (float) $field_value_single, 2 ) );
+			case 'currency':
+				$currency_symbol = $mapping['prefix'] ?? '$';
+				return sprintf( '%s%s', $currency_symbol, number_format( (float) $field_value_single, 2 ) );
+
+			case 'string_array':
+				return implode( ', ', $field_value );
```
shekharnwagh: Wondering if this should be a separate type - `string_array` or would it be better if we keep the type as `string` and if it finds an array, it should just join them using comma. The only reason I can think against keeping the type as `string` is if for some reason we want to expose the raw array to the block. However, I can't think of why we'd want to do that top of my mind.

maxschmeling: I think the latter is preferred. The value is a string. It's just our way of automatically mapping the array from the source. But once we join them, it's just a string and that's what it should be on our side. If you write your own config, you can use a different type and conversion.

maxschmeling:
Other than the small string array change you mentioned, this looks really good.

maxschmeling (approved):

PR #237, opened 2024-12-09
Improve schema validation and sanitization

chriszarate:
This PR improves the schema validation and sanitization code to work with arbitrary data, instead of a rigid, flat format that is effectively proprietary to our plugin.

This allows plugin users to work with *all* remote data—not just remote data that is intended for use in blocks. Using remote data with the Interactivity API is a good example of this use case: we sometimes need to work with complex data structures alongside simpler block-based data structures, but we want to use this plugin for both, so we can benefit from its abstractions and caching.

It also allows us to represent most data sources and queries in serializable data structures—currently PHP arrays, but it could easily be JSON or something else. This is a long-discussed and wished-for feature.

Changing the underlying schema pulls at a lot of threads, so this PR is not small. This PR is for illustrative purposes: See the changes to `ExampleApi` in particular to see the benefits of these changes.

### TODOS

This work is ongoing on [the `improve/query-registration` branch](https://github.com/Automattic/remote-data-blocks/compare/improve/query-registration?expand=1). Go there if you want to play or hack on something that is in a working state. (That branch further enables a new single block registration entry point that uses a config array).

- Fix and update tests to match new schema format
- Update integrations and examples to match new schema format
- Add additional tests for nested schema validation


chriszarate:
Closing in favor of #243

PR #238, opened 2024-12-12
Send X-Cors-Proxy-Allowed-Request-Headers header on proxied requests

maxschmeling:
With the release of https://github.com/WordPress/wordpress-playground/pull/2007 we're able to explicitly enable the `Authorization` header on the CORS proxy. This allows us to make authenticated requests to APIs.

### DIFF
--- a/inc/HttpClient/WPRemoteRequestHandler.php
+++ b/inc/HttpClient/WPRemoteRequestHandler.php
@@ -30,6 +30,7 @@ class WPRemoteRequestHandler {
 		try {
 			// Convert Guzzle request to arguments for wp_remote_request.
 			$url = (string) $request->getUri();
+			$headers = [];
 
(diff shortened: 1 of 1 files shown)


chriszarate (approved):
Looks good, in the future maybe this could actually respond to the headers in the request. It might look a little squirrelly on requests that don't need it

maxschmeling:
> Looks good, in the future maybe this could actually respond to the headers in the request. It might look a little squirrelly on requests that don't need it

I briefly considered two different options:
1. Stripping it off in the proxy so the `X-Cors-Proxy-Allowed-Request-Headers` header never gets passed through.
2. Doing what you suggest and not sending it to the proxy if the request doesn't contain the `Authorization` header.

But I figured the less logic we add to it, the better. However, I'm definitely still open to either option.

PR #240, opened 2024-12-13
add airtable rich text support

shekharnwagh:
# Description

> [!NOTE]
> JIRA: [CAFE-1090](https://vipjira.atlassian.net/browse/CAFE-1090)

- Adds a new type `markdown` in output query mappings. The values are assumed as markdown string and is converted to HTML.
- Adds support for Airtable rich text fields by mapping them to the newly added 'markdown' type. These changes allow rich text content from Airtable to be correctly parsed and displayed as HTML.

## Testing

To test these changes:

1. Start the local development server using instructions given [here](https://github.com/Automattic/remote-data-blocks/blob/trunk/docs/local-development.md)
2. Setup an Airtable Base with Table having a field with type "Long Text" and rich text enabled.
3. Open the WP-Admin and go to Settings > Remote Data Blocks and add new Airtable data source by selecting the above Airtable Base/Table combination and select the field in field selection input.
4. Verify that the rich text field is displayed correctly in the block.


alecgeatches:
Very cool! Also nice to have a `markdown` field we can use in other places. I'm able to get this to work, but I do not see rich text fields unless I add them manually, and I'm not sure if that's an error. For my reproduction:

1. I'm using a shared airtable called "AlecG Test Data" that looks like this:

    Screenshot 2024-12-16 at 2 09 15 PM

    Expanded, the first row's "Notes" look like this:

    Screenshot 2024-12-16 at 2 08 34 PM

3. Go through the steps to add the table and both "Title" and "Notes" fields to the data source.
4. In the editor, add the block and select an item:

    

5. Note that the "Notes" field is missing from the selection in the item selection, template default, and on-page once added.

I'm able to manually add a `core/paragraph` and bind it to the "Notes" field correctly, which works!

Screenshot 2024-12-16 at 2 18 31 PM

However, this seems like it should be part of the default setup. Is something broken, or is this intended behavior? Thank you!

shekharnwagh:
> I'm able to get this to work, but I do not see rich text fields unless I add them manually, and I'm not sure if that's an error. 

@alecgeatches Thank you for testing this. I've pushed a small fix which resolves this - 9500e5b

alecgeatches (approved):
The fix works, and the item/pattern view now includes rich text! I did notice that HTML in the editor has a lot of whitespace:

Screenshot 2024-12-18 at 11 44 02 AM

---

versus the actual post front-end, which better reflects the original spacing:

Screenshot 2024-12-18 at 11 44 12 AM

---

If that's easy to fix, that'd be great, but otherwise this is good to go from me. Thank you!

shekharnwagh:
> If that's easy to fix, that'd be great, but otherwise this is good to go from me. Thank you!

@alecgeatches  This seems to be happening because we add the HTML code inside the paragraph block and Gutenberg seems to add extra spacing for newlines. A proper fix for this is maybe using HTML block to render this markdown and also the markdown from the Github. As far as I know block binding doesn't yet have support for HTML block. Lets keep this and tackle it when we do have support for HTML block.

PR #241, opened 2024-12-13
[UI] Admin settings and setup/edit flow updates 

brookewp:
**Changes to admin screen:** 
Screenshot 2024-12-13 at 12 42 30 PM

**Changes to setup (just having it broken up into multiple steps):** 
Screenshot 2024-12-13 at 12 44 40 PM

**Changes to edit screen:**
Screenshot 2024-12-13 at 12 45 12 PM


maxschmeling (approved):
This is a really great start on the new design. I'm imagine there are some details that could be nitpicked but we can continue improving this over time. 🚢 it

PR #242, opened 2024-12-13
Use CSS grid for services dropdown

maxschmeling:
This got accidentally left out of the last PR merge. It just uses CSS grid instead of flex to make the services menu more consistent for different sizes of text.

image


src/settings/index.scss
```diff
@@ -107,22 +107,30 @@ body {
...
 			width: 368px;
 			align-items: flex-start;
 			align-content: flex-start;
 			gap: 8px;
 			padding: 8px 8px 16px 8px;
-			flex-wrap: wrap;
+
+			.rdb-settings-page_add-data-source-btn-wrapper {
+				height: 100%;
+			}
```
brookewp, line 124: Is this needed? I'm not seeing any change with or without it

src/settings/index.scss
```diff
@@ -107,22 +107,30 @@ body {
...
 	.components-menu-group {
 
-		div {
-			display: flex;
+		> div {
+			display: grid;
+			grid-template-columns: repeat(3, 1fr);
 			width: 368px;
 			align-items: flex-start;
 			align-content: flex-start;
```
brookewp: Smallest nit ever, we could change `flex-start` to `start`

src/settings/index.scss
maxschmeling: Without it I'm seeing it not fill the full height of the row which means the hover size is different for each item.

image

src/settings/index.scss
```diff
@@ -107,22 +107,30 @@ body {
...
 			.components-button.has-icon.has-text {
+				height: 100%;
 				display: flex;
 				width: 112px;
 				padding: var(--Spacing-Spacing-Base, 16px);
 				flex-direction: column;
 				align-items: center;
 				gap: 12px;
+				align-self: stretch;
+				justify-self: stretch;
```
brookewp: Screenshot 2024-12-13 at 2 03 46 PM

These properties aren't applied so they aren't needed

maxschmeling:
Made a small update to align the content of the buttons to `start` so the icons line up, while still having the button fill the full row and look better on hover / when focused.
image

src/settings/index.scss
brookewp: Thanks for the screenshot! I see this now using FF. In Chrome, it appears differently 🤷‍♀️

brookewp:
>Made a small update to align the content of the buttons to start so the icons line up, while still having the button fill the full row and look better on hover / when focused.

I agree that the icons look better lined up, but the hover background over HTTP looks a bit awkward. What about if we just wrap the text for Salesforce and use `space-around` or `space-evenly`?

**space-around**
Screenshot 2024-12-13 at 2 37 20 PM


```diff
diff --git a/src/settings/index.scss b/src/settings/index.scss
index 80451a5..f05db0e 100644
--- a/src/settings/index.scss
+++ b/src/settings/index.scss
@@ -121,7 +121,7 @@ body {
 			}
 
 			.components-button.has-icon.has-text {
-				height: 100%;
+				height: 101px;
 				display: flex;
 				width: 112px;
 				padding: var(--Spacing-Spacing-Base, 16px);
@@ -130,7 +130,7 @@ body {
 				gap: 12px;
 				align-self: stretch;
 				justify-self: stretch;
-				justify-content: start;
+				justify-content: space-around;
 
 				&:hover {
 					color: unset;
@@ -141,13 +141,21 @@ body {
 				.components-menu-item__item {
 					min-width: unset;
 					margin-right: unset;
-					text-wrap: auto;
 					text-align: center;
 				}
 			}
 
-			.rdb-settings-page_add-data-source-btn-salesforce-b2c svg {
-				width: 50px;
+			.components-button.has-icon.has-text.rdb-settings-page_add-data-source-btn-salesforce-b2c {
+
+				svg {
+					width: 50px;
+				}
+
+				span.components-menu-item__item {
+					text-wrap: auto;
+					min-width: 112px;
+				}
+
 			}
 		}
 	}
```

or **space-evenly**

Screenshot 2024-12-13 at 2 37 32 PM

brookewp:
I looked back at the designs and noticed the icons were a bit bigger which helps with the alignment. So I've updated the PR to match:  

| Figma  | PR |
|  Screenshot 2024-12-13 at 3 30 31 PM | Screenshot 2024-12-13 at 3 49 43 PM |

Edited PR screenshot to include active button color change and size: [`4d0eb98` (#242)](https://github.com/Automattic/remote-data-blocks/pull/242/commits/4d0eb980c440cda4ff3b9c5391a033dd0fcd4481)

PR #243, opened 2024-12-13
Improve query registration and schema validation

chriszarate:
- Improve schema validation and sanitization with formal Type grammar.
  - Address the limitation with the current schema system by enabling support for self-referential schemas, allowing for arbitrary nesting of data structures.
  - Ensure the system can handle complex data structures without requiring workarounds.
- Improve query registration by providing a single entrypoint with associative array config
  - Provide a more intuitive method for ad hoc use of remote data blocks, making it easier to execute queries outside the block editor context.
  - Remove the dependency on block names as identifiers, allowing for a straightforward function call to execute queries.
  - Consolidate the registration process into a single entry point, reducing the need for multiple function calls and improving code clarity.
- Provide a "service config" for each data source that can validate itself.
  - Instead of serializing the entire config, serialize only the "service config" which can be used to "compute" create the full config. This allows us to make minor changes to the full config without breaking existing configured data sources.

## Known issues

- ~Registered generic HTTP data sources is broken.~
- ~Psalm validation is broken because of `NumberFormatter`~
- ~Github markdown source seems to be broken.~
- Any other issues encountered, please report them!


inc/Integrations/Airtable/AirtableIntegration.php
```diff
@@ -2,34 +2,45 @@
...
-
-		$block_name = $airtable_data_source->get_display_name();
-		$query = $airtable_data_source->___temp_get_query();
-		$list_query = $airtable_data_source->___temp_get_list_query();
-
-		if ( is_wp_error( $query ) || is_wp_error( $list_query ) ) {
-			LoggerManager::instance()->error( 'Failed to get query for Airtable block' );
-			return;
-		}
+	public static function register_block_for_airtable_data_source( AirtableDataSource $data_source, array $block_overrides = [] ): void {
```
maxschmeling, line 17: Should we follow the `{type}_block` naming here and make this something like `register_item_block_...` or `register_single_item_block_...` ?

maxschmeling, line 17: Should we follow the `{type}_block` naming here and make this something like `register_item_block_...` or `register_single_item_block_...` ?

example/airtable/events/register.php
```diff
@@ -3,22 +3,39 @@
...
+				'id' => 'appVQ2PAl95wQSo9S',
+				'name' => 'Conference Events',
+			],
+			'display_name' => 'Conference Events',
+			'tables' => [], // AirtableDataSource does not formally provide queries.
+		],
+	] );
+
+	$block_options = [
+		'pages' => [
```
maxschmeling, line 29: Is it necessary for page registration to be part of the block registration? I mainly ask because we might want to have overrides for multiple blocks on a page.

brookewp:
A couple things I noticed when testing:

- When selecting an item from a Shopify block, pattern selection is skipped 

- Airtable data isn't available:
**Edit: looks like this happens when a photo/attachment field is selected** 
Screenshot 2024-12-17 at 8 59 50 AM

chriszarate:
Thanks @brookewp!

> * When selecting an item from a Shopify block, pattern selection is skipped

This is intentional when the registered pattern has a `role` of `inner_blocks`. This was a requested feature at one point; if I have a pattern I know I want to use, I should skip pattern selection.

https://github.com/Automattic/remote-data-blocks/blob/b2016ca6b3b8e29e2bfc62a708fcb7e4025ea979/src/blocks/remote-data-container/hooks/usePatterns.ts#L112

> * Airtable data isn't available:
>   **Edit: looks like this happens when a photo/attachment field is selected**

I believe I've "fixed" this type mismatch issue by selecting the first item in the array. Could you confirm?

shekharnwagh:
Github markdown source seems to be broken. When adding a markdown, it renders the raw markdown code instead of its HTML version like it used to earlier.

Screenshot 2024-12-18 at 8 57 31 PM

chriszarate:
> Github markdown source seems to be broken. When adding a markdown, it renders the raw markdown code instead of its HTML version like it used to earlier.

Thanks @shekharnwagh, fixed in https://github.com/Automattic/remote-data-blocks/pull/243/commits/912831c68580e1582fcd3e67c6981a27a4913e18

inc/Integrations/Airtable/AirtableIntegration.php
chriszarate: I'm fine with whatever naming scheme. I think we might even want to centralize the entrypoints for registering official integrations, like `Integration::register_block( [ /* options */ ] )`

example/airtable/events/register.php
chriszarate: Confirmed you can do this now. The page is only created once if you use the same page slug for multiple blocks. 

We really need some deeper product exploration around pages, though, because I don't really know what use cases we are solving.

docs/extending/block-patterns.md
```diff
@@ -18,11 +18,14 @@ Example:
...
-    register_remote_data_block_pattern( $block_name, 'Pattern Title', $block_pattern );
-}
-add_action( 'init', 'YourNamespace\\register_your_block_pattern', 10, 0 );
+register_remote_data_block( [
+	'title' => 'My Remote Data Block',
+	'queries' => [ /* ... */ ],
+	'patterns' => [
+		[
+			'title' => 'My Pattern',
+			'content' => file_get_contents( '/path/to/your-pattern.html' ),
```
alecgeatches: How about a relative example here?

```suggestion
			'content' => file_get_contents( __DIR__ . '/my-pattern.html' ),
```

docs/extending/data-source.md
```diff
@@ -1,59 +1,41 @@
...
-
-The `HttpDataSource` class implements `DataSourceInterface` and provides common reusable properties of an HTTP API:
-
-- `get_endpoint(): string`: Returns the base URL of the API endpoint. This can be overridden by a query.
-- `get_request_headers(): array`: Returns an associative array of HTTP headers to be sent with each request. This is a common place to set authentication headers such as `Authorization`. This array can be extended or overridden by a query.
+A data source defines basic reusable properties of an API and is used by a [query](query.md) to reduce boilerplate. It allows helps this plugin represent your data source in the plugin settings screen and other UI.
 
 ## Example
 
 Most HTTP-powered APIs can be represented by defining a class that extends `HttpDataSource`. Here's an example of a data source for US ZIP code data:
```
alecgeatches:
```suggestion
Most HTTP-powered APIs can be represented by defining a class that extends `HttpDataSource`. Here's an example of a data source for an example HTTP API:
```

docs/extending/query-runner.md
```diff
@@ -1,43 +1,35 @@
...
 
 - Your API does not respond with JSON or requires custom deserialization logic.
 - Your API uses a non-HTTP transport.
 - You want to implement custom processing of the response data that is not possible with the provided filters.
 
-## QueryRunner
+## Custom QueryRunner for HTTP queries
 
-If your API transacts over HTTP and you want to customize the query runner, consider extending the `QueryRunner` class and overriding select methods.
+If your API transacts over HTTP and you want to customize the query runner, consider extending the `QueryRunner` class and providing an instance to your query via the `query_runner` option. Here are the methods
```
alecgeatches:
```suggestion
If your API transacts over HTTP and you want to customize the query runner, consider extending the `QueryRunner` class and providing an instance to your query via the `query_runner` option. Here are the methods:
```

docs/extending/query.md
```diff
@@ -1,197 +1,209 @@
...
+### output_schema: array (required)
 
-The `get_data_source` method returns the data source associated with the query. By default, this method returns the data source that was provided to the class constructor. In most instances, you should not need to override this method.
+The `output_schema` property defines how to extract data from the API response. The method should return an associative array with the following structure:
 
-### get_endpoint( array $input_variables ): string
-
-By default, the `get_endpoint` method proxies to the `get_endpoint` method of query's data source. Override this method to set a custom endpoint for the query—for example, to construct the endpoints using an input variable. The input variables for the current request are provided as an associative array (`[ $var_name => $value ]`).
+- `format` (optional): A callable function that formats the output variable value.
+- `generate` (optional): A callable function that generates or extracts the output variable value from the response, as an alternative to `path`.
```
alecgeatches, line 113: I see that this is an alternative to `path`, but `path` below is listed as (required). A bit confusing, clarification here would be nice.

docs/extending/query.md
```diff
@@ -1,197 +1,209 @@
...
+The `request_body` property defines the request body for the query. It can be an associative array or a callable function that returns an associative array. The callable function accepts an associative array of input variables (`[ $var_name => $value ]`). If omitted, the query will not have a request body.
 
-### get_query_name(): string
+### cache_ttl: int|null|callable
 
-Override this method to specify a name that represents the query in UI.
+The `cache_ttl` property defines how long the query response should be cached, in seconds. It can be an integer, a callable function that returns an integer, or `null`. The callable function accepts an associative array of input variables (`[ $var_name => $value ]`).
 
-### get_query_runner(): QueryRunnerInterface
+A value of `-1` indicates the query should not be cached. A value of `null` indicates the default TTL should be used (60 seconds). If omitted, the default TTL is used.
```
alecgeatches, line 184: Can we add a bit of information here about what is required for this caching to work? It appears that internally this TTL is passed to `wp_cache_set()`/`wp_cache_get()`. How about:

```suggestion
A value of `-1` indicates the query should not be cached. A value of `null` indicates the default TTL should be used (60 seconds). If omitted, the default TTL is used.

Remote data blocks utilize the WordPress object cache (`wp_cache_get()` / `wp_cache_set()`) for response caching. Ensure a persistent object cache plugin is provided by your platform or installed for this value to be respected.
```

docs/workflows/index.md
```diff
@@ -4,6 +4,7 @@
 
 ## Code-based workflows
 
-- [Create an Airtable integration with code](airtable-with-code.md)
-- [Create a Google Sheets integration with code](google-sheets-with-code.md)
-- [Create a Zip Code integration with code contract](zip-code-with-contract.md)
+- [Create an Airtable integration](airtable.md)
+- [Create a Google Sheets integration](google-sheets.md)
+- [Create a REST API integration](rest-api.md)
+- [Create a REST API integration with code](rest-api-with-code.md)
```
alecgeatches: Should the top three bullet points be under the "UI-based workflows" header? It's a bit confusing to see "Create a REST API integration" and "Create a REST API integration with code" together here.

alecgeatches: Also "with code" / "Code-based" might be the wrong nomenclature since they both utilize code, it's more about the starting point. Maybe "UI-configured sources" versus "Code-generated sources" or something like that?

docs/workflows/rest-api-with-code.md
```diff
@@ -0,0 +1,67 @@
...
+use RemoteDataBlocks\Config\Query\HttpQuery;
+
+function register_zipcode_block(): void {
+	$zipcode_data_source = HttpDataSource::from_array( [
+		'service_config' => [
+			'__version' => 1,
+			'display_name' => 'Zip Code API',
+			'endpoint' => 'https://api.zippopotam.us/us/',
+		],
+    ] );
```
alecgeatches: Bit of inconsistent spacing here.

example/airtable/events/register.php
```diff
@@ -3,22 +3,39 @@
...
+			],
+		],
+	];
 
-	register_remote_data_block( $block_name, $airtable_get_event_query );
-	register_remote_data_list_query( $block_name, $airtable_list_events_query );
-	register_remote_data_loop_block( 'Conference Event List', $airtable_list_events_query );
-	register_remote_data_page( $block_name, 'conference-event' );
+	AirtableIntegration::register_block_for_airtable_data_source( $airtable_data_source, $block_options );
+	AirtableIntegration::register_loop_block_for_airtable_data_source( $airtable_data_source, $block_options );
```
alecgeatches, line 38: @chriszarate I'm curious why we use these new integration-specific wrappers. Why is that?

docs/extending/block-registration.md
```diff
@@ -1,14 +1,47 @@
...
+		],
+	] );
 
-    register_remote_data_block( $block_name, $your_query );
+    register_remote_data_block( [
+		'title' => 'My Block',
+		'queries' => [
+			'display' => $display_query,
+		],
+	] );
```
alecgeatches, line 44: Odd spacing here.

inc/Config/DataSource/HttpDataSource.php
```diff
@@ -16,94 +14,97 @@
...
 		}
 
 		return static::from_array( $config );
 	}
 
 	/**
 	 * @inheritDoc
-	 * @psalm-suppress ParamNameMismatch reason: we want the clarity provided by the rename here
+	 *
+	 * TODO: Do we need to sanitize this to prevent leaking sensitive data?
```
alecgeatches, line 82: I may be misunderstanding the security concern here, but I don't think so. We don't currently expose this in the UI (I think), we just slurp it into `ConfigStore` or use values in code. If a user can run arbitrary code to dump `::to_array()`, they'd also have the ability to read `define`s or env keys or WP config values anyhow. Am I missing something?

chriszarate: > We don't currently expose this in the UI (I think)

@alecgeatches Not visually, but anyone that can access the REST endpoint that supports the plugin settings screen (`GET remote-data-blocks/data-sources`) can view the full service config, including credentials. Currently this is limited to `manage_options` but I imagine there will be a push to lower that substantially.

This is currently intentional-by-design, because there is front-end code that needs those credentials in order to call the APIs to do schema discovery. However, we probably need to redact sensitive credentials and conduct these requests via the REST API, so that we don't expose them.

chriszarate: They even make their way into the DOM, since they are used to populate password fields when a user edits an existing data source.

docs/extending/block-registration.md
chriszarate: Fixed in https://github.com/Automattic/remote-data-blocks/pull/243/commits/2fb45d6ab5007529743762b2cc42a8cca43c1e5a

docs/workflows/rest-api-with-code.md
chriszarate: Fixed in https://github.com/Automattic/remote-data-blocks/pull/243/commits/2fb45d6ab5007529743762b2cc42a8cca43c1e5a

docs/workflows/index.md
chriszarate: Addressed in https://github.com/Automattic/remote-data-blocks/pull/243/commits/5060870e147da1b4895e26f87c1fea67b1d3d110

docs/extending/query.md
chriszarate: Addressed in https://github.com/Automattic/remote-data-blocks/pull/243/commits/5060870e147da1b4895e26f87c1fea67b1d3d110

example/airtable/events/register.php
chriszarate: Because they automatically provide the queries that are specific to the integration. Definitely open to different naming or organization.

maxschmeling (approved):
I really like the direction this moves us. I'm sure there will be some tweaks still, but lets get it merged before we have more conflicts! 🚢

PR #251, opened 2024-12-16
[UI] Implement DataViews in editor

brookewp:
This adds a basic DataViews into the editor. There will be a lot of follow ups to refine and implement available features. For example, this doesn't include filtering but that will require updating our schema. 

Screenshot 2024-12-20 at 2 38 02 PM
Screenshot 2024-12-20 at 2 38 12 PM


src/blocks/remote-data-container/components/modals/DataViewsModal.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+interface DataViewsModalProps {
+	blockName: string;
+	headerImage?: string;
+	onSelect: ( data: RemoteDataQueryInput ) => void;
+	queryKey: string;
+	title: string;
+}
+
+export function DataViewsModal( props: DataViewsModalProps ) {
```
maxschmeling: I personally prefer this style just because the `React.FC` type enforces the return value as well. This is super minor though and we don't have a defined standard.

```suggestion
export const DataViewsModal: React.FC< DataViewsModalProps > = props => {
```

maxschmeling (approved):

PR #257, opened 2024-12-19
Add DataViews and replace table in settings

brookewp:
🤞 Another attempt at #255 which was reverted in #256, but this includes an upgraded `@wordpress/dependency-extraction-webpack-plugin` which is required for the `/wp` build of DataViews as it's the first version to not extract it.


maxschmeling (approved):
Seems to work! Let's merge and hope nothing blows up this time! 🚢

PR #270, opened 2025-01-03
auto-register blocks for google sheets data sources

shekharnwagh:
# Description

> [!NOTE]
> JIRA: [CAFE-1069](https://vipjira.atlassian.net/browse/CAFE-1069)

Auto-registers single and loop blocks for Google Sheets data sources added via the Remote Data Blocks Settings page. This brings it on par with Airtable integration. All fields are mapped to string types as Google Sheets sends tabular data in response without type information, nor does it strictly enforce types for cell values. A drawback of defaulting to string types is that fields with other types like image_url, numbers won't be handled properly and would be displayed as strings.

# Testing

- Add a Google Sheets data source via the Remote Data Blocks Settings page
- Verify that the single and loop blocks are registered for the data source in the editor and they work as expected


maxschmeling (approved):
Works great! 🚢

PR #275, opened 2025-01-06
Fix documentation links

alecgeatches:
Fix links to Airtable/Sheets/REST API implementation in docs.


alecgeatches:
Some small documentation fixes. I tried adding [markdown-link-check](https://github.com/tcort/markdown-link-check) as a GitHub action, but removed it because the [action output](https://github.com/Automattic/remote-data-blocks/actions/runs/12638849765/job/35215963892) checks too many links, including API endpoint parts that we reference. We could probably find a way to ignore those, but I mostly wanted something to insure internal documentation links are correct, and this is a bit too heavy.

alecgeatches:
I'm just going to merge this because these are tiny fixes.

PR #278, opened 2025-01-07
add multi select for tables in data sources settings

shekharnwagh:
# Description

This PR adds option to add multiple sheets from a spreadsheet for Google Sheets data source and multiple tables from a base for Airtable data source. When sheets/tables are added, all supported fields from the sheet/table are added to the data source to be used in the remote data blocks.

A separate PR will add the auto-registering blocks for each of selected sheets/tables. For now, it continues with existing behavior of registering block for the first sheet/table.

| Screenshot 2025-01-10 at 5 45 40 PM |
| :----: |
| Airtable data source with multiple tables |

## Changes

- Converted the sheets/tables selection into a multi-select input.
- Removed the input to select the fields from the sheet/table. All the supported fields are added automatically.
- Removed debouncing API requests for Google Sheets/Airtable when it depends on discreet changes in the input fields (e.g. when it depends on change in select input).

# Testing

- Try adding new Airtable or Google Sheets data source and verify that you can now add multiple sheets/tables. By inspecting the network calls, verify that all the supported fields are added for each sheet/table selected for the data source.

### DIFF
--- a/inc/Editor/DataBinding/BlockBindings.php
+++ b/inc/Editor/DataBinding/BlockBindings.php
@@ -234,7 +234,7 @@ class BlockBindings {
 			$fallback_content = $block_attributes['content'] ?? null;
 		}
 
-		return $fallback_content;
+		return strval( $fallback_content );
 	}
 
 	public static function get_remote_value( array $block_context, array $source_args ): string|null {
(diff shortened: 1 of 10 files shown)


inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -234,7 +234,7 @@ private static function get_block_fallback_content( array $source_args, array $b
 			$fallback_content = $block_attributes['content'] ?? null;
 		}
 
-		return $fallback_content;
+		return strval( $fallback_content );
```
chriszarate, line 237: 🙇

chriszarate (approved):

inc/Editor/DataBinding/BlockBindings.php
chriszarate: I think this is not really adequate, because $fallback_content could be null. But I have a fix in a separate PR

chriszarate:
Merging to unblock other improvements

PR #280, opened 2025-01-09
Delegate override and rewrite rule logic to implementor

chriszarate:
Quick primer: An important use case for Remote Data Blocks is the concept of “overrides.” A user might select and place a remote data block for Product A on a page, but allow a query variable to override that selection, resulting in Product B being displayed.

image

Using a WordPress rewrite rule, the user could even construct a “product details page” where the query variable is supplied via the URL:

image

Overrides don’t have to completely override content; they might provide a “promo code” that affects the price of a product or a color way that alters the preview image. Anything that is an input variable to the underlying query is a potential target for an override.

The complexity arrives when we try to configure the WordPress behavior on behalf of the user. Currently, our plugin needs to know how to configure the rewrite rule and what to name the query variables. The resulting configuration is confusing and error-prone:

```php
register_remote_data_block( [
    'title' => 'Product',
    'render_query' => [
        'query' => $get_product_query,
        'input_overrides' => [
            [
                'source' => 'product_id',
                'source_type' => 'page',
                'target' => 'id',
                'target_type' => 'input_var',
            ],
        ],
    ],
    'pages' => [
        [
            'slug' => 'products',
            'title' => 'Product',
        ],
    ],
] );
```

Users need to understand this complex config structure, grasp the relationship between input_overrides and pages, and accept extremely limited control over the resulting page and rewrite rule. It’s frustrating, far from intuitive, and I have stumbled on it myself (despite writing it).

Users might also want to implement additional kinds of overrides beyond query variables, such as personalization or localization, which will require even more complexity and loss of user control.

### Delegation

@maxschmeling had the idea to reel things back and delegate the implementation of the override to the user. Instead of trying to implement the behavior of the override in our plugin, we simply ask the user to declare its existence. The implementation is then delegated to the user, which is a small code burden.

```php
register_remote_data_block( [
    'title' => 'Product',
    'render_query' => [
        'query' => $get_product_query,
        'input_overrides' => [
            [
                'display_name' => __( 'Override product from URL', 'text-domain' ),
                'name' => 'product-id-url',
            ],
        ],
    ],
] );
 
add_rewrite_rule( '^products/([0-9]+)/?$', 'index.php?pagename=products&product_id=$1', 'top' );
 
add_filter( 'query_vars', function ( $vars ) {
    $vars[] = 'product_id';
    return $vars;
}, 10, 1 );
 
add_filter(
    'remote_data_blocks_query_input_variables',
    function ( $input_vars, $enabled_overrides ) {
        if ( true === in_array( 'product-id-url', $enabled_overrides ) ) {
            $input_vars['id'] = intval( get_query_var( 'product_id', '123' ) );
        }
 
        return $input_vars;
    },
    10,
    2
);
```

Defining a rewrite rule and adding a query variable are extremely common WordPress dev tasks with mountains of prior art. The final piece is utilizing a filter on query input variables just prior to query execution.

While this does require code, delegating this logic to the user avoids all kinds of stumbling blocks and provides a ton of flexibility to the user to define URLs, variables, and behavior in the way that suits them best. The code itself is simple, “WordPress-y,” and can be easily documented.

### Overrides as a generic concept

Overrides don't have to be limited to modifying query input variables. They can be a vehicle or any kind of behavioral override or change—almost like a feature flag for implementors to hook into. This PR also adds a `remote_data_blocks_query_response` filter that also receives an `$enabled_overrides` array and allows a plugin user to modify the query response before it is used to satisfy a block binding.


tests/inc/Analytics/TracksAnalyticsTest.php
```diff
@@ -10,23 +10,11 @@
...
 use RemoteDataBlocks\Integrations\Shopify\ShopifyDataSource;
-
-// Define a mock class for Tracks.
-if ( ! class_exists( 'Automattic\VIP\Telemetry\Tracks' ) ) {
-	class MockTracks {
-		/**
-		 * @psalm-suppress PossiblyUnusedParam
-		 */
-		public function record_event( $_name, $_props ) {}
-	}
```
chriszarate, line 21: Moved to its own file to fix lint complaint

tests/inc/Analytics/TracksAnalyticsTest.php
```diff
@@ -58,7 +46,6 @@ public function testInitDoesSetTracksIfTrackingIsEnabledViaFilter(): void {
 		TracksAnalytics::init( $env_config_mock );
 
 		$this->assertInstanceOf( MockTracks::class, TracksAnalytics::get_instance() );
-		$this->assertEquals( [ 'activated_plugin', 'deactivated_plugin', 'save_post' ], array_keys( $GLOBALS['__wordpress_actions'] ) );
```
chriszarate, line 61: This assertion (and the one below) are mostly just testing WordPress core logic. Mocking WP actions and filters robustly for pure unit tests is a bit redundant and complex, and adds little value. Asserting that the class has been instantiated is enough IMO.

tests/inc/Mocks/MockWordPressFunctions.php
```diff
@@ -3,17 +3,43 @@
...
 
-	public static function set_query_var( string $key, $value ): void {
-		self::$query_vars[ $key ] = $value;
+	/** @var array<string, mixed> */
+	private static array $mocked_filters = [];
+
+	/** @var array<string, mixed> */
+	private static array $mocked_options = [];
+
+	public static function apply_filters( string $filter, mixed $thing ): mixed {
```
chriszarate, line 15: Encapsulated mock data here avoids the need for GLOBALS.

docs/extending/hooks.md
```diff
@@ -55,6 +55,46 @@ function custom_request_details( array $request_details, HttpQueryInterface $que
...
+		if ( ! empty( $override_value ) ) {
+			$input_variables['id'] = $override_value;
+		}
+	}
+
+	return $input_variables;
+}, 10, 4 );
+```
+
+Note that if your filter modify a query's input variables, it will modify the object cache key for that query execution, possibly resulting in a cache miss.
```
maxschmeling: This had a grammar issue, so I made a fix + some more tweaks. Feel free to just fix the grammar issue though if you want and undo my other suggestions.
```suggestion
Keep in mind modifying query input variables will affect the object cache key for that query execution. This could result in a cache miss.
```

docs/extending/hooks.md
```diff
@@ -55,6 +55,46 @@ function custom_request_details( array $request_details, HttpQueryInterface $que
...
+
+### remote_data_blocks_query_response
+
+Filter the query response just after query execution. This filter is useful for modifying the query response for the current page-load, e.g., by pulling in data from query variables or other context. See [Overrides](overrides.md) for more information.
+
+```php
+add_filter( 'remote_data_blocks_query_input_variables', function ( array $input_variables, array $enabled_overrides, string $block_name, array $block_context ): array {
```
maxschmeling: This code block needs updated for this filter.

inc/Validation/ConfigSchemas.php
```diff
@@ -108,6 +89,14 @@ private static function generate_remote_data_block_config_schema(): array {
 					] )
 				)
 			),
+			'overrides' => Types::nullable(
+				Types::list_of(
+					Types::object( [
+						'name' => Types::string(),
+						'display_name' => Types::nullable( Types::string() ),
```
maxschmeling: I don't know that we should allow this to be null. If we start from the beginning enforcing it to be there we don't ever have to think about it... and I imagine you really always should have it.

But if you think it makes sense to allow null I'm ok with it.

maxschmeling (approved):

inc/Validation/ConfigSchemas.php
chriszarate: Agreed, we can make it required

PR #282, opened 2025-01-10
auto-register blocks for each sheet/table

shekharnwagh:
# Description

### Auto-registering blocks for each sheet/table

- Moves the query generation to [Airtable/GoogleSheets]Integration from the data source files
- Auto-register blocks for each sheet/table instead of just the first sheet/table in the Airtable or Google Sheets data source
- Auto-register loop blocks for Airtable in addition to single entity blocks

### Data source list view fix

The meta column in data sources list view was displaying the name for the first table or sheet selected for Airtable and Google Sheets data sources. With multiple sheet/table selection now supported for those data sources, it was too much information to show all the selected sheet/table names with the current UI. Removing them and showing only spreadsheet and base name for them is tidier.

## Testing

- Add Airtable or Google Sheets data source with multiple sheet/table selection and verify that blocks are auto-registered for each sheet/table
- Verify that a loop block is also auto-registered for Airtable data sources


maxschmeling (approved):
Works great! 🚢

PR #283, opened 2025-01-10
Improve string sanitization and consolidate `BlockBindings::get_value()`

chriszarate:
Two intertwined improvements here:

1. Improve the string sanitization of `BlockBindings::get_value()` and the underlying private `get_block_fallback_content()`. Previously, it was possible for it to return the string `"null"` instead of `null`.
2. Consolidate `BlockBindings::get_value()` and `BlockBindings::get_remote_value()`. Having multiple entrypoints was confusing and led to diverging logic (the latter did not load fallback values).


alecgeatches (approved):
Looks great to me! I also tested the data fallback scenarios from #258 and they're all still working.

PR #285, opened 2025-01-10
Add `targetFolderName` to blueprint pluginData options

chriszarate:
Attempt to resolve recent errors when launching in playground


chriszarate:
https://github.com/WordPress/wordpress-playground/blob/9da46089b11241701f0a31493207bf6b33029d87/packages/playground/blueprints/src/lib/steps/install-plugin.ts#L79

PR #287, opened 2025-01-11
Cafe 1107/update/documentation

smithjw1:
Work to update the documentation ahead of a closed beta release.


README.md
```diff
@@ -1,16 +1,54 @@
+---
+### :warning: This plugin is currently in Beta. It is designed to run on [WordPress VIP](https://wpvip.com). This beta release is not intended for use on a production environment.
```
maxschmeling: Do we want to say "not intended for use on a production environment"?

Maybe something like, "breaking changes could occur in future releases ... use at your own risk"

smithjw1: It's what we have said for other betas we've done. If we don't think it's true, I'm fine with:

> Breaking changes could occur with any update. Please test each release thoroughly before updating.

chriszarate: So updated.

chriszarate (commented):
J'approve

PR #294, opened 2025-01-13
[UI] Add display name check back

brookewp:
At the end of the year when a lot of changes were happening, there were many conflicts and these changes were overwritten along the way. 

This adds back a check so a data source cannot be registered with an empty or existing display name. This is important now since blocks only contain the display name, so to find them, we need to make sure there aren't issues/conflicts. 

Screenshot 2025-01-13 at 1 04 36 PM

### DIFF
--- a/src/data-sources/hooks/useDataSources.ts
+++ b/src/data-sources/hooks/useDataSources.ts
@@ -1,6 +1,6 @@
 import apiFetch from '@wordpress/api-fetch';
 import { useDispatch } from '@wordpress/data';
-import { useEffect, useState } from '@wordpress/element';
+import { useCallback, useEffect, useState } from '@wordpress/element';
 import { __, sprintf } from '@wordpress/i18n';
 import { store as noticesStore, NoticeStoreActions, WPNotice } from '@wordpress/notices';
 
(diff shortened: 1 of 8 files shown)


src/data-sources/hooks/useDataSources.ts
```diff
@@ -17,6 +17,19 @@ export const useDataSources = < SourceConfig extends DataSourceConfig = DataSour
 		useDispatch< NoticeStoreActions >( noticesStore );
 	const { goToMainScreen } = useSettingsContext();
 
+	const checkDisplayNameConflict = useCallback(
```
chriszarate, line 20: Nit: `useCallback` isn't useful here, it's not being used as a prop so won't cause rerenders

src/data-sources/hooks/useDataSources.ts
```diff
@@ -17,6 +17,19 @@ export const useDataSources = < SourceConfig extends DataSourceConfig = DataSour
...
+		( displayName: string, uuid: string ): boolean => {
+			if ( ! displayName ) {
```
chriszarate, line 22:
```suggestion
			if ( ! displayName || ! uuid ) {
```

src/data-sources/hooks/useDataSources.ts
```diff
@@ -17,6 +17,19 @@ export const useDataSources = < SourceConfig extends DataSourceConfig = DataSour
...
+				return false;
+			}
+			const existingSource = dataSources.find(
+				source => source.uuid !== uuid && source.service_config.display_name === displayName
+			);
+			return ! existingSource;
```
chriszarate, line 28:
```suggestion
			return dataSources.some(
				source => source.uuid !== uuid && source.service_config.display_name === displayName
			);
```

chriszarate:
Good improvement, but eventually we should not use the data source name as the block name and let them be separately configurable.

PR #295, opened 2025-01-13
[UI] A few random fixes 

brookewp:
After adding validation for the display name, in #294, I introduced an issue where the button wasn't enabled after completing the last step. This fixes that, as well as updates the error message in Shopify to reflect the updated label. And CSS to ensure that unstyled text doesn't overflow its column in the editor.

### DIFF
--- a/src/data-sources/hooks/useDataSources.ts
+++ b/src/data-sources/hooks/useDataSources.ts
@@ -22,10 +22,10 @@ export const useDataSources = < SourceConfig extends DataSourceConfig = DataSour
 			if ( ! displayName ) {
 				return false;
 			}
-			const existingSource = dataSources.find(
+			const alreadyExists = dataSources.some(
 				source => source.uuid !== uuid && source.service_config.display_name === displayName
 			);
-			return ! existingSource;
+			return ! alreadyExists;
 		},
 		[ dataSources ]
 	);
--- a/src/data-sources/hooks/useDataSources.ts
+++ b/src/data-sources/hooks/useDataSources.ts
@@ -1,6 +1,6 @@
 import apiFetch from '@wordpress/api-fetch';
 import { useDispatch } from '@wordpress/data';
-import { useCallback, useEffect, useState } from '@wordpress/element';
+import { useEffect, useState } from '@wordpress/element';
 import { __, sprintf } from '@wordpress/i18n';
 import { store as noticesStore, NoticeStoreActions, WPNotice } from '@wordpress/notices';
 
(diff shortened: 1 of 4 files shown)


src/data-sources/hooks/useDataSources.ts
```diff
@@ -22,10 +22,10 @@ export const useDataSources = < SourceConfig extends DataSourceConfig = DataSour
 			if ( ! displayName ) {
 				return false;
 			}
-			const existingSource = dataSources.find(
+			const alreadyExists = dataSources.some(
 				source => source.uuid !== uuid && source.service_config.display_name === displayName
 			);
-			return ! existingSource;
+			return ! alreadyExists;
```
chriszarate: Isn't this the opposite of what you want?

chriszarate: The logic definitely flipped. Maybe `checkDisplayNameConflict` is just an ambiguous name that supports both

chriszarate: By which I mean, it flipped in my brain :)

brookewp: No, but it's confusing to read 😅 So I can make it more explicit: 

```
	const checkDisplayNameConflict = useCallback(
		( displayName: string, uuid: string ): boolean => {
			// ensure display name isn't an empty string or already in use by another data source
			if (
				! displayName ||
				dataSources.some(
					source => source.uuid !== uuid && source.service_config.display_name === displayName
				)
			) {
				return false;
			}

			return true;
		},
		[ dataSources ]
	);
```

or if the comment is helpful enough, I could simplify it: 

```
	const checkDisplayNameConflict = useCallback(
		( displayName: string, uuid: string ): boolean =>
			// ensure display name isn't an empty string or already in use by another data source
			! displayName ||
			! dataSources.some(
				source => source.uuid !== uuid && source.service_config.display_name === displayName
			),
		[ dataSources ]
	);
	```

Comment (brookewp) [reply src/data-sources/hooks/useDataSources.ts]: Or is it confusing that `checkDisplayNameConflict` returns `true` for no conflicts and `false` for conflicts?

Comment (chriszarate) [reply src/data-sources/hooks/useDataSources.ts]: > Or is it confusing that checkDisplayNameConflict returns true for no conflicts and false for conflicts?

It's this, for me. Renaming to `displayNameHasConflict` (or something with clear logical implemications) would fix my brain

Comment (brookewp) [reply src/data-sources/hooks/useDataSources.ts]: Hopefully this makes it more clear (renamed to `canUseDisplayName`) [`78a1137` (#295)](https://github.com/Automattic/remote-data-blocks/pull/295/commits/78a11377d494a6b19c50e7a4d76ed9dbdd832676)

File: src/data-sources/hooks/useDataSources.ts [hunk 2]
```diff
@@ -17,16 +17,13 @@ export const useDataSources = < SourceConfig extends DataSourceConfig = DataSour
...
-	const checkDisplayNameConflict = useCallback(
-		( displayName: string, uuid: string ): boolean => {
-			if ( ! displayName ) {
-				return false;
-			}
-			const existingSource = dataSources.find(
+	const canUseDisplayName = useCallback(
+		( displayName: string, uuid: string ): boolean =>
+			// ensure display name isn't an empty string or already in use by another data source
+			! displayName ||
```
chriszarate: I think this logic is wrong. An empty string returns `true`

chriszarate:
```
const canUseDisplayName = ( displayName: string, uuid: string ) =>
	displayName && dataSources.every(
		source => source.uuid === uuid || source.service_config.display_name !== displayName
	);
```

brookewp: You're right, thank you! Updated here: [`38b4dec` (#295)](https://github.com/Automattic/remote-data-blocks/pull/295/commits/38b4dec618ca09f64c8760cc06ecf80db98d84e5)

chriszarate (approved):

PR #296, opened 2025-01-14
fix github example

shekharnwagh:
- fix the variable name used to access github file path
- stop sanitizing every value as string in `get_value` which was leading to github file content being treated as normal string and does the html tags being stripped from them


inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -206,9 +206,6 @@ public static function get_value( array $source_args, WP_Block|array $block ): ?
 			return $fallback_content;
 		}
 
-		// Sanitize the value as string.
-		$value = Sanitizer::sanitize_primitive_type( 'string', $value );
```
chriszarate, line 210: Thanks for fixing. This was an alternate approach to `strval` so we should restore that instead.

chriszarate: But with a check that prevents things like `strval( null )`

shekharnwagh: We already have a check for that above this - https://github.com/Automattic/remote-data-blocks/blob/abe9e0ccb144da120c1268b09d01e9159e266eb2/inc/Editor/DataBinding/BlockBindings.php#L204-L207

chriszarate (approved):

chriszarate:
Thank you!

PR #300, opened 2025-01-17
Additional beta docs improvements

maxschmeling:
This is a shared branch for documentation updates prior to the closed beta release.


chriszarate:
A note on https://github.com/Automattic/remote-data-blocks/pull/300/commits/84bcf7e1bb661dd335f716326caf3b626c25aefc. I removed most of the installation instructions, because they were:

- Taking up more than half of the README
- Outdated: They promote riding `trunk` instead of using releases. We don't even do that.
- Oddly WPVIP specific. Explaining git subtrees on our landing page seemed a bit extra.

chriszarate:
Note on https://github.com/Automattic/remote-data-blocks/pull/300/commits/19fc8aa8451fc920666a6434fd57457bd103a6c8. I removed the link to Markdown files in the `example/` folder, because that leads to them being rendered on remotedatablocks.com. That gets confusing since the code itself can't be rendered.

Instead we can link directly to the GitHub folder and they can explore from there.

maxschmeling:
> A note on [84bcf7e](https://github.com/Automattic/remote-data-blocks/commit/84bcf7e1bb661dd335f716326caf3b626c25aefc). I removed most of the installation instructions, because they were:
> 
> * Taking up more than half of the README
> * Outdated: They promote riding `trunk` instead of using releases. We don't even do that.
> * Oddly WPVIP specific. Explaining git subtrees on our landing page seemed a bit extra.

Nice! I was planning to check if these instructions even actually worked today. Glad you removed them.

docs/extending/data-source.md
```diff
@@ -1,45 +1,87 @@
...
+])
+```
+
+In this case the `data_source` has a built in `get_endpoint()` method. You could also retrieve the endpoint this way:
+
+```php
+$zipcode_query = HttpQuery::from_array( [
+    'data_source' => $zipcode_data_source,
+    'endpoint' => function ( array $input_variables ) use ( $zipcode_data_source ): string {
+        return  $zipcode_data_source->to_array()['service_config']['endpoint'] . $input_variables['zip_code'];
```
smithjw1, line 62: @maxschmeling can you double check me on this?

docs/extending/query-output_schema.md
smithjw1, line 1: This totally tripped me up the first few times I read the documentation. so I wanted to expand this section.

docs/extending/query-output_schema.md
```diff
@@ -0,0 +1,201 @@
...
+			'type' => 'string',
+		],
+	],
+],
+```
+
+If we wanted to go further and pull out more data from each item, the schema could look like:
+
+```php
+'output_schema' => [
```
smithjw1, line 173: I did test this locally and it works as expected.

docs/extending/block-registration.md
```diff
@@ -56,8 +65,8 @@ The human-friendly name of the block. It is also used to construct the block's n
 
 The render query is executed when the block is rendered and fetches the data that will be provided to block bindings. It is an array with the following properties:
 
-- `query`: An instance of `QueryInterface` that fetches the data.
-- `loop`: A boolean that determines if the query returns a collection of data. If `true`, the block will be rendered for each item in the collection.
+- `query`: An instance of [`QueryInterface`](./query.md) that fetches the data.
+- `loop`: A boolean that indicates if the query returns a collection of data. If `true`, the block will be rendered for each item in the collection.
```
smithjw1, line 69: @chriszarate if the query has the `is_collection` property of the `output_schema` why do we also need to specify it here?

If we have to specify it twice should we call them both `is_collection` for clarity?

docs/extending/data-source.md
maxschmeling: You can just call `.get_endpoint()` on the data source because it is an `HttpDataSource` https://github.com/Automattic/remote-data-blocks/blob/trunk/inc/Config/DataSource/HttpDataSource.php#L25

docs/extending/block-registration.md
chriszarate: In general, we tried to keep the queries completely free of "block logic," because configuring some block behaviors in the query and others in the block config was reported as confusing in early internal feedback.

Therefore, deciding whether a block is a "loop block" based on something deep in the query config felt wrong and potentially confusing. "Where is this block behavior configured? Oh, in the query?"

Maybe that's silly and we should auto-detect. Or maybe we should align the names, but I'm not sure if `is_collection` adequately describes what is going to happen. (I'm not sure `loop` does either, to be fair.)

docs/extending/data-source.md
smithjw1: Yep, I just wanted to demonstrate this as an option for items without specifically defined methods.

maxschmeling: Yes, that should work. It might not be a pattern we want to advertise, but we can adjust down the road.

PR #301, opened 2025-01-17
Add type to manual input button

brookewp:
When hitting enter to submit a manual input, there was a redirect to the posts page and the submission in the field wasn't saved. 

https://github.com/user-attachments/assets/f3e25896-6beb-4ca2-8f83-d2a13451dcfe

Adding `type="submit"` should fix this

### DIFF
--- a/src/blocks/remote-data-container/components/modals/InputModal.tsx
+++ b/src/blocks/remote-data-container/components/modals/InputModal.tsx
@@ -60,7 +60,7 @@ export function InputModal( props: InputModalProps ) {
 						style={ { marginBottom: '8px' } }
 					/>
 				) ) }
-				<Button variant="primary" onClick={ onSelectItem }>
+				<Button variant="primary" onClick={ onSelectItem } type="submit">
 					{ __( 'Save' ) }
 				</Button>
 			</form>
--- a/src/blocks/remote-data-container/components/modals/InputModal.tsx
+++ b/src/blocks/remote-data-container/components/modals/InputModal.tsx
@@ -48,7 +48,13 @@ export function InputModal( props: InputModalProps ) {
 			onOpen={ open }
 			title={ props.title }
 		>
-			<form style={ { marginTop: '1rem' } }>
+			<form
+				style={ { marginTop: '1rem' } }
+				onSubmit={ ( event: React.FormEvent ) => {
+					event.preventDefault();
+					onSelectItem();
+				} }
+			>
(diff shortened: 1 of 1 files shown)


src/blocks/remote-data-container/components/modals/InputModal.tsx
```diff
@@ -60,7 +60,7 @@ export function InputModal( props: InputModalProps ) {
 						style={ { marginBottom: '8px' } }
 					/>
 				) ) }
-				<Button variant="primary" onClick={ onSelectItem }>
+				<Button variant="primary" onClick={ onSelectItem } type="submit">
```
maxschmeling: We should probably remove `onClick` here and do it in the form `onSubmit` with `e.preventDefault();` there, unless I'm missing something. Although this probably does the job well enough.

maxschmeling (changes requested):

src/blocks/remote-data-container/components/modals/InputModal.tsx
brookewp: I don't believe it makes too much of a difference since the button is inside a form. But I can see how it makes more sense having an `onSubmit` instead. Updated here: [`2146c94` (#301)](https://github.com/Automattic/remote-data-blocks/pull/301/commits/2146c942a5168dbcf00353ca7a566f301f93ecfd)

src/blocks/remote-data-container/components/modals/InputModal.tsx
```diff
@@ -48,7 +48,13 @@ export function InputModal( props: InputModalProps ) {
 			onOpen={ open }
 			title={ props.title }
 		>
-			<form style={ { marginTop: '1rem' } }>
+			<form
+				style={ { marginTop: '1rem' } }
+				onSubmit={ ( event: React.FormEvent ) => {
```
maxschmeling: This explicit type *shouldn't* be necessary.

```suggestion
				onSubmit={ event => {
```

maxschmeling (approved):

src/blocks/remote-data-container/components/modals/InputModal.tsx
maxschmeling: Yeah, it shouldn't make much of a difference. In the event that the form was ever submitted another way though, this would work properly now. I don't know if there is a way that can happen with screen readers or other accessibility tools or anything... but probably better to use the proper form submission mechanism either way..

PR #314, opened 2025-01-21
Add HTML binding block

alecgeatches:
Adds a bindable remote HTML block for direct HTML sources.

## Testing

To test, add a data source that returns content with an `html` type. This code can be directly added to the end of `remote-data-blocks.php` to define a mocked data source with HTML content:

```php
add_action( 'init', function (): void {
    $test_data_source = \RemoteDataBlocks\Config\DataSource\HttpDataSource::from_array( [
        'service_config' => [
            '__version' => 1,
            'display_name' => 'Test HTML content',
            'endpoint' => 'https://wpvip.com/not-a-real-api',
        ],
    ] );

    $test_query_runner = get_query_runner_with_response([
        'header' => 'Test Header',
        'blurb' => 'Test Blurb',
        'htmlContent' => '
            <div style="display: flex; flex-direction: column; justify-content: center; align-items: center; font-family: sans-serif; max-width: 800px">
                <div>
                    <div style="text-align: center;">
                        <h2>High-Quality Materials</h2>
                        <p class="paragraph2 pt-3">Crafted with precision and designed for durability, this product blends modern aesthetics with exceptional functionality.</p>
                    </div>

                    <div class="col-lg-6 my-5 d-none d-lg-block">
                        <img style="max-width: 100%" src="https://placehold.co/1200x800?text=Image+1" alt="">
                    </div>
                </div>

                <div style="display: flex; flex-direction: row; justify-content: center; align-items: center; margin-top: 2rem">
                    <div style="margin-right: 1rem">
                        <h2>Reliable Performance</h2>
                        <p class="paragraph2 pt-3">Engineered for everyday use, this product ensures consistent performance and exceptional reliability in all conditions.</p>
                    </div>
                    <div>
                        <img style="max-width: 100%" src="https://placehold.co/1200x800?text=Image+2" alt="">
                    </div>
                </div>
            </div>
        ',
    ]);

    $test_query = \RemoteDataBlocks\Config\Query\HttpQuery::from_array( [
        'data_source' => $test_data_source,
        'query_runner' => $test_query_runner,
        'output_schema' => [
            'is_collection' => false,
            'type' => [
                'header' => [
                    'name' => 'Header',
                    'path' => '$["header"]',
                    'type' => 'string',
                ],
                'blurb' => [
                    'name' => 'Blurb',
                    'path' => '$["blurb"]',
                    'type' => 'string',
                ],
                'htmlContent' => [
                    'name' => 'HTML Content',
                    'path' => '$["htmlContent"]',
                    'type' => 'html',
                ],
            ],
        ],
    ] );

    register_remote_data_block( [
        'title' => 'Test HTML API',
        'render_query' => [
            'query' => $test_query,
        ],
    ] );
} );

function get_query_runner_with_response( array $response_data ) {
    return new class($response_data) extends \RemoteDataBlocks\Config\QueryRunner\QueryRunner {
        private $response_data;

        public function __construct( $response_data ) {
            $this->response_data = $response_data;
        }

        protected function get_raw_response_data( \RemoteDataBlocks\Config\Query\HttpQueryInterface $query, array $input_variables ): array|\WP_Error {
            return [
                'metadata' => [
                    'age' => 100,
                    'status_code' => 200,
                ],
                'response_data' => $this->response_data,
            ];
        }
    };
}
```

Next, follow these instructions to test the block

1. Open up a new post.
2. Insert the "Test HTML API" block.
3. In the pattern selection box, you should see an auto-generated pattern with remote HTML visible. Select the default pattern.
4. On the post, see that the HTML has been bound to a Remote HTML block.

    image

6. Save the post and preview it.
7. Verify that the remote HTML is rendered on the frontend.

### Other things to test

1. Try inserting a Remote HTML block outside of a remote data container. You should see a placeholder for the unconnected block:

    image

    That's because we need to show something here when no data is connected, or else we'd have an empty space with an invisible block. Happy to improve the text here if you have any ideas.

2. Try modifying the mocked data source, e.g.

    ```diff
    -   <h2>High-Quality Materials</h2>
    +   <h2>Decent Materials</h2>
    ```

    Then save the PHP file and refresh the frontend page. Verify that the data is dynamically updated.

3. Try removing the entire mocked data source. Refresh the frontend page, and verify that the saved post markup continues to show up as static data.
4. Try it with a real API that returns HTML markup. The main thing necessary to map to HTML is adding an output with type `html` to the query:

    ```php
    $query = HttpQuery::from_array( [
        /* ... */
        'output_schema' => [
            'myHtml' => [
                'name' => 'My HTML',
                'path' => '$["pathToMarkup"]',
                'type' => 'html',
            ],
        ],
    ] );
    ```

    Note that this PR won't handle issues like missing CSS that the HTML may rely on for layout. This could be solved by queueing CSS onto the frontend, or inlined like the example data for testing.


src/blocks/remote-data-html/render.php
```diff
@@ -0,0 +1,24 @@
...
+<div <?php echo get_block_wrapper_attributes(); ?>>
+	<?php
+	/**
+	 * @psalm-suppress UndefinedGlobalVariable
+	 * $block is provided by WordPress for rendering, see header comments
+	 */
+	$binding_value = BlockBindings::get_value( $source_args, $block );
+
+	// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- This output is specifically for raw HTML.
+	echo $binding_value;
```
chriszarate: We still need to escape with `wp_kses_post` or similar. Remote data is untrusted

src/blocks/remote-data-container/components/BlockBindingControls.tsx
```diff
@@ -161,6 +162,20 @@ export function BlockBindingControls( props: BlockBindingControlsProps ) {
...
+					<BlockBindingFieldControl
+						availableBindings={ availableBindings }
+						fieldTypes={ HTML_FIELD_TYPES }
+						label="Content"
+						target="content"
+						updateFieldBinding={ updateFieldBinding }
+						value={ contentField }
+					/>
+				</>
+			);
```
alecgeatches, line 178: Something we could consider in the future is making this filter-driven so plugin users can also reuse our remote data binding sidebar. Right now, we need to add supported blocks to this section of code.

src/blocks/remote-data-html/render.php
alecgeatches: Good thinking, completed in https://github.com/Automattic/remote-data-blocks/pull/314/commits/01839371e6df57da9394765f5d98962f1db19252.

src/blocks/remote-data-html/edit.tsx
```diff
@@ -0,0 +1,99 @@
...
+		</div>
+	);
+}
+
+const PlaceholderInstructions = () => {
+	return (
+		<Placeholder
+			label={ __( 'Remote HTML' ) }
+			instructions={ __(
+				'Place this block in a remote data container and bind to an attribute to view HTML.'
```
alecgeatches: @smithjw1 Any thoughts on this copy? This is what's shown in the editor if a user inserts a remote data block that's unbound:

Screenshot 2025-01-21 at 3 01 32 PM

smithjw1: Ok, so the issue is we have a new block that only works for very specific parents. It sounds like you need some sort of block governance to prevent its insertion.

Sort of kidding, sort of not. Ideally, we just make this particular thing impossible with the Block Governance plugin, or at least suggest it.

I would make the text more explicit about what happens:

> This block only works when placed inside a remote data block container and bound to an attribute. This block will be ignored as currently configured.

src/blocks/remote-data-container/components/BlockBindingControls.tsx
```diff
@@ -161,6 +162,20 @@ export function BlockBindingControls( props: BlockBindingControlsProps ) {
...
 				</>
 			);
+
+		case 'remote-data-blocks/remote-html':
+			return (
+				<>
+					<BlockBindingFieldControl
+						availableBindings={ availableBindings }
+						fieldTypes={ HTML_FIELD_TYPES }
+						label="Content"
```
chriszarate:
```suggestion
						label="Raw HTML"
```

src/blocks/remote-data-container/config/constants.ts
```diff
@@ -16,6 +16,7 @@ export const CONTAINER_CLASS_NAME = getClassName( 'container' );
 
 export const BUTTON_TEXT_FIELD_TYPES = [ 'button_text' ];
 export const BUTTON_URL_FIELD_TYPES = [ 'button_url' ];
+export const HTML_FIELD_TYPES = [ 'html' ];
```
chriszarate, line 19: Need to remove `'html'` from `TEXT_FIELD_TYPES` below

src/blocks/remote-data-html/block.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "https://schemas.wp.org/trunk/block.json",
+  "apiVersion": 3,
+  "name": "remote-data-blocks/remote-html",
```
chriszarate: Name mismatch between this field and folder name

src/blocks/remote-data-html/edit.tsx
```diff
@@ -0,0 +1,99 @@
...
+const DEFAULT_STYLES = `
+	html,body,:root {
+		margin: 0 !important;
+		padding: 0 !important;
+		overflow: visible !important;
+		min-height: auto !important;
+	}
+`;
+
+export default function RemoteDataHTML(
```
chriszarate: We are currently inconsistent, but I'd prefer named exports instead of default exports

https://basarat.gitbook.io/typescript/main-1/defaultisbad

src/blocks/remote-data-html/edit.tsx
```diff
@@ -0,0 +1,99 @@
...
+	props: BlockEditProps< RemoteDataInnerBlockAttributes >
+): JSX.Element {
+	const { attributes, setAttributes, isSelected } = props;
+	const blockProps = useBlockProps();
+
+	// Pass the content provided by bindings to save() via setAttributes with changed content
+	const { content } = attributes;
+	useEffect( () => {
+		setAttributes( { content } );
+	}, [ content, setAttributes ] );
```
chriszarate: o_O

What does benefit does this provide? I think some memoization inside `setAttributes` is the only thing saving you from infinite loop.

src/blocks/remote-data-html/edit.tsx
```diff
@@ -0,0 +1,99 @@
...
+		[]
+	);
+
+	const styles = useMemo(
+		() => [
+			DEFAULT_STYLES,
+			...transformStyles( ( settingStyles ?? [] ).filter( ( style: EditorStyle ) => style.css ) ),
+		],
+		[ settingStyles ]
+	);
```
chriszarate: `transformStyles` is internally cached, don't need memo

```suggestion
	const styles = [
		DEFAULT_STYLES,
		...transformStyles( ( settingStyles ?? [] ).filter( ( style: EditorStyle ) => style.css ) ),
	];
```

src/blocks/remote-data-html/edit.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+	const hasBindings = attributes?.metadata?.bindings?.content !== undefined;
+
+	const contentString = content?.toString() ?? '';
+	const hasContent = contentString.length > 0;
```
chriszarate: If there is a binding but the HTML is empty, do you still want to render the placeholder? That could be confusing.

src/blocks/remote-data-html/edit.tsx
```diff
@@ -0,0 +1,99 @@
...
+	return (
+		<div { ...blockProps }>
+			<SandBox
+				html={ contentString }
+				styles={ styles }
+				title={ __( 'Remote Data Block HTML Preview' ) }
+				tabIndex={ -1 }
+			/>
+			{
+				/* Similar to core/html, add an overlay register click events. */
```
chriszarate: Could you expand on why this is useful? To prevent interaction?

src/blocks/remote-data-html/editor.scss
```diff
@@ -0,0 +1,14 @@
+// Styled after core/html. This overlay will allow clicking into the block
+// when HTML is present and bypass iframe bubbling issues.
```
chriszarate: accurate?
```suggestion
// when HTML is present and intercept all mouse events.
```

package.json
```diff
@@ -74,6 +75,7 @@
     "@wordpress/dependency-extraction-webpack-plugin": "^6.14.0",
     "@wordpress/dom-ready": "4.14.0",
     "@wordpress/e2e-test-utils-playwright": "^1.14.0",
+    "@wordpress/edit-post": "8.14.0",
```
chriszarate: Seems like we aren't even using `@wordpress/edit-post` anymore? Remove.

chriszarate:
@alecgeatches Could you add some tests and docs?

For tests, it would be good to have coverage of `render.php`:
1. It picks up the binding from `metadata`
2. It executes the query
3. It falls back to the sourced `content` attribute

For docs, having a short video showing it being inserted would be awesome .

src/blocks/remote-data-html/block.json
alecgeatches: Fixed them both to be `remote-html`, I don't think "data" was adding anything here.

src/blocks/remote-data-html/edit.tsx
alecgeatches: The `hasBindings` and `if` check work together to ensure if this block is bound to an empty string, we won't show the placeholder:

https://github.com/Automattic/remote-data-blocks/blob/f2b5d4f94df4572a25ceb799dd24dc0b2f6b587f/src/blocks/remote-html/edit.tsx#L61

https://github.com/Automattic/remote-data-blocks/blob/f2b5d4f94df4572a25ceb799dd24dc0b2f6b587f/src/blocks/remote-html/edit.tsx#L66-L72

src/blocks/remote-data-html/editor.scss
alecgeatches: That's correct, updated and expanded in https://github.com/Automattic/remote-data-blocks/pull/314/commits/ff57a4db2069e9feb27a8546f158a45497dcc214!

src/blocks/remote-data-html/edit.tsx
alecgeatches: It's to allow interaction in the editor, mostly. I've expanded that comment here: https://github.com/Automattic/remote-data-blocks/commit/ff57a4db2069e9feb27a8546f158a45497dcc214

package.json
alecgeatches: Good catch. Removed!

src/blocks/remote-data-html/edit.tsx
alecgeatches: Updated! Here's how it looks:

remote-html-empty

src/blocks/remote-data-html/edit.tsx
alecgeatches: This is an interesting problem. The reason this was originally added is because the `save()` component was not receiving any `content` attribute, and so we were unable to persist the raw HTML within the block for content fallback. Calling `setAttributes()` directly fixed it.

But why is that? I think it's because our wrapper binding component [passes merged attributes](https://github.com/Automattic/remote-data-blocks/blob/8b4c2d9089198cb9fa790f13ab92884e83a7d8d0/src/blocks/remote-data-container/filters/withBlockBinding.tsx#L133) to the Edit component. So `Edit` views `attributes.content` with a value, but the `Save` component does not have a similar wrapper to know what our injected `content` value is. In testing, explicitly calling `setAttributes()` on content change made it "real" so that `Save` could view the value.

You're correct that this looks wild and maybe is only working through a fluke of memoization. I'm going to look for another workaround.

alecgeatches: This is probably why we don't see fallback data persisted in some scenarios described in https://github.com/Automattic/remote-data-blocks/pull/258 under **Notes** -> Point 2. I'm going to test adding a `save` wrapper function with some shared code to see if that fixes things.

chriszarate: The main reason we merge attributes is because of the new client-side `getValue()` API, which would otherwise overwrite the attributes. Might be worth checking to see where @maxschmeling is in his implementation of `getValues()`, perhaps this can all come together nicely.

alecgeatches: @chriszarate I looked into @maxschmeling's work in https://github.com/Automattic/remote-data-blocks/tree/add/get-values-impl, but there's still some debug code in that branch and I wasn't able to test if it fixed the issue with a simple merge.

I decided to skip adding a `Save` wrapper mentioned above, because this would add some new complexity that we won't need after a `getValues()` implementation works. Instead, I made it a *lot* less confusing to behold by adding a separate `saveContent` attribute with no circular dependency and a longer explanation of why it's being used.

This should work as a stopgap until `getValues()` works, and won't introduce any new big changes.

I'd love another review!

alecgeatches:
> For docs, having a short video showing it being inserted would be awesome .

@chriszarate I forgot about this part. Will add this shortly.

tests/integration/RDBTestCase.php
```diff
@@ -0,0 +1,174 @@
...
+use RemoteDataBlocks\Config\Query\HttpQueryInterface;
+use RemoteDataBlocks\Config\QueryRunner\QueryRunner;
+use RemoteDataBlocks\Editor\BlockManagement\BlockRegistration;
+use RemoteDataBlocks\Editor\BlockManagement\ConfigStore;
+
+class RDBTestCase extends WP_UnitTestCase {
+	// Query mocking
+
+	/**
+	 * @psalm-suppress PossiblyUnusedParam
```
alecgeatches: Psalm decided [every method parameter in this class](https://github.com/Automattic/remote-data-blocks/actions/runs/12957874218/job/36147086879) was "possibly unused" for some reason I wasn't able to determine, so I used these to quiet the failures.

chriszarate: Weird, I "fixed" it by using a regular PHPUnit `TestCase` instead of `WP_UnitTestCase`. Guessing Psalm somehow needs to know about `WP_UnitTestCase` via some stubs or something?

At some point we'll probably need the factory methods that `WP_UnitTestCase` class provides, but ok for now.

chriszarate (approved):
Thanks for adding tests and docs, they're great!!!

tests/integration/RDBTestCase.php
alecgeatches: I'm actually surprised this still works! `WP_UnitTestCase` does all sorts of stuff (in [`WP_UnitTestCase`](https://github.com/WordPress/wordpress-develop/blob/a82874058f58575dbba64ce09b6dcbd43ccf5fdc/tests/phpunit/includes/testcase.php) -> [`WP_UnitTestCase_Base`](https://github.com/WordPress/wordpress-develop/blob/a82874058f58575dbba64ce09b6dcbd43ccf5fdc/tests/phpunit/includes/abstract-testcase.php)) beyond factory methods that actually seem useful for integration tests, like a lot of [test cleanup](https://github.com/WordPress/wordpress-develop/blob/a82874058f58575dbba64ce09b6dcbd43ccf5fdc/tests/phpunit/includes/abstract-testcase.php#L84-L97). I'm going to look further into this to see if I can add whatever stubs we need to keep static analysis happy.

alecgeatches: @chriszarate Managed to make psalm happy with `WP_UnitTestCase` using [`preloadClasses="true"`](https://psalm.dev/docs/running_psalm/configuration/#stubs) in the stubs section of https://github.com/Automattic/remote-data-blocks/pull/314/commits/dd7d4aa0bccd07735aa4ec0a1301a9715f185d28. I'm honestly not sure why this fixes `PossiblyUnusedParam`, but the docs mention abstract classes (which `WP_UnitTestCase` is) and it works now. Weird!

PR #317, opened 2025-01-22
add block registration snippet generation api

shekharnwagh:
# Description

> JIRA: [CAFE-1132](https://vipjira.atlassian.net/browse/CAFE-1132)

Adds a REST API endpoint that generates code snippets for registering Remote Data Blocks. The endpoint is accessible at:

`GET wp-json/remote-data-blocks/v1/data-sources/snippets/:data_source_uuid`

The endpoint returns ready-to-use PHP code snippets for registering blocks that connect to configured data sources. It currently supports:

- Shopify
- Airtable
- Google Sheets
- Salesforce B2C

The snippets are generated by substituting values from the data source configuration (created in Remote Data Blocks settings) into predefined templates. The generated code can be used as-is without modification to register functional blocks that fetch and display remote data.

# Testing

- Create a Shopify and/or Airtable data source
- From the data sources list view in Remote Data Blocks settings page, copy the UUID of the data source you want to test
- Make a request to the endpoint, e.g. `GET wp-json/remote-data-blocks/v1/data-sources/snippets/{{data_source_uuid}}`
- The response will be a PHP code snippet that you can use to register a block
- Copy the code to example directory, e.g. `example/shopify/<block_name>/register.php`
- Add the following line to `remote-data-blocks.php` file: `require_once __DIR__ . '/example/shopify/<block_name>/register.php';`
- Verify that the block is registered and working as expected


inc/Integrations/Airtable/templates/block_registration.template
```diff
@@ -0,0 +1,81 @@
...
+ * - AIRTABLE_: The fields of the table to make available for the block.
+ */
+
+namespace RemoteDataBlocks\Snippets\Airtable;
+
+use RemoteDataBlocks\Integrations\Airtable\AirtableDataSource;
+use RemoteDataBlocks\WpdbStorage\DataSourceCrud;
+use RemoteDataBlocks\Config\Query\HttpQuery;
+
+function register_airtable_events_block(): void {
```
brookewp: The Airtable block might not be for an event so we could make the name more generic for the template: 
```suggestion
function register_airtable_block(): void {
```

brookewp:
As I've been working with the array for the UI, I noticed a couple of things: 

If a data source has multiple snippets and they are copied to separate files, there will be errors because of the same uuid and function name (e.g. `register_google_sheets_blocks`). So we might want to add a template variable for the function name to make it unique. And we could include info about either combining these into one file or editing the `uuid` for each (or we could generate new uuids to make them all unique). I also ran into an error when following the testing steps because the uuid is the same as the data source created in the UI. Also, do we need to add the access token/a placeholder for that? 
 
Alternatively, if we want to match the auto-generated code 1:1, we would need to combine them initially. I think we'd need to isolate the declarations/statements (perhaps as the first array item) and then add the remaining function snippets to the array. For example, for Google Sheets: 

```tsx
[
"<?php declare(strict_types = 1);

		namespace RemoteDataBlocks\Snippets\Google;
		
		use RemoteDataBlocks\Config\Query\HttpQuery;
		use RemoteDataBlocks\Integrations\Google\Sheets\GoogleSheetsDataSource;
		use RemoteDataBlocks\WpdbStorage\DataSourceCrud;"
, 
"function register_google_sheets_blocks...", 
];
```

And then loop over the snippets containing the functions (if they all have unique names) or combine them all within one function. So they would end up with just one file to copy. What do you think? 

Otherwise, if we stick with an array of snippets, it would be helpful for the UI if we could pass along the name of the table/sheet etc so we can distinguish each snippet. For example: 

```diff
diff --git a/inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php b/inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php
index 369d77d8..5c9ffde0 100644
--- a/inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php
+++ b/inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php
@@ -171,7 +171,10 @@ class GoogleSheetsIntegration {
 				'{{SHEETS_OUTPUT_SCHEMA_MAPPINGS}}' => self::get_output_schema_mappings_snippet( $sheet ),
 			] );
 
-			$snippets[] = $snippet;
+			$snippets[] = [
+				'name' => $sheet['name'],
+				'snippet' => $snippet
+			];
 		}
 
 		return $snippets;
diff --git a/inc/Snippets/Snippets.php b/inc/Snippets/Snippets.php
index 3555f38d..cff04ac5 100644
--- a/inc/Snippets/Snippets.php
+++ b/inc/Snippets/Snippets.php
@@ -37,7 +37,10 @@ class Snippets {
 				return new WP_Error( 'invalid_service', __( 'Invalid service', 'remote-data-blocks' ) );
 		}
 
-		return array_map( [ __CLASS__, 'strip_template_comments' ], $snippets );
+		return array_map(function($item) {
+			$item['snippet'] = self::strip_template_comments($item['snippet']);
+			return $item;
+		}, $snippets);
 	}
 
 	protected static function strip_template_comments( string $content ): string {
```

So we could have three tabs with the corresponding snippets: 

Screenshot 2025-01-29 at 2 07 13 PM

If this makes sense, I can make the changes if helpful. 😄

shekharnwagh:
@brookewp Thank you for taking a look!

> If a data source has multiple snippets and they are copied to separate files, there will be errors because of the same uuid and function name (e.g. register_google_sheets_blocks)

I've fixed this in [02b03ab](https://github.com/Automattic/remote-data-blocks/pull/317/commits/02b03ab2abde07697d209cd67df926c8a1c38e90) - a unique function name is generated using the data source display name + sheet/table name if present

> If we stick with an array of snippets, it would be helpful for the UI if we could pass along the name of the table/sheet etc so we can distinguish each snippet. 

I am leaning towards this and have pushed this in [a8a60e5](https://github.com/Automattic/remote-data-blocks/pull/317/commits/a8a60e554032d700a85a9b36e39613c404f274d0). Let me know if this works well with the front-end PR.

brookewp:
> I am leaning towards this and have pushed this in a8a60e5. Let me know if this works well with the front-end PR.

Yes, thank you! It works perfectly 🙂

inc/Integrations/Google/Sheets/templates/block_registration.template
```diff
@@ -0,0 +1,83 @@
...
+ * - SHEETS_OUTPUT_SCHEMA_MAPPINGS: The field mappings for the sheet data.
+ */
+
+namespace RemoteDataBlocks\Snippets\Google;
+
+use RemoteDataBlocks\Config\Query\HttpQuery;
+use RemoteDataBlocks\Integrations\Google\Sheets\GoogleSheetsDataSource;
+use RemoteDataBlocks\WpdbStorage\DataSourceCrud;
+
+function register_google_sheets__{{BLOCK_REG_FN_SLUG}}__blocks(): void {
```
brookewp, line 21: Was it intentional to have double underscores in the function name? E.g.: 

`register_google_sheets__sheets__formatted__blocks()`

shekharnwagh: Yes, to have a clear separation between values that can contain multiple words, we use double underscores. For example, if the spreadsheet name is `Cities` and the sheet name is `United States`, the function name would be `register_google_sheets__cities__united_states__blocks`. This only serves aesthetic purposes, so open to reverting to single underscores if we prefer that.

maxschmeling: I personally prefer single underscore there... just looks strange to me. No other reason. The reason for adding it makes sense though.

maxschmeling: We can tweak in a future PR as we iterate on this.

maxschmeling (approved):

maxschmeling:
I think we'll refine this approach in future changes, but this is a good working start. Let's get it merged so we can get the UI in there too.

PR #320, opened 2025-01-22
Update/documentation/extending last looks

smithjw1:
This finishes my pass at the documentation for now.

Thanks all for the thoughtful reviews!

### DIFF
--- a/docs/extending/block-patterns.md
+++ b/docs/extending/block-patterns.md
@@ -1,10 +1,16 @@
 # Block patterns
 
-Patterns allow you to represent your remote data in different ways. The plugin registers an unstyled block pattern any time you register a remote data block either in the WordPress admin or with `register_remote_data_block`. You can create additional patterns in the WordPress Site Editor or programmatically by passing a `patterns` property to your block options.
+Patterns allow you to represent your remote data in different ways. 
+
+The plugin registers an unstyled block pattern any time you register a remote data block either in the WordPress admin or with `register_remote_data_block`. 
+
+You can create additional patterns in the WordPress Site Editor or programmatically by passing a `patterns` property to your block options.
 
 You cannot edit the default pattern, but you can duplicate it and make changes.
 
-We recommend duplicating the default pattern and then making changes in the Site Editor. Once you've created your preferred pattern, you can associate it with the block in the `register_remote_data_block` call. If you want to make the pattern uneditable in the Site Editor, you can copy the block markup to a file and commit it to your repository.
+We recommend duplicating the default pattern and then making changes in the Site Editor. Once you've created your preferred pattern, you can associate it with the block in the `register_remote_data_block` call.
+
+If you want to make the pattern uneditable in the Site Editor, you can copy the block markup to a file and commit it to your repository.
 
 ## Example
 
@@ -19,6 +25,7 @@ We recommend duplicating the default pattern and then making changes in the Site
     <!-- /wp:paragraph -->
 </div>
 <!-- /wp:group -->
+ ```
 
 You could save this file as `my-pattern.html` in the same directory as the code that registers your block.
 
--- a/docs/extending/block-patterns.md
+++ b/docs/extending/block-patterns.md
@@ -1,4 +1,4 @@
-# Block patterns
+# Block Patterns
 
 Patterns allow you to represent your remote data in different ways. The plugin registers an unstyled block pattern anytime you register a remote data block either in the WordPress admin or with `register_remote_data_block`. You can create additional patterns in the WordPress Site Editor or programmatically by passing a `patterns` property to your block options.
 
@@ -9,27 +9,11 @@ You cannot edit the default pattern, but you can duplicate it and make changes. 
 ```html
 <!-- wp:group {"layout":{"type":"constrained"}} -->
 <div class="wp-block-group">
-	<!-- wp:heading {"metadata":{"bindings":{"content":{"source":"remote-data/binding","args":{"field":"title"}}}}} -->
-	<h2 class="wp-block-heading"></h2>
-	<!-- /wp:heading -->
-	<!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"remote-data/binding","args":{"field":"description"}}}}} -->
-	<p></p>
-	<!-- /wp:paragraph -->
+    <!-- wp:heading {"metadata":{"bindings":{"content":{"source":"remote-data/binding","args":{"field":"title"}}}}} -->
+    <h2 class="wp-block-heading"></h2>
+    <!-- /wp:heading -->
+    <!-- wp:paragraph {"metadata":{"bindings":{"content":{"source":"remote-data/binding","args":{"field":"description"}}}}} -->
+    <p></p>
+    <!-- /wp:paragraph -->
 </div>
 <!-- /wp:group -->
-```
-
-You could save this file as `my-pattern.html` in the same directory as the code that registers your block.
-
-```php
-register_remote_data_block( [
-    'title' => 'My Remote Data Block',
-    'render_query' => [ /* ... */ ],
-    'patterns' => [
-        [
-            'title' => 'My Pattern',
-            'content' => file_get_contents( __DIR__ . '/my-pattern.html' ),
-        ],
-    ],
-] );
-```
--- a/docs/extending/block-patterns.md
+++ b/docs/extending/block-patterns.md
@@ -1,8 +1,10 @@
 # Block patterns
 
-Patterns allow you to represent your remote data in different ways. The plugin registers an unstyled block pattern anytime you register a remote data block either in the WordPress admin or with `register_remote_data_block`. You can create additional patterns in the WordPress Site Editor or programmatically by passing a `patterns` property to your block options.
+Patterns allow you to represent your remote data in different ways. The plugin registers an unstyled block pattern any time you register a remote data block either in the WordPress admin or with `register_remote_data_block`. You can create additional patterns in the WordPress Site Editor or programmatically by passing a `patterns` property to your block options.
 
-You cannot edit the default pattern, but you can duplicate it and make changes. We recommend starting with a duplicate and then making changes in the Site Editor. If you want to lock the pattern down from further edits, copy the block markup from the editor and associate the pattern via code.
+You cannot edit the default pattern, but you can duplicate it and make changes.
+
+We recommend duplicating the default pattern and then making changes in the Site Editor. Once you've created your preferred pattern, you can associate it with the block in the `register_remote_data_block` call. If you want to make the pattern uneditable in the Site Editor, you can copy the block markup to a file and commit it to your repository.
 
 ## Example
 
@@ -17,3 +19,18 @@ You cannot edit the default pattern, but you can duplicate it and make changes. 
     <!-- /wp:paragraph -->
 </div>
 <!-- /wp:group -->
+
+You could save this file as `my-pattern.html` in the same directory as the code that registers your block.
+
+```php
+register_remote_data_block( [
+    'title' => 'My Remote Data Block', 
+    'render_query' => [ /* ... */ ],
+    'patterns' => [
+        [
+            'title' => 'My Pattern',
+            'content' => file_get_contents( __DIR__ . '/my-pattern.html' ),
+        ],
+    ],
+] );
+```
--- a/docs/extending/block-patterns.md
+++ b/docs/extending/block-patterns.md
@@ -1,4 +1,4 @@
-# Block Patterns
+# Block patterns
 
 Patterns allow you to represent your remote data in different ways. The plugin registers an unstyled block pattern anytime you register a remote data block either in the WordPress admin or with `register_remote_data_block`. You can create additional patterns in the WordPress Site Editor or programmatically by passing a `patterns` property to your block options.
 
--- a/docs/extending/block-patterns.md
+++ b/docs/extending/block-patterns.md
@@ -1,8 +1,8 @@
 # Block patterns
 
-Patterns allow you to represent your remote data in different ways. 
+Patterns allow you to represent your remote data in different ways.
 
-The plugin registers an unstyled block pattern any time you register a remote data block either in the WordPress admin or with `register_remote_data_block`. 
+The plugin registers an unstyled block pattern any time you register a remote data block either in the WordPress admin or with `register_remote_data_block`.
 
 You can create additional patterns in the WordPress Site Editor or programmatically by passing a `patterns` property to your block options.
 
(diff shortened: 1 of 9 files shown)


docs/quickstart.md
```diff
@@ -1,6 +1,8 @@
 # Quickstart
 
-The most basic way to see Remote Data Blocks in action is to launch the plugin in WordPress Playground. You can do this by clicking the "Launch in WordPress Playground" badge in the [README](../README.md).
+The most basic way to see Remote Data Blocks in action is to launch the plugin in WordPress Playground.
```
stoph: _basic_ sounds derogatory. How about _easiest_ or _quickest_?

docs/extending/overrides.md
```diff
@@ -1,8 +1,10 @@
...
 
-The actual implementation of overrides is left to you. Here is an example of an override that modifies the query input variables based on a query variable (supplied by a rewrite rule):
+If you have multiple instnaces of the same remote data block in a piece of content, each instance can have a different number of overrides enabled. The default behavior is no overrides.
+
+The actual implementation of overrides is left to you. Here is an example of an override that modifies the query input variables based on a URL query string variable. You would use this to build a one "single product page" in the WordPress admin that would be able to display any product, based on the product id. The example takes advantage of the [`add_rewrite_rule`](https://developer.wordpress.org/reference/functions/add_rewrite_rule/) method and [`query_vars`](https://developer.wordpress.org/reference/hooks/query_vars/) filter that are built into WordPress.
```
stoph: "The actual implementation of overrides is left to you." Doesn't sound like a helpful or necessary line.

Technically it is "based on a URL query string variable." but that is obfuscated by the rewrite rule. i.e. this example doesn't expect the URL in the browser to have a query string (`/products/?acme_product_id=123456`). How about... "based on a value in the URL. For example, to load product ID 123456, this would define the URL as `https://example.com/product/123456`." (or something better written than that ;)

Also, since you mention this is for a "single product page," we should use `/product/` as opposed to `/products/` in the rewrite rule.

docs/extending/overrides.md
```diff
@@ -1,8 +1,10 @@
 # Overrides
 
-Overrides provide a way to customize the behavior of remote data blocks on a per-block basis. You can use them to modify the underlying query input variables, adjust the query response, or change the caching behavior. Overrides are defined when you register a remote data block and can be enabled or disabled via the block settings in the editor.
+Overrides provide a way to customize the behavior of remote data blocks on a per-block basis. You can use them to modify the underlying query input variables, adjust the query response, or change the caching behavior. Overrides are defined when you register a remote data block and can be enabled or disabled via the block settings in the WordPress editor. 
 
-The actual implementation of overrides is left to you. Here is an example of an override that modifies the query input variables based on a query variable (supplied by a rewrite rule):
+If you have multiple instnaces of the same remote data block in a piece of content, each instance can have a different number of overrides enabled. The default behavior is no overrides.
```
stoph: instnaces -> instances

docs/extending/block-patterns.md
```diff
@@ -1,4 +1,4 @@
-# Block patterns
+# Block Patterns
```
chriszarate: Pretty sure our house style is "Sentence case" for top-level titles, but Yoli could confirm.

docs/extending/block-patterns.md
```diff
@@ -9,27 +9,11 @@ You cannot edit the default pattern, but you can duplicate it and make changes.
...
+    <p></p>
+    <!-- /wp:paragraph -->
 </div>
 <!-- /wp:group -->
-```
-
-You could save this file as `my-pattern.html` in the same directory as the code that registers your block.
-
-```php
-register_remote_data_block( [
```
chriszarate: Why is this example removed?

docs/extending/block-registration.md
```diff
@@ -142,3 +142,4 @@ Here you can see the input variable of `search_terms` is used in the endpoint me
 ### `patterns`: array (optional)
 
 [Block patterns](./block-patterns.md) allow you to customize the display of your remote data.
+````
```
chriszarate: remove

docs/extending/data-source.md
```diff
@@ -53,13 +53,29 @@ $zipcode_query = HttpQuery::from_array( [
 ])
 ```
 
-In this case the `data_source` has a built in `get_endpoint()` method. You could also retrieve the endpoint this way:
+In this case the `data_source` has a built in `get_endpoint()` method. Other configuration options can be retireved directly. If you had a Data Source config like:
```
chriszarate:
```suggestion
In this case the `data_source` has a built in `get_endpoint()` method. Other configuration options can be retireved directly. If you had a data source config like:
```

docs/extending/data-source.md
```diff
@@ -53,13 +53,29 @@ $zipcode_query = HttpQuery::from_array( [
...
+
+```php
+$comp_data_source = HttpDataSource::from_array( [
+    'service_config' => [
+        '__version' => 1,
+        'display_name' => 'More Complicated Example API',
+        'endpoint' => 'https://api.complexexample.com/',
+		'some_identifier' => 'table-2',
```
chriszarate: Not sure what this is, it's invalid syntax

docs/extending/data-source.md
```diff
@@ -53,13 +53,29 @@ $zipcode_query = HttpQuery::from_array( [
 ])
 ```
 
-In this case the `data_source` has a built in `get_endpoint()` method. You could also retrieve the endpoint this way:
+In this case the `data_source` has a built in `get_endpoint()` method. Other configuration options can be retireved directly. If you had a Data Source config like:
+
+```php
+$comp_data_source = HttpDataSource::from_array( [
```
chriszarate: Can we just standardize on a variable name like `$example_data_source` or `$data_source`? It's not particularly complicated

docs/extending/data-source.md
```diff
@@ -53,13 +53,29 @@ $zipcode_query = HttpQuery::from_array( [
...
+You could retrieve `some_identifier` directly:
 
 ```php
 $zipcode_query = HttpQuery::from_array( [
-    'data_source' => $zipcode_data_source,
-    'endpoint' => function ( array $input_variables ) use ( $zipcode_data_source ): string {
-        return  $zipcode_data_source->to_array()['service_config']['endpoint'] . $input_variables['zip_code'];
+    'data_source' => $comp_data_source,
+    'endpoint' => function ( array $input_variables ) use ( $comp_data_source ): string {
+        return  return $comp_data_source->get_endpoint() . $zipcode_data_source->to_array()['service_config']['some_identifier'] . "/" . $input_variables['search'];
```
chriszarate: Mentioned on the other PR, but I don't think we should promote this pattern. It constrains us in future changes and this alternate approach is unnecessary. They can just hold a reference to `some_identifier` in their own code and pass it in via `use()`. This should definitely not be idiomatic.

docs/extending/index.md
```diff
@@ -23,10 +23,10 @@ Providing a custom data source, query, or query runner gives you complete contro
 
 Here are some detailed overviews of these classes with notes on how and why to extend them:
 
-- [Data source](data-source.md)
+- [Data Source](data-source.md)
 - [Query](query.md)
 
-Once you've defined your data source and queries, you can [register a remote data block](block-registration.md) that uses them. That block can use a [pattern](block-patterns.md) for display. You can also use [overrides](./overrides.md) to dynamically select the displayed content.
+Once you've defined your data source and queries, you can [register a remote data block](block-registration.md) that uses them. That block can use a [pattern](block-patterns.md) for display. You can also use [overrides](./overrides.md) to dynamically select the appropriate data source or query.
```
chriszarate: I don't think this is an accurate description of overrides.

docs/extending/overrides.md
```diff
@@ -1,8 +1,14 @@
 # Overrides
 
-Overrides provide a way to customize the behavior of remote data blocks on a per-block basis. You can use them to modify the underlying query input variables, adjust the query response, or change the caching behavior. Overrides are defined when you register a remote data block and can be enabled or disabled via the block settings in the editor.
+Overrides provide a way to customize the behavior of remote data blocks on a per-block basis. You can use them to modify the underlying query input variables, adjust the query response, or change the caching behavior. Overrides are defined when you register a remote data block and can be enabled or disabled via the block settings in the WordPress editor. 
 
-The actual implementation of overrides is left to you. Here is an example of an override that modifies the query input variables based on a query variable (supplied by a rewrite rule):
+If you have multiple instances of the same remote data block in a piece of content, each instance can have a different number of overrides enabled. The default behavior is no overrides.
```
chriszarate:
```suggestion
If you have multiple instances of the same remote data block in a piece of content, each instance can have different overrides enabled. By default, no overrides are enabled.
```

docs/extending/overrides.md
```diff
@@ -1,8 +1,14 @@
...
+
+Here is an example of an override that modifies usesthe query input variables based on the URL.
```
chriszarate:
```suggestion
Here is an example of an override that modifies the query input variables based on the URL. 
```

docs/extending/overrides.md
```diff
@@ -1,8 +1,14 @@
...
+Overrides provide a way to customize the behavior of remote data blocks on a per-block basis. You can use them to modify the underlying query input variables, adjust the query response, or change the caching behavior. Overrides are defined when you register a remote data block and can be enabled or disabled via the block settings in the WordPress editor. 
 
-The actual implementation of overrides is left to you. Here is an example of an override that modifies the query input variables based on a query variable (supplied by a rewrite rule):
+If you have multiple instances of the same remote data block in a piece of content, each instance can have a different number of overrides enabled. The default behavior is no overrides.
+
+Here is an example of an override that modifies usesthe query input variables based on the URL. 
+
+You would use this to build a one "single product page" in the WordPress admin that would be able to display any product, based on the product id at a url like: https://example.com/product/123456
```
chriszarate:
```suggestion
You could use this approach to build a "product page" that displays any product using an ID from a URL, e.g.: https://example.com/products/123456
```

docs/extending/overrides.md
```diff
@@ -1,8 +1,14 @@
...
+
+The example takes advantage of the [`add_rewrite_rule`](https://developer.wordpress.org/reference/functions/add_rewrite_rule/) method and [`query_vars`](https://developer.wordpress.org/reference/hooks/query_vars/) filter that are built into WordPress.
```
chriszarate:
```suggestion
The example takes advantage of the [`add_rewrite_rule`](https://developer.wordpress.org/reference/functions/add_rewrite_rule/) function and the [`query_vars`](https://developer.wordpress.org/reference/hooks/query_vars/) filter that are built into WordPress.
```

docs/extending/overrides.md
```diff
@@ -19,7 +25,7 @@ register_remote_data_block( [
     ],
 ] );
 
-add_rewrite_rule( '^products/([0-9]+)/?', 'index.php?pagename=products&acme_product_id=$matches[1]', 'top' );
+add_rewrite_rule( '^product/([0-9]+)/?', 'index.php?pagename=products&acme_product_id=$matches[1]', 'top' );
```
chriszarate:
```suggestion
add_rewrite_rule( '^products/([0-9]+)/?', 'index.php?pagename=products&acme_product_id=$matches[1]', 'top' );
```

docs/extending/overrides.md
```diff
@@ -39,6 +45,8 @@ add_filter( 'remote_data_blocks_query_input_variables', function ( array $input_
 }, 10, 2 );
 ```
 
+As you can see the `remote_data_blocks_query_input_variables` filter is passed a list of enabled overrides. You need to add logic to identify which filters are enabled and act accordingly.
```
chriszarate:
```suggestion
As you can see, the `remote_data_blocks_query_input_variables` filter is passed a list of enabled overrides. You need to add logic to identify which filters are enabled and act accordingly.
```

docs/extending/query-output_schema.md
```diff
@@ -1,6 +1,6 @@
-# The Query `output_schema` property
+# The Query `output_schema` Property
 
-The `output_schema` property is where your data shape definition happens. It should be created with care and requires updates anytime the incoming response changes.
+The `output_schema` property is where your data shape definition happens. It should be created with care and requires updates any time the incoming response changes.
```
chriszarate:
```suggestion
The `output_schema` property is where your data shape definition happens. It should be created with care and requires updates whenever the incoming response changes.
```

docs/extending/query-output_schema.md
```diff
@@ -98,6 +98,7 @@ An example of collection JSON can be found in the [Chicago Institue of Art examp
 			"timestamp": "2025-01-14T22:26:21-06:00"
 		},
 		{
+
```
chriszarate:
```suggestion
```

docs/extending/query.md
```diff
@@ -6,7 +6,7 @@ A common approach is to define a data source on the settings screen and then com
 
 ## HttpQuery
 
-Most HTTP-powered APIs can be queried using an `HttpQuery`. Here's an example of a query for US ZIP code data. This examples assumes you have configured the data source in the UI, and have the UUID.
+Most HTTP-powered APIs can be queried using an `HttpQuery`. Here's an example of a query for US ZIP
```
chriszarate: Seems accidental, this is important context.
```suggestion
Most HTTP-powered APIs can be queried using an `HttpQuery`. Here's an example of a query for US ZIP code data. This examples assumes you have configured the data source in the UI, and have the UUID.
```

docs/extending/overrides.md
chriszarate: > we should use /product/ as opposed to /products/ in the rewrite rule.

@stoph I just saw this, so my earlier suggested changes contradict this. FWIW, I see lots of examples of both singular and plural slugs in product pages. Some plural ones:

- https://www.petsmart.com/dog/treats/training-treats/wiggles-and-wags-freeze-dried-chicken-crunchy-dog-treats-81254.html
- https://www.hallmark.com/cards/greeting-cards/peanuts-snoopy-and-woodstock-hearts-3d-pop-up-valentines-day-card-549VWF8004.html
- https://www.michelinman.com/auto/tires/michelin-defender-ltx-platinum
- https://docs.wpvip.com/guidebooks/get-started/

docs/extending/block-patterns.md
smithjw1: I don't recall. If I had to guess it's because:

1. The example code doesn't work because it doesn't have a render query.
2. The example block markup won't work in many cases, and editing it directly is fragile.

I think the suggestion to duplicate the pattern and then copy that markup is better, but the example is still useful. I'll revise.

docs/extending/data-source.md
smithjw1: Sorry I'm not sure what's invalid, could you say more?

docs/extending/data-source.md
smithjw1: When I look at the Airtable Data Source we add a `base` property to the data source definition, that's was my inspiration.

I think your suggesting that additional configuration should happen outside of service_config, so something like:

```php
$data_source = HttpDataSource::from_array( [
    'service_config' => [
        '__version' => 1,
        'display_name' => 'More Complicated Example API',
        'endpoint' => 'https://api.complexexample.com/',
        'request_headers' => [
            'Content-Type' => 'application/json',
            'X-Api-Key' => MY_API_KEY_CONSTANT,
        ],
    ],
] );

$my_config = [
 'some_identifier' => 'id-123'
]

$zipcode_query = HttpQuery::from_array( [
    'data_source' => $data_source,
    'endpoint' => function ( array $input_variables ) use ( $data_source, $my_config ): string {
        return  return $data_source->get_endpoint() . $my_config['some_identifier'] . "/" . $input_variables['search'];
    },
])
```

docs/extending/overrides.md
smithjw1: I agree in part with these changes and disagree in part. I've revised, can you give it another look?

docs/extending/data-source.md
chriszarate: Yes, exactly

docs/extending/data-source.md
chriszarate: It's been cleaned up, ignore

docs/extending/data-source.md
chriszarate: I applied your suggestion in https://github.com/Automattic/remote-data-blocks/pull/320/commits/a781b4cc19527e960673855485c1d39a9ebc43d7

chriszarate (approved):

PR #321, opened 2025-01-22
[UI] Data source code snippets

brookewp:
WIP, based on #317 

https://github.com/user-attachments/assets/99ae03bd-e357-4285-b553-491d91a0d0b6

### DIFF
--- a/src/data-sources/components/CodeSnippet.tsx
+++ b/src/data-sources/components/CodeSnippet.tsx
@@ -1,4 +1,4 @@
-import { Button, ExternalLink } from '@wordpress/components';
+import { Button } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { copy } from '@wordpress/icons';
 import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';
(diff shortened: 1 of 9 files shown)


src/data-sources/components/CodeSnippet.tsx
```diff
@@ -0,0 +1,52 @@
+import { Button } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { copy } from '@wordpress/icons';
+import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';
+import { coy } from 'react-syntax-highlighter/dist/esm/styles/prism';
```
chriszarate: If this is the only file we're using from this dependency, can we just pull it into our code base?

https://github.com/react-syntax-highlighter/react-syntax-highlighter/blob/master/src/styles/prism/coy.js

Does this increase the bundle size or is the rest tree-shaken out?

chriszarate (approved):
Just a question on the dependency addition

src/data-sources/components/CodeSnippet.tsx
brookewp: Good point @chriszarate! 

We are using the Prism component in addition to the 'coy' style: 
`import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter';`

However, they offer a 'light build' to decrease the size: https://github.com/react-syntax-highlighter/react-syntax-highlighter#light-build

So I've swapped to that in: [`b01d71c` (#321)](https://github.com/Automattic/remote-data-blocks/pull/321/commits/b01d71c22fc007cc77486d4e14a736349f535597) which does decrease the bundle size 🎉

chriszarate (approved):

PR #322, opened 2025-01-22
Add Blocks step with toggle to control auto-generation of blocks

brookewp:
Adds optional `enable_blocks` boolean (default `true`) to config of integrations. This provides the option of disabling the auto-generation of blocks.

https://github.com/user-attachments/assets/5971b205-3d4a-4be9-99fe-baa252575431

This could be expanded to specify block types, as seen in the design here:

Screenshot 2025-01-22 at 2 05 07 PM


inc/Integrations/Airtable/AirtableIntegration.php
```diff
@@ -16,8 +16,13 @@ public static function register_blocks(): void {
 			REMOTE_DATA_BLOCKS_AIRTABLE_SERVICE
 		);
 
-		foreach ( $data_source_configs as $config ) {
+		foreach ( $data_source_configs as $config ) {
```
maxschmeling:
```suggestion
		foreach ( $data_source_configs as $config ) {
```

inc/Integrations/Airtable/AirtableIntegration.php
```diff
@@ -16,8 +16,13 @@ public static function register_blocks(): void {
...
+		foreach ( $data_source_configs as $config ) {       
 			$data_source = AirtableDataSource::from_array( $config );
+
+			if ( false === $config['service_config']['enable_blocks'] ) {
+				continue;
+			}
+			
 			self::register_blocks_for_airtable_data_source( $data_source );
 			self::register_loop_blocks_for_airtable_data_source( $data_source );
 		}
```
maxschmeling: I personally prefer this style, but it's just personal preference. Take it or leave it.
```suggestion

			if ( false !== $config['service_config']['enable_blocks'] ) {
				self::register_blocks_for_airtable_data_source( $data_source );
				self::register_loop_blocks_for_airtable_data_source( $data_source );
			}
		}
```

inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php
```diff
@@ -18,6 +18,11 @@ public static function register_blocks(): void {
 
 		foreach ( $data_source_configs as $config ) {
 			$data_source = GoogleSheetsDataSource::from_array( $config );
+	
+			if ( false === $config['service_config']['enable_blocks'] ) {
+				continue;
+			}
+			
 			self::register_blocks_for_google_sheets_data_source( $data_source );
 			self::register_loop_blocks_for_google_sheets_data_source( $data_source );
```
maxschmeling: Same thing here. Take it or leave it
```suggestion
	
			if ( false !== $config['service_config']['enable_blocks'] ) {
				self::register_blocks_for_google_sheets_data_source( $data_source );
				self::register_loop_blocks_for_google_sheets_data_source( $data_source );
			}
```

inc/Integrations/SalesforceB2C/SalesforceB2CIntegration.php
```diff
@@ -17,6 +17,10 @@ public static function register_blocks(): void {
 
 		foreach ( $data_source_configs as $config ) {
 			$data_source = SalesforceB2CDataSource::from_array( $config );
+	
+			if ( false === $config['service_config']['enable_blocks'] ) {
+				continue;
+			}
 			self::register_blocks_for_salesforce_data_source( $data_source );
```
maxschmeling:
```suggestion
	
			if ( false !== $config['service_config']['enable_blocks'] ) {
				self::register_blocks_for_salesforce_data_source( $data_source );
			}
```

inc/Integrations/Shopify/ShopifyIntegration.php
```diff
@@ -17,6 +17,11 @@ public static function register_blocks(): void {
 
 		foreach ( $data_source_configs as $config ) {
 			$data_source = ShopifyDataSource::from_array( $config );
+
+			if ( false === $config['service_config']['enable_blocks'] ) {
+				continue;
+			}
+				
 			self::register_blocks_for_shopify_data_source( $data_source );
```
maxschmeling:
```suggestion

			if ( false === $config['service_config']['enable_blocks'] ) {
				self::register_blocks_for_shopify_data_source( $data_source );
			}
```

src/data-sources/http/HttpSettings.tsx
```diff
@@ -92,6 +92,11 @@ export const HttpSettings = ( { mode, uuid, config }: SettingsComponentProps< Ht
 
 				<HttpAuthSettingsInput auth={ state.auth } onChange={ handleAuthOnChange } />
 			</DataSourceForm.Setup>
+			<DataSourceForm.Blocks
+				handleOnChange={ handleOnChange }
+				// TO DO
+				hasEnabledBlocks={ true }
+			/>
```
maxschmeling: We can just remove this, right? No auto-generated config here.

maxschmeling (dismissed):

inc/Integrations/Airtable/AirtableDataSource.php
```diff
@@ -17,6 +17,7 @@ protected static function get_service_config_schema(): array {
 				'id' => Types::string(),
 				'name' => Types::nullable( Types::string() ),
 			] ),
+			'enable_blocks' => Types::nullable( Types::boolean(), true ),
```
chriszarate: This second arg (`true`) doesn't do anything. Were you trying to set a default value? Also, this should trigger a lint / type error. 🤔

inc/Integrations/Airtable/AirtableIntegration.php
```diff
@@ -16,8 +16,13 @@ public static function register_blocks(): void {
 			REMOTE_DATA_BLOCKS_AIRTABLE_SERVICE
 		);
 
-		foreach ( $data_source_configs as $config ) {
+		foreach ( $data_source_configs as $config ) {       
 			$data_source = AirtableDataSource::from_array( $config );
+
+			if ( false === $config['service_config']['enable_blocks'] ) {
```
chriszarate: You probably want to coalesce to a default value here, this will generate a `WARNING` (sometimes elevated to `ERROR`) if `enable_blocks` is not defined. E.g., if another integration is added or the config is filtered

```suggestion
			if ( false === $config['service_config']['enable_blocks'] ?? true ) {
```

inc/Integrations/Airtable/AirtableIntegration.php
chriszarate: FWIW, I love early returns / breaks! As I read the code, it helps me "drop" logical gates that I don't need to think about anymore.

```
if ( condition_a ) {
  return;
}

if ( condition_b ) {
  return;
}

// Yay, all code from here out is guaranteed not to have to
// deal with `condition_a` or `condition_b`!!
```

maxschmeling:
Found a validation issue when actually creating and using Airtable blocks.

image

inc/Integrations/Airtable/AirtableIntegration.php
```diff
@@ -30,26 +35,26 @@ public static function register_blocks_for_airtable_data_source(
...
+								],
 							],
 						],
-					],
-					$block_overrides
-				)
-			);
+						$block_overrides
+					)
+				);
```
chriszarate, line 59: I think the indent got off kilter here

inc/Integrations/Shopify/ShopifyIntegration.php
chriszarate: Max's code suggestion here is the wrong way 'round

src/data-sources/components/DataSourceForm.tsx
```diff
@@ -70,15 +71,45 @@ const DataSourceForm = ( { children, onSave }: DataSourceFormProps ) => {
 	const steps = Children.toArray( children );
 	const singleStep = steps.length === 1 || screen === 'editDataSource';
 
-	const stepHeadings = [ 'Setup', 'Scope' ];
+	const getStepHeadings = () => {
+		const headings = [ __( 'Setup' ) ]; // First step is always Setup
+
+		if (
+			Children.toArray( children ).some(
```
chriszarate:
```suggestion
			steps.some(
```

src/data-sources/components/DataSourceForm.tsx
```diff
@@ -70,15 +71,45 @@ const DataSourceForm = ( { children, onSave }: DataSourceFormProps ) => {
...
+				child =>
+					isValidElement( child ) &&
+					( child.type === DataSourceForm.Scope || child.type === 'Scope' )
+			)
+		) {
+			headings.push( __( 'Scope' ) );
+		}
+
+		if (
+			Children.toArray( children ).some(
```
chriszarate:
```suggestion
			steps.some(
```

src/data-sources/components/DataSourceForm.tsx
```diff
@@ -70,15 +71,45 @@ const DataSourceForm = ( { children, onSave }: DataSourceFormProps ) => {
...
+		) {
+			headings.push( __( 'Blocks' ) );
+		}
+
+		return headings;
+	};
+
+	const stepHeadings = getStepHeadings();
 
 	const canProceedToNextStep = () => {
```
chriszarate, line 95: Curious, why is this a function (with no args) instead of just a variable computed in the render scope?

src/data-sources/components/DataSourceForm.tsx
```diff
@@ -70,15 +71,45 @@ const DataSourceForm = ( { children, onSave }: DataSourceFormProps ) => {
...
 		if ( ! isValidElement< { canProceed?: boolean; displayName: string; uuid: string } >( step ) ) {
 			return false;
 		}
 		const { canProceed, displayName, uuid } = step.props;
-		return Boolean( canProceed ) && ( currentStep !== 1 || canUseDisplayName( displayName, uuid ) );
+		return (
+			( Boolean( canProceed ) &&
+				( currentStep !== 1 || canUseDisplayName( displayName, uuid ) ) ) ||
+			step.type === DataSourceForm.Blocks
+		);
```
chriszarate: Getting a little gnarly, I find it helpful to split out complex boolean logic into semantic variables (u can probably think of better names).
```suggestion
		const isBlocksStep = step.type === DataSourceForm.Blocks;
		const noConflicts = Boolean( canProceed ) &&
			( currentStep !== 1 || canUseDisplayName( displayName, uuid ) );
		
		return isBlocksStep || noConflicts;
```

src/data-sources/components/DataSourceForm.tsx
```diff
@@ -332,7 +357,53 @@ const DataSourceFormScope = ( {
...
 };
 
+const DataSourceFormBlocks = ( {
+	handleOnChange,
+	hasEnabledBlocks,
+}: {
+	handleOnChange: ( key: string, value: boolean ) => void;
+	hasEnabledBlocks: boolean;
+} ) => {
+	const [ enableBlocks, setEnableBlocks ] = useState( hasEnabledBlocks );
```
chriszarate: Why are we setting local state here instead of relying on the prop?

inc/Integrations/Airtable/AirtableIntegration.php
ingeniumed: I'd lean towards early returns/breaks as well. It makes the rest of the code so much easier to read and work with, as the problematic conditions have been dealt with already

src/data-sources/http/HttpSettings.tsx
brookewp: Yup, good catch! Was in auto-pilot here

inc/Integrations/Airtable/AirtableIntegration.php
brookewp: Thanks @ingeniumed for chiming in so I didn't have to chose a side 😄 majority wins!

inc/Integrations/Airtable/AirtableDataSource.php
brookewp: > Were you trying to set a default value? 

Yes I was. I probably don't have a proper lint setup for PHP, but then I guess our tests didn't catch it either

src/data-sources/components/DataSourceForm.tsx
brookewp: Because I like to overcomplicate things 🥲 Thanks for pointing this out! Fixed here: [`ae21968` (#322)](https://github.com/Automattic/remote-data-blocks/pull/322/commits/ae219689eab60ef948979ec8687253fab6efaa45)

src/data-sources/components/DataSourceForm.tsx
brookewp: [This reply]( https://github.com/Automattic/remote-data-blocks/pull/322#discussion_r1936237909) can work here too

inc/Integrations/Airtable/AirtableIntegration.php
brookewp: Yet `No fixable errors were found` when running `npm run lint:phpcs:fix` 😕

brookewp:
>Found a validation issue when actually creating and using Airtable blocks.

@maxschmeling are you seeing this in settings after registering an airtable block and when using it in the editor? I've tried to replicate both of those scenarios but haven't come across the error

chriszarate:
> @maxschmeling are you seeing this in settings after registering an airtable block and when using it in the editor? I've tried to replicate both of those scenarios but haven't come across the error

I think this was likely an existing pre-PR Airtable data source without an `enable_blocks` property, and before you added the coalescing to a default value.

inc/Integrations/Airtable/AirtableIntegration.php
maxschmeling: I love simple early returns also, but for some reason a `continue` in a loop just feels more complex.

But I'm fine with it.

inc/Integrations/Airtable/AirtableIntegration.php
```diff
@@ -18,6 +18,11 @@ public static function register_blocks(): void {
 
 		foreach ( $data_source_configs as $config ) {
 			$data_source = AirtableDataSource::from_array( $config );
+
+			if ( false === $config['service_config']['enable_blocks'] ?? true ) {
```
chriszarate: Fun fact! Order of operations means that the comparison is done before the coalescing. Need parentheses here to avoid a bug:
```suggestion
			if ( false === ( $config['service_config']['enable_blocks'] ?? true ) ) {
```

https://www.php.net/manual/en/language.operators.precedence.php

inc/Integrations/Airtable/AirtableIntegration.php
chriszarate: 🤷 I can look at it, still off though

inc/Integrations/Google/Sheets/GoogleSheetsDataSource.php
```diff
@@ -27,6 +27,7 @@ protected static function get_service_config_schema(): array {
 				'universe_domain' => Types::string(),
 			] ),
 			'display_name' => Types::string(),
+			'enable_blocks' => Types::nullable( Types::boolean(), true ),
```
chriszarate: remove unused arg

inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php
```diff
@@ -18,6 +18,11 @@ public static function register_blocks(): void {
 
 		foreach ( $data_source_configs as $config ) {
 			$data_source = GoogleSheetsDataSource::from_array( $config );
+	
+			if ( false === $config['service_config']['enable_blocks'] ) {
```
chriszarate: add coalescing here too (with parens)

chriszarate: repeat for other instances

inc/Integrations/Google/Sheets/GoogleSheetsDataSource.php
chriszarate: repeat for other instances

chriszarate (changes requested):
Just a little cleanup needed! Looking good

inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php
maxschmeling: This should be fine. An explicit `false` value is the only time we want to skip out. If it's `undefined` we want to default to enabling blocks.

inc/Integrations/Airtable/AirtableIntegration.php
maxschmeling: There actually isn't a bug, but the coalescing isn't necessary.

If it's `null` then the comparison is `false` and it will not `continue`, which is the desired behavior.
If it's `false`, then the coalescing won't change it, and it will go in and hit `continue`.
If it's `true` then the comparison is `false`, and still, it will not `continue`.

I'm not sure if it's better to use `array_key_exists` before getting the value or not...

chriszarate: There's no logical bug, but there is a bug in intended behavior. The coalescing is necessary to prevent a warning when `$config['service_config']['enable_blocks']` is undefined (like the one [you reported](https://github.com/Automattic/remote-data-blocks/pull/322#issuecomment-2622838714) on this PR). In some environments, these warnings can be escalated to errors which effectively interrupt.

That's part of the semantic value of the `??` operator, because it swallows these warnings while being less verbose than `isset` or `array_key_exists`. But the order of operations matters.

inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php
chriszarate: Yes, but will generate a warning without coalescing or explicit check.

inc/Integrations/Airtable/AirtableIntegration.php
brookewp: TIL, thank you!

chriszarate (approved):
🚀

PR #325, opened 2025-01-23
Update useRemoteData to be more explicit about state management

chriszarate:
My work towards pagination has led to some necessary refactoring, so I'm opening them in separate PRs as much as possible.

This PR:

- Makes state management of `useRemoteData` much more explicit. Previously you could await the result of `execute`, which is non-idiomatic in React—everything should be state-driven. Now you can essentially pass in your own state, which makes our `Edit` component _much_ simpler!
- Renames some props to make their purpose more clear:
  - `PatternSelection#onSelectPattern`
  - `Placeholder#onSelect`
- Moves some state and logic from `usePatterns` that didn't really belong there because it is specific to `Edit`.

In general, `Edit` is much more straightforward and easier to follow. And it paves the way for pagination updates to `useRemoteData`.

### DIFF
--- a/src/blocks/remote-data-container/edit.tsx
+++ b/src/blocks/remote-data-container/edit.tsx
@@ -1,7 +1,7 @@
-import { InspectorControls, useBlockProps } from '@wordpress/block-editor';
+import { BlockPattern, InspectorControls, useBlockProps } from '@wordpress/block-editor';
 import { BlockEditProps } from '@wordpress/blocks';
 import { Spinner } from '@wordpress/components';
-import { useEffect, useState } from '@wordpress/element';
+import { useState } from '@wordpress/element';
 
 import { InnerBlocks } from '@/blocks/remote-data-container/components/InnerBlocks';
 import { DataPanel } from '@/blocks/remote-data-container/components/panels/DataPanel';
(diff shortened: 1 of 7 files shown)


src/blocks/remote-data-container/edit.tsx
```diff
@@ -32,120 +32,97 @@ export function Edit( props: BlockEditProps< RemoteDataBlockAttributes > ) {
...
+	function onSelectPattern( pattern: BlockPattern ): void {
+		insertPatternBlocks( pattern );
+		setShowPatternSelection( false );
+	}
 
-	function refreshRemoteData() {
-		if ( ! props.attributes.remoteData?.queryInput ) {
-			if ( hasInputVariables ) {
+	function onSelectRemoteData( queryInput: RemoteDataQueryInput ): void {
+		void fetch( queryInput ).then( () => {
```
brookewp, line 69: Would it make sense for `queryInput` to be an array? For a future where we allow multiple selection

maxschmeling: `queryInput` is an object that could have an array inside of it. (It's `Record<string, unknown>` right now)

chriszarate: > Would it make sense for queryInput to be an array? For a future where we allow multiple selection

We would need a construct around that dispatches multiple queries for each selected item. Will be fun!

brookewp (approved):
✅ This tests well and is a lot easier to understand 🎉

PR #326, opened 2025-01-23
Correctly apply default input variable values

chriszarate:
Our schema allows users to supply a `default_value` for input variables, but it isn't respected. It's useful, so we should.

### DIFF
--- a/tests/inc/Config/QueryRunnerTest.php
+++ b/tests/inc/Config/QueryRunnerTest.php
@@ -347,4 +347,46 @@ class QueryRunnerTest extends TestCase {
 		$this->assertCount( 1, $result['results'] );
 		$this->assertSame( $expected_result, $result['results'][0] );
 	}
+
+	public function testQueryRunnerAppliesDefaultInputVariables(): void {
+		$query = MockQuery::from_array( [
+			'data_source' => $this->http_data_source,
+			'endpoint' => function ( array $input_variables ): string {
+				return sprintf(
+					'https://example.com/api?foo=%s&baz=%s',
+					$input_variables['foo'] ?? 'MISSING',
+					$input_variables['baz'] ?? 'MISSING',
+				);
+			},
+			'input_schema' => [
+				'baz' => [
+					'name' => 'Baz',
+					'type' => 'string',
+				],
+				'foo' => [
+					'default_value' => 'bar',
+					'name' => 'Foo',
+					'type' => 'string',
+				],
+			],
+			'query_runner' => new QueryRunner( $this->http_client ),
+		] );
+
+		$response_body = $this->createMock( \Psr\Http\Message\StreamInterface::class );
+		$response = new Response( 200, [], $response_body );
+
+		$this
+			->http_client
+			->expects( $this->exactly( 1 ) )
+			->method( 'request' )
+			->willReturn( $response )
+			->with( 'GET', '/api?foo=bar&baz=MISSING' );
+
+		$result = $query->execute( [] );
+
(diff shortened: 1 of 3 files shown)


tests/inc/Config/QueryRunnerTest.php
```diff
@@ -347,4 +347,46 @@ public function testExecuteSuccessfulResponseWithObjectResponseData(): void {
...
+
+		$response_body = $this->createMock( \Psr\Http\Message\StreamInterface::class );
+		$response = new Response( 200, [], $response_body );
+
+		$this
+			->http_client
+			->expects( $this->exactly( 1 ) )
+			->method( 'request' )
+			->willReturn( $response )
+			->with( 'GET', '/api?foo=bar&baz=MISSING' );
```
brookewp, line 383: Looking at this test as an example, it looks like a use case for `search_terms`. So I tried to implement it into the art institute example: 

```php
	$search_art_query = HttpQuery::from_array([
		'data_source' => $aic_data_source,
			$query = $input_variables['search_terms'];
			$endpoint = $aic_data_source->get_endpoint() . '/search';

			return add_query_arg( [ 'q' => $query ], $endpoint );
		},
		'input_schema' => [
			'search_terms' => [
				'default_value' => 'stars',
				'name' => 'Search Terms',
				'type' => 'string',
			],
		],
```

Which I thought would give me this endpoint initially: `https://api.artic.edu/api/v1/artworks/search?q=stars`
But `search_terms` is an empty string. Curious to learn about what I'm missing here.

chriszarate: > it looks like a use case for search_terms

It's not per se, just a use case for optional input variables in generally.

I'm guessing your query is passing in a `search_terms` input variable with an empty string value, which is not overridden by the default value since it's a defined value. Try to add a different input variable and use in the endpoint callback, then observe the result in the Query Monitor or `qm` property of REST response.

```php
			$query = $input_variables['search_terms'];
			$endpoint = $aic_data_source->get_endpoint() . '/search';

			return add_query_arg( [ 'foo' => $input_variables['foo'], 'q' => $query ], $endpoint );
		},
		'input_schema' => [
			'search_terms' => [
				'name' => 'Search Terms',
				'type' => 'string',
			],
			'foo' => [
				'default_value' => 'bar',
				'name' => 'Foo',
				'type' => 'string',
			],
		],
```

brookewp: Thanks! I understand this better after seeing the use case in your other PR: https://github.com/Automattic/remote-data-blocks/pull/338/commits/6acf40762a89abce02689cea9f6f06413db59bf1

So I'm glad an example is being added for this. 😄

maxschmeling (approved):
The only nit I have is the order of the input schema variables in the mock query... the order is different than the use of them in the endpoint so it made me pause for a second.

Otherwise the test looks good.

I do wonder if we should validate and throw an error when there is no default and no value was provided. Future improvement there?

chriszarate:
> The only nit I have is the order of the input schema variables in the mock query... the order is different than the use of them in the endpoint so it made me pause for a second.

Makes sense, I'll fix up later.

> I do wonder if we should validate and throw an error when there is no default and no value was provided. Future improvement there?

You mean when the variable is marked as `required`? We probably should. We currently don't do any validation of the input variables or the API response. This is mainly for fault tolerance, but also because at runtime we don't have a good way to surface errors. You're right that some limited validation would be useful if we can surface it in a way that helps users.

PR #327, opened 2025-01-23
adding issue templates

smithjw1:
We want to provide some guidance for those who submit issues to make sure we have the information we need.


chriszarate:
Looks good, can you please run `npm run format`

smithjw1:
> Looks good, can you please run `npm run format`

Done. Should we have this as some sort of git hook to make sure it happens? We don't have it documented anywhere.

chriszarate:
> Should we have this as some sort of git hook to make sure it happens?

We have the ability to opt-in to these hooks via `npx husky install`, but they are surprisingly controversial among developers.

PR #329, opened 2025-01-24
updating links to absolute

smithjw1:
So RDB.com works links, outside the docs folder to GitHub have to be absolute.


chriszarate (approved):
Absolute-ly great!

PR #337, opened 2025-01-27
Add `ui:search_input` input variable

chriszarate:
This PR is based on #325.

Currently, we require a "magically named" input variable for search queries so that we know how to map search input to the query. This pattern works and is documented, but deserves formalization so that this declarative pattern can be reused—for pagination and other future capabilities. 

Instead of a magic name, we will introduce a special type with a `ui:*` naming convention. After improvements to `useRemoteData`, we incorporate search capabilities into it.


chriszarate:
@brookewp Can you help me resolve the merge conflicts here?

chriszarate:
> @brookewp Can you help me resolve the merge conflicts here?

I think I got them resolved here, but could use you help with testing. And the conflicts look worse on the pagination PR

PR #338, opened 2025-01-27
Implement query pagination

chriszarate:
Based on #337. ~Still working on docs, but feel free to review!~ Ready for review!


brookewp:
This is an exciting addition! 🎉 

I tested out the pagination page selection dropdown after searching and the data never loaded or timed out (maybe eventually it does but I couldn't wait forever):

https://github.com/user-attachments/assets/29133042-8107-4bd6-bb40-8a192c8d1933

I didn't run into this using the pagination dropdown on the initial un-searched set or when searching and manually clicking the next page button.

chriszarate:
@brookewp Thanks! FYI inspecting the network response is useful here. Above page 90 or so I get an error from the API:

> Client error: `GET https://api.artic.edu/api/v1/artworks/search?q=van+gogh&limit=12&page=90` resulted in a `403 Forbidden` response:\n{\"status\":403,\"error\":\"Invalid number of results\",\"detail\":\"You have requested too many results. Please refine your para...

It seems like the reported total items for this search (`2341`) does not align with reality, since `90 * 12 = 1,080`. 

It does point to the need to catch and surface these errors. What would be the best way to do this in DataViews / the block editor?

brookewp:
>It does point to the need to catch and surface these errors. What would be the best way to do this in DataViews / the block editor?

I don't think there is any way to do this in `DataViews` currently... 🤔 the only control we have is over the data, if there is no data, (an empty array passed to the `data` prop), it shows 'No results found' without the table, but then we lose the pagination:  

Storybook example

Storybook example here: [Storybook example link](https://wordpress.github.io/gutenberg/?path=/story/dataviews-dataviews--empty)

The only alternative I can think of right now is: 
1. Hide the page dropdown via CSS (I don't think it adds much value, as I don't think anyone would know which page contains which data. I can only think of consumers going to the end, but if we add filtering, they could reverse the order alphabetically to achieve the same result). 
2. Replace the table element via a portal 😬 to replace with our own error screen. 

What do you think?

ingeniumed:
Using [error notices](https://developer.wordpress.org/block-editor/reference-guides/data/data-core-notices/) could be a potential solution to this. It can either be a snackbar in a corner, or a banner at the very top so as to allow the editor itself to control it. I have been trying to experiment with this in dataviews and it's been tricky to get it working as the notice shows up on the editor while the dataview is covering it up. You also get a warning saying that you can't render a notice when the itemList is still being rendered. I suspect that's due to me putting the errorNotice within `itemList.tsx`. 

There is a prototype that @oandregal has made [here](https://github.com/wptrainingteam/devblog-dataviews-plugin/blob/main/src/App.js#L152) that might shed some light on how to use it correctly

brookewp:
Data doesn't show for any of my **Airtable** sources. However, I can select an item and then see it in pattern selection.

| Trunk  |  PR |
|  Screenshot 2025-02-04 at 3 51 22 PM  | Screenshot 2025-02-04 at 3 45 12 PM |

**Shopify** data loads, but no pagination and 'Item Price' column is missing (but shows when selecting like above): 

| Trunk  |  PR |
|  Screenshot 2025-02-04 at 3 49 42 PM | Screenshot 2025-02-04 at 3 46 12 PM  |

But of course the example I tested in my earlier comment still works 😕 : 

Screenshot 2025-02-04 at 3 48 40 PM

chriszarate:
> Data doesn't show for any of my **Airtable** sources. 

Thanks for testing! I've had to fix some pretty gnarly merge conflicts and I flubbed one. Fixed in https://github.com/Automattic/remote-data-blocks/pull/338/commits/0613bf3527dbc3f4cf119c86c070c090eedd442b.

That said, the logic with all the memos and the effect is really hard to follow, so I cleaned it up a bit in https://github.com/Automattic/remote-data-blocks/pull/338/commits/33f04d09757d554f6811c8afc390293afd7ac144 and https://github.com/Automattic/remote-data-blocks/pull/338/commits/eed54e2dedff4db8b5610738dbfec724f7605f56. It makes the diff a bit more severe but hopefully the resulting code is easy to follow.

src/blocks/remote-data-container/components/item-list/ItemList.tsx
```diff
@@ -1,135 +1,133 @@
...
+				field === mediaField
+					? ( { item }: { item: RemoteDataResult } ) => {
+							return (
+								<img alt={ ( item.image_alt as string ) ?? '' } src={ item[ field ] as string } />
+							);
+					  }
+					: undefined,
+			enableSorting: field !== mediaField,
+		};
+	} );
```
maxschmeling:
```suggestion
	const fields = fieldNames.map( field => ( {
		id: field,
		label: availableBindings[ field ]?.name ?? field,
		enableGlobalSearch: true,
		getValue: ( { item }: { item: RemoteDataResult } ) => item[ field ] as string,
		render:
			field === mediaField
				? ( { item }: { item: RemoteDataResult } ) => {
						return (
							<img alt={ ( item.image_alt as string ) ?? '' } src={ item[ field ] as string } />
						);
				  }
				: undefined,
		enableSorting: field !== mediaField,
	} ) );
```

src/blocks/remote-data-container/components/item-list/ItemList.tsx
```diff
@@ -1,135 +1,133 @@
...
-	useEffect( () => {
-		if ( view.search !== searchTerms ) {
-			setSearchTerms( view.search ?? '' );
-		}
-	}, [ view, searchTerms ] );
-
-	// filter, sort and paginate data
-	const { data: filteredData, paginationInfo } = useMemo( () => {
-		return filterSortAndPaginate( data ?? [], view, fields );
-	}, [ data, view ] );
```
brookewp, line 191: @chriszarate are we removing the search for upcoming changes in #337? 

I noticed because our integrated sources don't have pagination, since now it'll fallback to `data.length` if pagination isn't explicitly set: https://github.com/Automattic/remote-data-blocks/blob/eed54e2dedff4db8b5610738dbfec724f7605f56/src/blocks/remote-data-container/components/item-list/ItemList.tsx#L107-L108
https://github.com/Automattic/remote-data-blocks/blob/eed54e2dedff4db8b5610738dbfec724f7605f56/src/blocks/remote-data-container/components/item-list/ItemList.tsx#L154-L157

chriszarate: @brookewp The reasons for removing search (and filtering) are summarized here:

https://wp.me/pfvsEV-tn#behavior-change-for-searching-sorting-and-filtering-in-dataviews

After this PR is approved, I'll work on implementing pagination for our other integrations.

maxschmeling (approved):
The issues we were seeing are fixed. This looks great. I don't see anything major, though it'll be nice to get this in and see how it works as we use it more.

brookewp (approved):
Data is loading again 🎉 Looks good to me!

PR #340, opened 2025-01-28
Editor UI: Find fields by available binding types

brookewp:
In the DataViews component, `mediaField` provides an option to set an image for the row styled to fit the various layout types. It's [a property of view](https://github.com/WordPress/gutenberg/tree/trunk/packages/dataviews#view-object).

For example, in a grid: 
Screenshot 2025-01-28 at 1 49 34 PM

If we don't use it for an image, we'd have to implement the styling for each layout view as it's just treated as any other field/column in the table: 
Screenshot 2025-01-28 at 1 51 19 PM

When we do this, we have to exclude the fields (`titleField` is another like this) when they are defined or we'll end up with a duplicate field (not sure why this would be a desired feature so might be good to ask about).

https://github.com/Automattic/remote-data-blocks/blob/45152873496e2d512664cfbeaecee8c869ec5fe9/src/blocks/remote-data-container/components/item-list/ItemList.tsx#L117

Prior to this PR, we were finding the image by checking for fields with a property of 'image' or 'url', which isn't ideal. Instead, we can use the `type` from the schema. So if the type has been defined as `image_url`, we can display it properly.


ingeniumed (approved):
Tested this out and it works as expected. It solves the problem where sometime the image would be marked as hidden or there'd be a duplicate column.

PR #341, opened 2025-01-29
Update SFCC datasource 

ingeniumed:
### Description

This PR is meant to:

- Update the SFCC data source docs to add some steps in setting it up
- Add a 401 error code check when the client creds are wrong, so the plugin handles it gracefully
- Show the `no results` page upon hitting up the `choose` option and opening the search page. For SFCC this previously showed a spinning logo and that was because the results were set to 0. Ideally, if a call isn't done then the results shouldn't be set and if it's done and the result is empty then it should be treated as such. For SFCC, since no product ID is provided upon startup the search query is coming back empty as expected. Once a search is done, then something would be returned.

Screenshot of what the search page looks like when it's first opened now:

Screenshot 2025-01-30 at 9 09 28 am


docs/tutorials/salesforce-commerce.md
```diff
@@ -10,6 +10,9 @@ This tutorial will walk you through connecting a [Salesforce Commerce B2C](https
 2. Click on the "Connect new" button.
 3. Choose "Salesforce Commerce B2C" from the dropdown menu as the data source type.
 4. Name the data source. This name is only used for display purposes.
+5. Provide the merchant short code. This is the region-specific merchant identifier.
+6. Provide the organization ID. This is the organization idenitifer.
```
alecgeatches:
```suggestion
6. Provide the organization ID.
```

docs/tutorials/salesforce-commerce.md
```diff
@@ -10,6 +10,9 @@ This tutorial will walk you through connecting a [Salesforce Commerce B2C](https
...
+7. Provide the client ID and the client Secret. Ensure these are correct or else authentication will fail.
```
alecgeatches:
```suggestion
7. Provide the client ID and the client secret. Ensure these are correct or else authentication will fail.
```

alecgeatches (approved):
Tiny couple of docs suggestions. Looks good, tested locally and happy to see product images and improved empty UI. Thank you!

PR #345, opened 2025-01-29
Add function_exists() check to avoid fatal during vip dev-env setup

alecgeatches:
Similar to https://github.com/Automattic/remote-data-blocks/pull/220. We've encountered a similar fatal error during initial startup of `vip dev-env`:

```
Fatal error: Uncaught Error: Call to undefined function add_settings_error()
in /wp/wp-content/plugins/remote-data-blocks/inc/PluginSettings/PluginSettings.php on line 183
```

It seems like initial setup or database upgrades cause the `decrypt_option()` function to be called in a context outside of the regular admin UI where WordPress isn't fully loaded, although it is hard to replicate. 

From my testing, it doesn't really matter if we're unable to decrypt our option during this process, and adding a simple check for `add_settings_error()` to plugin code allowed it to work and does not appear introduce any other issues.


chriszarate:
There is also a call to `add_settings_error()` in hooked code on line 149, should that we wrapped in a check too?

alecgeatches:
> There is also a call to `add_settings_error()` in hooked code on line 149, should that we wrapped in a check too?

@chriszarate We haven't seen the same sort of startup error occur there, but that's a good catch we should cover as well. Addressed in https://github.com/Automattic/remote-data-blocks/pull/345/commits/d5b7dcef308f50542ea7efb170e89aafd243b07e.

chriszarate (approved):

PR #346, opened 2025-01-30
Register pattern only with a valid script handle

cbravobernal:
## Issue

Not having an editor script handle is causing a WSOD (white screen of death) when trying to log in `/wp-admin`.

Screenshot:

Screenshot 2025-01-30 at 12 34 28

## Fix:

Add a default string as a default to prevent this error.

### DIFF
--- a/inc/Editor/BlockManagement/BlockRegistration.php
+++ b/inc/Editor/BlockManagement/BlockRegistration.php
@@ -101,12 +101,17 @@ class BlockRegistration {
 
 		$block_type = register_block_type( $block_path, $block_options );
 
-		$script_handle = $block_type->editor_script_handles[0];
+		$script_handle = $block_type->editor_script_handles[0] ?? '';
 
-		// Register a default pattern that simply displays the available data.
-		$default_pattern_name = BlockPatterns::register_default_block_pattern( $block_name, $config['title'], $config['queries'][ ConfigRegistry::DISPLAY_QUERY_KEY ] );
-		$block_config['patterns']['default'] = $default_pattern_name;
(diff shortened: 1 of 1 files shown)


cbravobernal:
A npm run build fixes it 😅 , but still, is good not to crash the admin if someone forgets to.

chriszarate:
Thanks @cbravobernal!

When setting up a local development environment from the GitHub repo (instead of a directly using a [packaged release](https://github.com/Automattic/remote-data-blocks/releases)), you need to use `npm run dev` to start the local WP environment and build the required assets:

https://github.com/Automattic/remote-data-blocks/blob/trunk/docs/local-development.md

We can do a better job of detecting at runtime if this step was missed and point the user to helpful documentation. I'll merge this PR and follow up with a detection strategy. Thanks for raising this!

chriszarate (approved):

cbravobernal:
> Thanks @cbravobernal[](https://profiles.wordpress.org/cbravobernal)!
> 
> When setting up a local development environment from the GitHub repo (instead of a directly using a [packaged release](https://github.com/Automattic/remote-data-blocks/releases)), you need to use `npm run dev` to start the local WP environment and build the required assets:
> 
> https://github.com/Automattic/remote-data-blocks/blob/trunk/docs/local-development.md
> 
> We can do a better job of detecting at runtime if this step was missed and point the user to helpful documentation. I'll merge this PR and follow up with a detection strategy. Thanks for raising this!

I just saw it!
I checked that the repo was using `wp-env`, and I used the global package I have installed instead of `npm run dev` 😅

PR #347, opened 2025-02-03
Add live playground links using woocommerce as inspiration

ingeniumed:
### Description

This will add the ability to have a test playground link on every PR. This replicates what works for Woo.

### Testing

Hit the playground link in the comment and ensure that it works.


.github/workflows/scripts/generate-playground-blueprint.js
```diff
@@ -0,0 +1,110 @@
...
+		preferredVersions: {
+			php: '8.2',
+			wp: 'latest',
+		},
+
+		steps: [
+			{
+				step: 'setSiteOptions',
+				options: {
+					blogname: 'Remote Data Blocks (PR)',
```
stoph: Could we add the PR # reference to the site name so it's easy to reference while you're using it?

ingeniumed: Def can. My intention was to first get it working using the Woo code as a starting point and then start improving it with things like what you suggested, and also dropping the number of artifacts made to just 1 (PR instead of runner id using in the naming)

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,85 @@
...
+          unzip "$GITHUB_WORKSPACE/zips/remote-data-blocks.zip" -d "$GITHUB_WORKSPACE/unzips/remote-data-blocks"
+          echo "Unzipped zip file..."
+
+      - name: Create playground artifact
+        id: create-playground-artifact-step
+        uses: actions/upload-artifact@v4
+        with:
+          name: remote-data-blocks-${{ github.event.pull_request.number }}
+          path: unzips/remote-data-blocks
+          retention-days: 2
```
maxschmeling: Can this be longer? Maybe a week?

.github/workflows/scripts/generate-playground-blueprint.js
```diff
@@ -0,0 +1,104 @@
...
+			currentComment.body.includes( 'Test using WordPress Playground' )
+		) {
+			existingCommentId = currentComment.id;
+			break;
+		}
+	}
+
+	const defaultSchema = generateWordpressPlaygroundBlueprint( context.issue.number, artifactUrl );
+
+	const url = `https://playground.wordpress.net/#${ JSON.stringify( defaultSchema ) }`;
```
maxschmeling: Do we need to escape the stringified JSON so it doesn't break the markdown URL tag?

.github/workflows/pr-build-live-branch.yml
maxschmeling: https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#artifact-and-log-retention-policy

It can be up to 90 days.

ingeniumed: It def can be, I wanted to start with a lower number and go from there just to avoid any cost blowouts. A week def works

.github/workflows/scripts/generate-playground-blueprint.js
ingeniumed: Yip, that is the reason. Once it's escaped, it won't be broken. I just haven't done it yet as I've been trying to fix the playground link to the zip file first.

Note: Usually, I'd keep such PRs as draft PRs but this one made sense as a non-draft PR so it could be tested properly.

ingeniumed: Edit: Think escaping cause some weird quirks with playground that has to be dug into.

PR #352, opened 2025-02-04
Update field shortcode UI

brookewp:
This replaces the field shortcode modal with dropdown menus. This is more accessible, as before we had stacking modals when selecting a new item. 

| Before  | After |
| Screenshot 2025-02-04 at 3 20 47 PM  | Screenshot 2025-02-04 at 3 26 04 PM |

https://github.com/user-attachments/assets/24d7721f-a5bb-4dfb-b8b3-3725fee36796


inc/Editor/DataBinding/FieldShortcode.php
```diff
@@ -48,10 +48,10 @@ public static function render_frontend_fields( string $content ): string {
 			} else {
 				$block = [
 					'context' => [
-						BlockBindings::$context_name => [
```
brookewp, line 51: The following fatal error occurred when adding metadata values to a page/post via field shortcodes: 

```php
Fatal error: Uncaught Error: RemoteDataBlocks\Editor\BlockManagement\ConfigStore::get_block_configuration(): Argument #1 ($block_name) must be of type string, null given, called in /var/www/html/wp-content/plugins/remote-data-blocks/inc/Editor/DataBinding/BlockBindings.php on line 111
in /var/www/html/wp-content/plugins/remote-data-blocks/inc/Editor/BlockManagement/ConfigStore.php on line 45
```

This appears to be due to the nested context: 

```php
$block_context = [
    'remote-data-blocks/remoteData' => [
        'blockName' => 'remote-data-blocks/real-estateclien...',
        'queryInput' => []
    ]
];

```

Instead of other values which have this structure: 

```php
$block_context = [
        'blockName' => 'remote-data-blocks/real-estateclien...',
        'queryInput' => []
];
```

@chriszarate — I believe you added this so just wanted to check if it was necessary for something else. Also, are we safe to add `int` as a type in `get_meta_field_value`? When I removed the nested array, I then ran into an error due to expecting a string rather than int.

chriszarate: I'm not sure I follow, but I pushed up https://github.com/Automattic/remote-data-blocks/pull/352/commits/59fbcb086e9d7d1fa3bea33cc07e304d486e84a9 which should address the issue you were experiencing. Before that commit, the field shortcode working in the editor but failing on the frontend:

```
Missing block context for block binding remote-data-blocks/remoteData remote-data-blocks/remoteData (block: unknown; operation: unknown)
```

chriszarate: > Also, are we safe to add int as a type in get_meta_field_value? When I removed the nested array, I then ran into an error due to expecting a string rather than int.

Yup, fine, good catch

chriszarate:
So this doesn't use any experimental / private UIs?

brookewp:
> So this doesn't use any experimental / private UIs?

That's right! I initially considered using the [private Menu ](https://wordpress.github.io/gutenberg/?path=/docs/components-menu--docs)from `@wordpress/components,` which we could still do, and use the DropdownMenu added here as a fallback.

The only difference I found was in Menu, you can navigate to submenus using arrow keys. The submenus created from DropdownMenu can only be navigated with tab and enter. 

For now, using a private API doesn't seem worth the risk. And I believe this is an improvement as it's more accessible than our nested modals. 

If we want something more like Menu, we could use [Ariakit](https://ariakit.org/examples/menu-combobox), which Menu is built from. Then, we'd have more opportunities to use features like an integrated search in the menu.

inc/Editor/DataBinding/FieldShortcode.php
brookewp: >I'm not sure I follow, but I pushed up 59fbcb0 which should address the issue you were experiencing. Before that commit, the field shortcode working in the editor but failing on the frontend

Thank you! Also a good catch 😄

chriszarate (approved):
sweet

PR #354, opened 2025-02-07
load data source configs from constants

shekharnwagh:
# Description

Allows loading the data source configs from a global constant (`REMOTE_DATA_BLOCKS_CONFIGS`). This can be used by VIP Integrations to inject configurations. It follows the same schema as stored in the WP_OPTIONS table and like depending on the config it'll lead to blocks being automatically registered.

Considered different names for the new source of configs - `constant`, `injected`, `static` etc. but ended up with `constant` as its generic. The name `injected` also seemed appropriate given that we would be injecting the configs into the constant from the VIP Integrations code, but didn't go that route since we talked about keeping it generic and open to other uses. Open to changing it if we think another name would be more descriptive.

## Changelog

- Adds a field `config_source` to the data source config to indicate the source of the configuration. This field is not stored in the underlying source and is injected by the `DataSourceConfigManager` so the change is backwards compatible and should not break existing configured data sources.
- Adds a new class `DataSourceConfigManager` wraps CRUDS for different config sources (code, constants, storage) and delegates operations to the appropriate source.
- Adds a new class `ConstantConfigStore` that handles the constant-based configurations.
- Adds a new class `DataSourceAnalytics` to extract analytics code from the DataSourceController.
- Adds tests for the new classes.
- Adds a new `Constant` badge in the UI to indicate that the configuration was injected via the `REMOTE_DATA_BLOCKS_CONFIGS` constant. Like code configs they are not editable.
- Defaults `enable_block` in service config to false instead of defaulting to true when value is null/empty.

# How to Test

- Add a new config to the `REMOTE_DATA_BLOCKS_CONFIGS` constant. An example -

    ```php
    define( 'REMOTE_DATA_BLOCKS_CONFIGS', [
        [
            'service' => 'shopify',
            'uuid' => '8f47e5c2-9388-4f3b-a9d6-e1d4c2b3a7f9',
            'service_config' => [
                '__version' => 1,
                'access_token' => '<shopify-access-token>',
                'store_name' => '<store-name>',
                'display_name' => '<display-name>',
                'enable_blocks' => true,
            ],
        ],
    ] );
    ```

- Verify that the config is loaded correctly, is listed in the data source list and that the blocks are registered.


inc/Config/DataSource/DataSourceConfigManager.php
```diff
@@ -0,0 +1,233 @@
...
+
+class DataSourceConfigManager {
+	public const CONFIG_SOURCE_CODE = 'code';
+	public const CONFIG_SOURCE_STORAGE = 'storage';
+	public const CONFIG_SOURCE_CONSTANT = 'constant';
+	public const MUTABLE_CONFIG_SOURCES = [ self::CONFIG_SOURCE_STORAGE ];
+
+	private static function get_all_from_storage(): array {
+		return array_map(
+			fn ( array $config ) => array_merge( $config, [ 'config_source' => self::CONFIG_SOURCE_STORAGE ] ),
```
chriszarate: Do you think we should be consistent and use either all arrow functions or all anonymous function closures? The behavior of scope is different between the two

inc/Analytics/DataSourceAnalytics.php
```diff
@@ -0,0 +1,69 @@
...
+
+use RemoteDataBlocks\Analytics\TracksAnalytics;
+use RemoteDataBlocks\Config\DataSource\DataSourceConfigManager;
+
+defined( 'ABSPATH' ) || exit();
+
+class DataSourceAnalytics {
+	const DATA_SOURCE_INTERACTION_EVENT_NAME = 'remotedatablocks_data_source_interaction';
+	const DATA_SOURCE_VIEW_EVENT_NAME = 'remotedatablocks_view_data_sources';
+	const DATA_SOURCE_VIEW_TRACK_TRANSIENT_KEY = 'remotedatablocks_view_data_sources_tracked';
```
chriszarate: We should be specifying a prefix in the `Tracks` constructor instead of repeating it each time

inc/Analytics/DataSourceAnalytics.php
```diff
@@ -0,0 +1,69 @@
+<?php declare(strict_types = 1);
+
+namespace RemoteDataBlocks\Analytics;
```
chriszarate: Can we please name this class and folder using `Telemetry` instead of analytics?

inc/Analytics/DataSourceAnalytics.php
```diff
@@ -0,0 +1,69 @@
...
+		self::track_interaction( $config, 'update' );
+	}
+
+	public static function track_delete( array $config ): void {
+		self::track_interaction( $config, 'delete' );
+	}
+
+	public static function track_view( array $configs ): void {
+		/**
+		 * Tracks Analytics. Only once per day to reduce noise.
```
chriszarate: Why is it noisy and why wouldn't we want an accurate count? Can we call this when rendering the screen instead of from the REST endpoint?

inc/Config/DataSource/ConstantConfigStore.php
```diff
@@ -0,0 +1,95 @@
...
+
+namespace RemoteDataBlocks\Config\DataSource;
+
+use RemoteDataBlocks\Config\DataSource\DataSourceInterface;
+use RemoteDataBlocks\Logging\LoggerManager;
+use WP_Error;
+
+use const RemoteDataBlocks\REMOTE_DATA_BLOCKS__DATA_SOURCE_CLASSMAP;
+
+class ConstantConfigStore {
```
chriszarate: I think it's important for `inc/Config` to stayed focused on interfaces and query primitives for users to explore when extending. Can we move this to `include/Integrations` or another location?

inc/Config/DataSource/ConstantConfigStore.php
```diff
@@ -0,0 +1,95 @@
...
+		return $data_source_class::from_array( $config );
+	}
+
+	/**
+	 * Check if constant configuration is available and valid.
+	 *
+	 * @return bool Whether the constant configuration is available and valid.
+	 */
+	public static function is_available(): bool {
+		if ( ! defined( self::CONFIG_CONSTANT_NAME ) || '' === constant( self::CONFIG_CONSTANT_NAME ) ) {
```
chriszarate: covered by `is_array` below
```suggestion
		if ( ! defined( self::CONFIG_CONSTANT_NAME ) ) {
```

inc/Config/DataSource/ConstantConfigStore.php
```diff
@@ -0,0 +1,95 @@
...
+		$validation_result = self::validate_config( $config );
+		
+		if ( is_wp_error( $validation_result ) ) {
+			return $validation_result;
+		}
+
+		return $config;
+	}
+
+	public static function validate_config( array $config ): DataSourceInterface|WP_Error {
```
chriszarate: Can you update methods to `private` where not needed externally? Tests should target true public methods

inc/Config/DataSource/ConstantConfigStore.php
```diff
@@ -0,0 +1,95 @@
...
+		$data_source_class = REMOTE_DATA_BLOCKS__DATA_SOURCE_CLASSMAP[ $config['service'] ] ?? null;
```
chriszarate: Instead of using the classmap, we should use the new `__class` property from https://github.com/Automattic/remote-data-blocks/pull/370

inc/Config/DataSource/DataSourceConfigManager.php
```diff
@@ -0,0 +1,233 @@
...
+use RemoteDataBlocks\Editor\BlockManagement\ConfigStore;
+use RemoteDataBlocks\WpdbStorage\DataSourceCrud;
+use RemoteDataBlocks\Config\DataSource\ConstantConfigStore;
+use WP_Error;
+
+class DataSourceConfigManager {
+	public const CONFIG_SOURCE_CODE = 'code';
+	public const CONFIG_SOURCE_STORAGE = 'storage';
+	public const CONFIG_SOURCE_CONSTANT = 'constant';
+	public const MUTABLE_CONFIG_SOURCES = [ self::CONFIG_SOURCE_STORAGE ];
```
chriszarate: nice

inc/Config/DataSource/DataSourceConfigManager.php
```diff
@@ -0,0 +1,233 @@
...
+	 * @param string $uuid The UUID of the data source to delete.
+	 * @return true|WP_Error True on success, WP_Error on failure.
+	 */
+	public static function delete( string $uuid ): bool|WP_Error {
+		return DataSourceCrud::delete_config_by_uuid( $uuid );
+	}
+
+	/**
+	 * Get all configured data sources for a specific service from both storage and constants.
+	 * This includes sources configured via storage and constants, but not those defined in code.
```
chriszarate: Silently excluding these configs from a method named `get_all_...` feels wrong. Shouldn't the configs declare how they want to be treated? E.g., if they want queries and blocks registered for them?

chriszarate: What if a code-configured data source wants to have queries registered for it?

inc/REST/DataSourceController.php
```diff
@@ -164,40 +160,9 @@ public function create_item( mixed $request ): WP_REST_Response|WP_Error {
...
-			$ui_configured_data_sources_count = count( $ui_configured_data_sources );
-
-			TracksAnalytics::record_event( 'remotedatablocks_view_data_sources', [
-				'total_data_sources_count' => $code_configured_data_sources_count + $ui_configured_data_sources_count,
-				'code_configured_data_sources_count' => $code_configured_data_sources_count,
-				'ui_configured_data_sources_count' => $ui_configured_data_sources_count,
-			] );
-			set_transient( $track_transient_key, true, DAY_IN_SECONDS );
-		}
+		$data_sources = DataSourceConfigManager::get_all();
```
chriszarate, line 163: So nice and clean now, excellent work!

src/data-sources/DataSourceMetaTags.tsx
```diff
@@ -61,10 +62,19 @@ const CodeBadge = () => {
 	);
 };
 
+const ConstantsBadge = () => {
+	return (
+		<Tooltip text={ __( 'This data source is configured in constants.', 'remote-data-blocks' ) }>
```
chriszarate, line 67: Do we want to explain that it is configured by the platform? Some input from PM would be useful here.

maxschmeling:
When we implement this on the Seamless Integrations side, can all the integrations be put into a single constant? Wondering if that's going to be an issue if the Seamless Integrations code is walled off between integrations.

inc/Config/DataSource/ConstantConfigStore.php
shekharnwagh: 6784e56

inc/Config/DataSource/ConstantConfigStore.php
shekharnwagh: 278f46d

inc/Config/DataSource/ConstantConfigStore.php
shekharnwagh: Created a new dir `inc/Store/DataSource` and moved it there.

63e9183

inc/Analytics/DataSourceAnalytics.php
shekharnwagh: The logic was already present, I just re-organized it to a separate file. I am not aware of the reasoning behind this. I can't think of a reason why we should dedup this, so removed it and moved the tracking to when data fetched in UI.

44c869b

inc/Config/DataSource/DataSourceConfigManager.php
shekharnwagh: Converted them to use anonymous functions to align with existing usage of them.

90e85e2

inc/Analytics/DataSourceAnalytics.php
shekharnwagh: 3b63e94

inc/Analytics/DataSourceAnalytics.php
shekharnwagh: 9d83693

inc/Config/DataSource/DataSourceConfigManager.php
shekharnwagh: Code configured sources already have queries and blocks for them via the `register_remote_data_block` method. Would there be a use case where we only add the config directly [`ConfigStore`](https://github.com/Automattic/remote-data-blocks/blob/b842c6fad317fa974c3df94625b19f0d8082477b/inc/Editor/BlockManagement/ConfigStore.php) ?

I've anyway removed this method in favour of adding filters to the `get_all` method in 59c97a4

src/data-sources/DataSourceMetaTags.tsx
shekharnwagh: @smithjw1

inc/Store/DataSource/DataSourceConfigManager.php
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	/**
+	 * Get all data sources from all origins with optional filters.
+	 * 
+	 * Supported filters:
+	 * - service: Filter by service name (e.g. 'airtable', 'google-sheets', 'shopify')
+	 * - enable_blocks: Filter by blocks enabled status (false matches with null/false and true matches with true)
+	 * 
+	 * Any unsupported filter keys will be ignored.
```
chriszarate: I'd prefer to return an error

inc/Store/DataSource/DataSourceConfigManager.php
```diff
@@ -0,0 +1,248 @@
...
+	 * Apply filters to an array of configs.
+	 *
+	 * @param array $configs The configs to filter.
+	 * @param array{
+	 *   service?: string,
+	 *   enable_blocks?: bool
+	 * } $filters The filters to apply.
+	 * @return array The filtered configs.
+	 */
+	private static function apply_filters( array $configs, array $filters ): array {
```
chriszarate: This is a kind of confusing name, since `apply_filters` is a core WP function. Maybe `apply_config_array_filters`?

chriszarate: (I am often searching codebases for core functions like this, so less false positives is useful)

inc/Store/DataSource/DataSourceConfigManager.php
```diff
@@ -0,0 +1,248 @@
...
+			foreach ( $filters as $key => $value ) {
+				$passes_filter = match ( $key ) {
+					'service' => $config['service'] === $value,
+					'enable_blocks' => ( $config['service_config']['enable_blocks'] ?? false ) === $value,
+					default => true,
+				};
+
+				if ( ! $passes_filter ) {
+					return false;
+				}
```
chriszarate: ~I don't see how`$passes_filter` can ever be `false` given the final default expression. Maybe you want the default expression to resolve to `null` or some constant that represents an empty set?~

Also, can you add type hints (param and return) to the `match`? They are new to me.

inc/Store/DataSource/DataSourceConfigManager.php
```diff
@@ -0,0 +1,248 @@
...
+			}
+
+			return true;
```
chriszarate: It seems like this function always returns true. What am I missing?

inc/Store/DataSource/DataSourceConfigManager.php
```diff
@@ -0,0 +1,248 @@
...
+	 *   service: string,
+	 *   service_config: array<string, mixed>,
+	 *   config_source: string,
+	 *   __metadata?: array{
+	 *     created_at: string,
+	 *     updated_at: string
+	 *   }
+	 * }>
+	 */
+	public static function get_all( array $filters = [] ): array {
```
chriszarate: Can you write some tests for this, especially test cases that assert that an item not matched by the filter?

src/data-sources/DataSourceList.tsx
```diff
@@ -67,6 +74,28 @@ const DataSourceList = () => {
 		pushState( newUrl );
 	};
 
+	useEffect( () => {
+		/**
+		 * Track view when transitioning from loading to loaded state.
+		 */
+		if ( wasLoading.current && ! loadingDataSources ) {
```
chriszarate: I'm confused why the `ref` is needed. Can't you just early return if `loadingDataSources === true`?

src/data-sources/google-sheets/GoogleSheetsSettings.tsx
```diff
@@ -274,7 +275,7 @@ export const GoogleSheetsSettings = ( {
 			</DataSourceForm.Scope>
 			<DataSourceForm.Blocks
 				handleOnChange={ handleOnChange }
-				hasEnabledBlocks={ state.enable_blocks ?? true }
+				hasEnabledBlocks={ Boolean( state.enable_blocks ) }
```
chriszarate, line 278: Nice catch, these boolean bugs are touch to spot!

chriszarate (changes requested):
Thanks for all the improvements! 💪 Very close, just a couple questions / suggestions

inc/Store/DataSource/DataSourceConfigManager.php
chriszarate: Ok, I reread the `match` docs and I get it now. I was missing that the argument passed to `match()` is compared against the left-side expression. If it matches, the right-side expression is returned. 

I was confusing each entry in the `match` block as an arrow function. 

`match()` is kinda confusing! Lots going on.

inc/Store/DataSource/DataSourceConfigManager.php
chriszarate: Ignore the first part, but still curious about types.

src/data-sources/DataSourceList.tsx
shekharnwagh: The initial value for loading is `false` and it becomes `true` when it starts fetch. Without this ref, this would log the track event twice - once when the loading is `false` before making the fetch and once when it starts the actual fetch. This avoids it as we check the transition for loading from `true` to `false` which should only happen once.

inc/Store/DataSource/DataSourceConfigManager.php
shekharnwagh: We do have tests for scenario when empty configs are returned due to filters not matching -
- `testGetAllReturnsEmptyWhenServiceDoesNotMatch`
- `testGetAllReturnsEmptyWhenEnableBlocksDoesNotMatch`
- `testGetAllReturnsEmptyWhenMultipleFiltersDoNotMatch`

inc/Store/DataSource/DataSourceConfigManager.php
shekharnwagh: Does an early false return above when one of the filters fail.

inc/Store/DataSource/DataSourceConfigManager.php
shekharnwagh: 61d87d3

shekharnwagh:
> When we implement this on the Seamless Integrations side, can all the integrations be put into a single constant? Wondering if that's going to be an issue if the Seamless Integrations code is walled off between integrations.

@maxschmeling I was thinking of delegating the work of constructing this single constant from multiple integrations configs in GOOP to the Remote Data Blocks specific code in MU Plugins. An example of similar code for enterprise-search integration code in MU Plugins -

https://github.com/Automattic/vip-go-mu-plugins/blob/2ad65aa0076b4c1b4896ff3344259fb4b72699eb/integrations/enterprise-search.php#L67-L73

src/data-sources/DataSourceList.tsx
chriszarate: Ah. This is because we use an empty array as the default state for `dataSources` instead of a nullish value:

https://github.com/Automattic/remote-data-blocks/blob/b842c6fad317fa974c3df94625b19f0d8082477b/src/data-sources/hooks/useDataSources.ts#L15

If it was nullish, we could early return `if ( ! dataSources || loadingDataSources )`. I might argue that's a better pattern for loading remote data.

I think I asked this elsewhere, but why can't / don't we just fire this event in PHP when the screen is loaded?

shekharnwagh: I think it was here - https://github.com/Automattic/remote-data-blocks/pull/354#discussion_r1956307688

I assumed you meant to track them when the data sources list component is rendered, so added it here. 


> why can't / don't we just fire this event in PHP when the screen is loaded?

Where would that be ? Sorry, I haven't worked enough with WP to understand what/where you meant by this. Is that where we enqueue the JS for settings page ?

chriszarate: Here, where we output the markup for that page:

https://github.com/Automattic/remote-data-blocks/blob/b842c6fad317fa974c3df94625b19f0d8082477b/inc/PluginSettings/PluginSettings.php#L33

If this is simply a "view page" event, that's how I would implement it.

inc/Store/DataSource/DataSourceConfigManager.php
ingeniumed: Looks like the comment needs to be updated as well, with the latest changes?

inc/Store/DataSource/DataSourceConfigManager.php
```diff
@@ -0,0 +1,264 @@
...
+	 */
+	public static function get_all( array $filters = [] ): array|WP_Error {
+		$code_configured = self::get_all_from_code();
+		$constant_configured = self::get_all_from_constant();
+		$storage_configured = self::get_all_from_storage();
+
+		/**
+		 * De-duplicate configs.
+		 * 
+		 * Precedence (lowest to highest):
```
ingeniumed, line 93: Is there an existing doc within the repo that this could be added to? If not, could one be made? This seems like an inner working of the plugin that should be easily searchable for the future. I'm thinking something like Architecture Decisions maybe?

inc/Telemetry/DataSourceTelemetry.php
```diff
@@ -0,0 +1,42 @@
...
+
+defined( 'ABSPATH' ) || exit();
+
+class DataSourceTelemetry {
+	const DATA_SOURCE_INTERACTION_EVENT_NAME = 'data_source_interaction';
+
+	private static function get_interaction_track_props( array $config ): array {
+		$props = [];
+
+		if ( 'generic-http' === $config['service'] ) {
```
ingeniumed: Could `REMOTE_DATA_BLOCKS_GENERIC_HTTP_SERVICE` be used instead of `generic-http`? Would it be worth minimizing the hardcoded strings like the above. I'm not sure if it makes sense to move the rest of them to the constants file, but this one would imo.

src/blocks/remote-data-container/components/popovers/InputPopover.tsx
```diff
@@ -39,7 +39,7 @@ export function InputPopover( props: InputPopoverProps ) {
 	function onSelectItem(): void {
 		onSelect( inputState );
 		close();
-		sendTracksEvent( 'remotedatablocks_add_block', {
+		sendTracksEvent( 'add_block', {
```
ingeniumed, line 42: In case we have an internal list of these events, we should update that with these new names. Guessing @smithjw1 might have something like that?

src/data-sources/DataSourceList.tsx
ingeniumed: Just to add to this, this would be easier as the plugin settings page is under our control on the PHP side.

ingeniumed (commented):
Minor comments from my end, love how this simplifies a lot of Data source loading 👍🏾

src/data-sources/DataSourceList.tsx
shekharnwagh: The current tracks event also records the number of data sources fetched, including where they were configured from the `total_data_sources_count`, `code_configured_data_sources_count`, `ui_configured_data_sources_count` and `constants_configured_data_sources_count`. Would we be fine with losing that data in the view event ?

chriszarate: I don't know the answer to that. I tend to think it would be fine. But why couldn't that be added to the server-side event?

shekharnwagh: > why couldn't that be added to the server-side event?

The data is fetched on client side in `DataSourceList.tsx` so we won't have access to the data fetched in `PluginSettings.php`. Is there a way to pass it there using wp filters ?

chriszarate: > The data is fetched on client side in DataSourceList.tsx

That client-side fetch uses the REST API, which eventually resolves to server-side PHP code: `DataSourceController` and then to `DataSourceConfigManager`, which can be called in `PluginSettings`

chriszarate (approved):
Some minor remaining comments, but looks great overall

src/data-sources/DataSourceList.tsx
shekharnwagh: f5bb800

inc/Telemetry/DataSourceTelemetry.php
shekharnwagh: 15e16b9ba7286925fdb56fc24559ac41ad9fc6c2

PR #366, opened 2025-02-11
Add a migration feature

ingeniumed:
### Description

This PR is meant introduce the concept of migration with the plugin. This will ensure that we are able to migrate the schema of a config over time, and aren't stuck in a limbo when it changes. This has been done in a way that allows the implementor to stay in control while also being powerful enough for us to use it as well.

Ex: A future PR is going to move the `__version` from the service config level to the root config level. That would mean adding in the following within HttpDataSource#migrate_config:

```php
	public static function migrate_config( array $config ): array|WP_Error {
		// Migration: Move __version from service_config to the root level.
		if ( isset( $config['service_config']['__version'] ) ) {
			$config['__version'] = $config['service_config']['__version'];
			unset( $config['service_config']['__version'] );
		}

		return $config;
	}
```

### Testing

- Everything should work as is when it comes to the plugin itself because nothing has been migrated or changed.
- Run the test that's been added.
- Add in a sample migration for any data source, make the schema changes necessary to expose it and ensure it's given back. There's some changes necessary to `getAll` and `getById` to get them to show it. It'd be better to use one of the code level data sources.


chriszarate:
Summary of quick chat with @ingeniumed 

- Migrations should be implemented for every class that implements `ArraySerializableInterface`.
- Migrations can run in `from_array` before validation.
- The result of migrations don't need to be persisted. They are pure functions, so they can just run every time the config is loaded (until such time as they are updated by the user).
- `__version` should probably move to the top-level of config, and not be restricted to "service config". Unfortunately, this probably means updating the front-end code that manages data sources.

ingeniumed:
Thanks @chriszarate for the pairing session, working on the feedback now. I'll comment back here on it's ready to go as right now it's broken.

inc/Config/DataSource/HttpDataSource.php
```diff
@@ -44,6 +44,8 @@ final public function get_service_name(): string {
 	 * define their own validation schema.
 	 */
 	public static function preprocess_config( array $config ): array|WP_Error {
+		$config = static::perform_migration( $config );
```
ingeniumed: @chriszarate - the new `preprocess_config` method seems like the right place to call the migration method from within `HttpDataSource` looks like. Though if a custom implementor is used then it wouldn't be called via `from_array` so that is a downside.

What do you think about the way the migration methods are called here?

What I haven't done yet:

- Move the version field up one level
- Update the server and UI to account for that
- Added in a sample migration. The active/version changes would really be examples of those
- Modify the schema of a data source, and see if the migration works or not properly.

inc/Config/DataSource/HttpDataSource.php
```diff
@@ -63,6 +65,31 @@ public static function preprocess_config( array $config ): array|WP_Error {
...
+		if ( static::SERVICE_SCHEMA_VERSION === $config['service_config']['__version'] ) {
+			return $config;
+		}
+
+		$migrated_config = static::migrate_config( $config );
+
+		if ( is_wp_error( $migrated_config ) ) {
+			$config['active'] = false;
+		} else {
+			$config['service_config']['__version'] = static::SERVICE_SCHEMA_VERSION;
```
chriszarate: This should probably be delegated to the `migrate_config` version. Maybe there are scenarios where they don't want to advance to the latest version because there is missing info. I think we want to completely delegate to `migrate_config` and remove this method entirely.

inc/Config/DataSource/HttpDataSource.php
```diff
@@ -105,4 +132,12 @@ protected static function map_service_config( array $service_config ): array {
 			'request_headers' => $service_config['request_headers'] ?? [],
 		];
 	}
+
+	/**
+	 * Migrates the config to the current schema version.
+	 * Can be overridden by child classes to perform custom migrations.
+	 */
+	protected static function migrate_config( array $config ): array {
```
chriszarate:
```suggestion
	protected static function migrate_config( array $config ): array|WP_Error {
```

inc/Config/DataSource/HttpDataSource.php
chriszarate:
```suggestion
		$config = static::migrate_config( $config );
		if ( is_wp_error( $config ) ) {
			return $config;
		}
```

chriszarate: > Though if a custom implementor is used then it wouldn't be called via from_array so that is a downside.

`from_array` is now marked as `final` so that shouldn't be an issue

chriszarate: In general, I think we should delegate as much as possible and follow the pattern we've used elsewhere of "return the thing or a `WP_Error`, which we then elevate to the user to remedy"

inc/Config/DataSource/HttpDataSource.php
ingeniumed: That's a fair point. Having us control would negate the point of letting data source owners take charge of the migration process.

inc/Config/DataSource/HttpDataSource.php
ingeniumed: My intention was to allow us to be able to migrate the fields that we control within the config as well.

A good example of this would be the first migration:

- Adding in active field to indicate if a data source is active or not. This could be useful for cases where a migration was unsuccessful or if a data source has errors and we need it to be fixed. This would help build that UI feature.
- Moving the version field to the top of the config rather than within the service config.

There could be more down the line but that's why it's been done that way.

Are you saying to move what I have in `perform_migration` into `migrate_config` instead and allow for the data from the overriden methods to be merged with that? 

Hope that makes sense, and I correctly articulated what you are suggesting.

chriszarate: > Adding in active field to indicate if a data source is active or not.

Can't we just use `WP_Error` as a signal for that? The migration either works or it doesn't. And if it doesn't we have no way to inspect the problem other than to surface an error. In that case we can also treat it as inactive.

chriszarate: > Moving the version field to the top of the config rather than within the service config.

This is a design mistake that I think we should just fix. Doesn't need to be handled by a migration, IMO

chriszarate: Put another way, if the migration fails and you set active = false, how are you going to read and respect that? The config still needs to pass validation (which it likely won't, so it will get elevated to an error anyway).

inc/Config/DataSource/HttpDataSource.php
```diff
@@ -37,6 +37,10 @@ final public function get_service_name(): string {
 		return static::SERVICE_NAME;
 	}
 
+	final public function get_config_schema_version(): int {
```
chriszarate: Doesn't seem like this is needed, I don't see it called anywhere

ingeniumed: I haven't incorporated it in, that's why. Once I finish off this PR, I'll see if its still needed

ingeniumed:
I have reverted my attempt at moving the `__version` field to the root level. This is because there are pathways called via the DataSourceController that don't inflate the config using our various DataSource implementations and so migration isn't called on them. They'll need to be modified, and as a result the files changes exceeds 40. That's way too much given the scope of this PR. 

I'll make a follow up PR for that. This PR will be focused on introducing migration and adding tests for it.

tests/inc/Config/HttpDataSourceTest.php
```diff
@@ -4,11 +4,11 @@
 
 use PHPUnit\Framework\TestCase;
 use RemoteDataBlocks\Tests\Mocks\MockDataSource;
-
+use WP_Error;
 class HttpDataSourceTest extends TestCase {
-	private MockDataSource $http_data_source;
+	private MockDataSource|WP_Error $http_data_source;
```
ingeniumed, line 10: This was necessary so that the error pathway for migrations could be tested.

@chriszarate  - could this become method scoped, or is there a plan down the line to add more tests that would benefit from this being class scoped?

tests/inc/Config/HttpDataSourceTest.php
```diff
@@ -4,11 +4,11 @@
...
 
-	public function testGetServiceMethodCannotBeOverriddenl(): void {
+	public function testGetServiceMethodCannotBeOverridden(): void {
```
ingeniumed, line 12: Noticed the name had a typo but I hope that wasn't on purpose

tests/inc/Mocks/MockDataSource.php
```diff
@@ -30,4 +30,21 @@ public static function create( ?array $config = self::MOCK_CONFIG, ?ValidatorInt
 	public function set_endpoint( string $endpoint ): void {
 		$this->config['endpoint'] = $endpoint;
 	}
+
+	/**
+	 * Override the migrate_config method to adjust the config for testing.
+	 */
+	public static function migrate_config( array $config ): array|WP_Error {
```
ingeniumed, line 37: Added in this `migrate_config` implementation that helps to setup a successful and unsuccessful migration.

The idea is that all the other uses of this mock data source won't be impacted as they won't have this test field set or won't care about it.

tests/inc/Config/HttpDataSourceTest.php
```diff
@@ -26,4 +26,44 @@ public function testGetServiceMethodReturnsCorrectValue(): void {
 
 		$this->assertEquals( 'generic-http', $this->http_data_source->get_service_name() );
 	}
+
+	public function testMigrateConfigMethodCanBeOverridden_user_id_is_added_to_config(): void {
```
ingeniumed, line 31: These three methods test out the `migrate_config` implementation in the MockDataSource. Hopefully there's enough comments here to explain it all

inc/Config/ArraySerializable.php
```diff
@@ -29,11 +29,20 @@ protected function get_or_call_from_config( string $property_name, mixed ...$cal
 	 * @inheritDoc
 	 */
 	final public static function from_array( array $config, ?ValidatorInterface $validator = null ): static|WP_Error {
+		// The purpose of this is to ensure that when from_array runs, it is statically bound to the correct child class.
```
ingeniumed, line 32: @chriszarate - I hope I have captured the purpose behind this logic correctly.

brookewp (approved):
I tested this out with the art institute example and it worked as expected ✅

tests/inc/Config/HttpDataSourceTest.php
chriszarate: Not sure what you are referring to when you say method- or class-scoped.

chriszarate: This `$http_data_source` property isn't very useful if each test is creating their own instance, it can just be removed and the tests updated to use a local variable.

tests/inc/Mocks/MockDataSource.php
chriszarate: Seems fine, in the future we might want to give control to the test, similar to `MockQuery#set_response_data`

chriszarate (approved):
double stamp

tests/inc/Config/HttpDataSourceTest.php
ingeniumed: I hit that merge way too quickly. I'm going to make two follow up PRs in which I'll do that:

- Add some docs on how to do migration
- Add a migration for the `__version` property

PR #367, opened 2025-02-12
Update schema to allow nested serialized config

chriszarate:
This PR updates the core schema to allow nested serialized config. We can now represent remote data config as a single serialized array or object, which would allow loading config via JSON or any other serialization format:

```php
register_remote_data_block( [
  'title' => 'My block',
  'render_query' => [
    'query' => [
      'data_source' => [
        'service_config' => [
          '__version' => 1,
          'display_name' => 'My Data Source',
          'endpoint' => 'https://example.com/api',
        ],
      ],
      'display_name' => 'Mock Query',
      'input_schema' => [ /* ... */ ],
      'output_schema' => [ /* ... */ ],
    ],
  ],
] );
```

You can target any class that extends `HttpQuery` or `HttpDataSource`—for example, our integrations like `ShopifyDataSource` by referencing your desired class names:

```php
register_remote_data_block( [
  'title' => 'My block',
  'render_query' => [
    'query' => [
      '__subclass' => 'RemoteDataBlocks\\Integrations\\Shopify\\GetProductQuery',
      'data_source' => [
        '__subclass' => 'RemoteDataBlocks\\Integrations\\Shopify\\ShopifyDataSource',
        'service_config' => [
          '__version' => 1,
          'access_token' => 'token',
          'store_name' => 'My Shopify Store',
          /* etc */
        ],
      ],
    ],
  ],
] );
```

IMO, this is a much more durable target for "generated" code and is less likely to break with minor code updates. It can also participate in "config migrations" that Gopal is working on.

The changes in this PR are relatively light and mainly deal with updated interfaces, type annotations, validation, and test cases. The main addition is a new type: `Types::serialized_config_for()`, which allows you to provide config for a class that implements `ArraySerializableInterface`. This config will be validated and used to inflate the class instance on startup.

### DIFF
--- a/inc/Config/ArraySerializable.php
+++ b/inc/Config/ArraySerializable.php
@@ -28,7 +28,17 @@ abstract class ArraySerializable implements ArraySerializableInterface {
 	/**
 	 * @inheritDoc
 	 */
-	public static function from_array( array $config, ?ValidatorInterface $validator = null ): self|WP_Error {
+	public static function from_array( array $config, ?ValidatorInterface $validator = null ): static|WP_Error {
+		$subclass = static::get_subclass( $config );
+		if ( null !== $subclass ) {
+			return $subclass::from_array( $config, $validator );
+		}
+
+		$config = static::preprocess_config( $config );
+		if ( is_wp_error( $config ) ) {
+			return $config;
+		}
+
 		$schema = static::get_config_schema();
 
 		$validator = $validator ?? new Validator( $schema, static::class );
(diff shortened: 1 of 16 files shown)


inc/Config/ArraySerializable.php
```diff
@@ -51,5 +61,29 @@ public function to_array(): array {
 		return $this->config;
 	}
 
-	abstract protected static function get_config_schema(): array;
+	/**
+	 * @inheritDoc
+	 */
+	public static function preprocess_config( array $config ): array|WP_Error {
+		return $config;
```
alecgeatches, line 68: Is this a placeholder for migration code?

alecgeatches: Oh I see [from the lines below](https://github.com/Automattic/remote-data-blocks/blob/2e2b2af00d4a93d5b48b18ce4bc6a875e4927017/inc/Config/ArraySerializableInterface.php#L22-L30) this is meant to be used as an extendable method.

alecgeatches (approved):
Great idea, this is very cool!

inc/Config/ArraySerializable.php
chriszarate: Yes! But also the place to put / call migration code.

chriszarate:
Merging but happy to revisit naming decisions pre-release

PR #369, opened 2025-02-13
Implement bulk selection in editor

brookewp:
**Updated visual**: 

Screenshot 2025-02-21 at 2 37 11 PM

Adds optional `supports_bulk` to input variable which will add an option to bulk select data in the editor:

https://github.com/user-attachments/assets/ea3aeab8-c3d0-48cc-9db0-32656d42564d

To support this, the endpoint must allow for multiple items to be queried. I've updated the Airtable Integration and the Art Institute example to accept single or multiple IDs. 

Instead of returning an object with a single value, it'll return the object with comma separated strings. For the examples I've mentioned above, this includes using a comma separated string entered into the manual input (Thanks @alecgeatches for pointing this out)

The solution I've come up with is adding a total count, and hiding the DataViews button, and adding our own within the Modal instead, so we can control when it's shown, and make sure it doesn't affect blocks that don't support bulk select.


alecgeatches:
Howdy @brookewp! Just started testing, and I really like the UI for selection! I noticed a couple of things so far:

1. Items can only be multi-selected across the same page:

    

    When changing pages, the selection resets. I'm not sure if this is a data views limitation, but it seems likely users will need to select from items that appear across multiple pages.

2. Manual input on a multiselect still only asks for a single ID.
     
     Screenshot 2025-02-18 at 11 30 54 AM

    Maybe we add below the "ART ID" label to include "Comma-separate multiple values" and handle that? Just seems useful to cover this case, but not super important.

example/rest-api/art-institute/art-institute.php
```diff
@@ -31,7 +31,32 @@ function register_aic_block(): void {
...
 		'data_source' => $aic_data_source,
 		'endpoint' => function ( array $input_variables ) use ( $aic_data_source ): string {
-			return sprintf( '%s/%s', $aic_data_source->get_endpoint(), $input_variables['id'] ?? '' );
+			$endpoint = $aic_data_source->get_endpoint();
+			$ids = [];
+				
+			if ( isset( $input_variables['id'] ) ) {
+				$ids[] = $input_variables['id'];
+			} else {
+				foreach ( $input_variables as $input ) {
```
alecgeatches: This `foreach` seems a bit complicated, what case does this cover?

example/rest-api/art-institute/art-institute.php
```diff
@@ -40,20 +65,28 @@ function register_aic_block(): void {
 			],
 		],
 		'output_schema' => [
-			'is_collection' => false,
-			'path' => '$.data',
+			'is_collection' => true,
```
alecgeatches, line 59: Does `supports_bulk = true` imply `is_collection = true`?

alecgeatches: I actually should rephrase, I think that's the wrong terminology. Does a block having `supports_bulk = true` on the selection query imply that the get query needs to be `is_collection = true`?

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -199,6 +199,21 @@ public static function get_value( array $source_args, WP_Block|array $block ): ?
...
+		}
+
+		// If we have multiple results and this is not already a loop item (no index specified),
+		// switch to loop rendering
+		if ( count( $query_response['results'] ) > 1 && !isset( $source_args['index'] ) ) {
+			if ( $block instanceof WP_Block ) {
+				return self::loop_block_render_callback( $block_attributes, '', $block );
+			}
+			// For non-WP_Block instances, fallback to single result
+			$index = 0;
```
alecgeatches, line 215: @brookewp Can you help me understand why the `$block instanceof WP_Block` check and this code exists? When does the type of block coming in make a difference here?

alecgeatches: I see `loop_block_render_callback()` requires a `WP_Block`, but I'm not sure why a block comes in as a `WP_Block` versus a array, and what that means.

inc/Integrations/Airtable/AirtableIntegration.php
```diff
@@ -93,14 +94,47 @@ public static function get_query( AirtableDataSource $data_source, array $table
 		];
 
 		$output_schema = [
-			'is_collection' => false,
+			'is_collection' => true,
```
alecgeatches, line 94: The transformation here [and above](https://github.com/Automattic/remote-data-blocks/pull/369/files#r1960318391) makes me wonder if we should write all getter queries to assume `is_collection` to be true, and require them to implement multi-select gets.

alecgeatches: A couple of ideas:

1. It's possible that some API only can do GET requests one at a time. Maybe we could implement an example of an API that doesn't support multi-select, but makes multiple requests internally in the `get_query` call.
2. OR we instead have a flag that indicates a request only supports single queries `supports_multi = false`, so all queries are multi-select by default.

I think most APIs will support bulk selection, and it would be great if we could have this all supported and working out of the box without needing to add these extra flags.

inc/Integrations/Airtable/AirtableIntegration.php
```diff
@@ -93,14 +94,47 @@ public static function get_query( AirtableDataSource $data_source, array $table
...
-				return $data_source->get_endpoint() . '/' . $table['id'] . '/' . $input_variables['record_id'];
+				// Normalize input variables to ensure consistent structure
+				$record_ids = [];
+				
+				// Handle both direct record_id and array of record_ids
+				if ( isset( $input_variables['record_id'] ) ) {
+					// Single record case
+					$record_ids[] = $input_variables['record_id'];
+				} else {
+					// Multiple records or array case
```
alecgeatches: Maybe refactor this with the other use [in the art institute block](https://github.com/Automattic/remote-data-blocks/pull/369/files#r1960311595) into a reusable function? I'm still a bit confused on what's happening here, but it seems like an operation we need multiple places.

inc/REST/RemoteDataController.php
```diff
@@ -61,10 +61,6 @@ public static function execute_query( WP_REST_Request $request ): array|WP_Error
 		$block_config = ConfigStore::get_block_configuration( $block_name );
 		$query = $block_config['queries'][ $query_key ];
 
-		// The frontend might send more input variables than the query needs or
-		// expects, so only include those defined by the query.
-		$query_input = array_intersect_key( $query_input, $query->get_input_schema() );
```
alecgeatches: Curious why this no longer is necessary

alecgeatches (dismissed):
I left a comment above with a couple of potential UI changes, and then a bunch of questions below. Overall this is really cool, and I'm curious to see if we can change the configuration to get this "for free" for most queries, rather than having it be an add-on.

example/rest-api/art-institute/art-institute.php
brookewp: Ya 😬 This is to ensure that a single item can still be selected and transforms it into an array to handle either a single item or multi-item selection.

In the code, it first checks if an 'id' key and if it's there, it creates a single-item array. Otherwise, it filters the array of input variables for items with 'id' and maps them to extract the id values. 

I added comments and tried to make the code easier to read, but not sure if I've achieved it. If you have any suggetions, please share! 

https://github.com/Automattic/remote-data-blocks/blob/8ca83120bce3d91fc4238816d2464fd96bba4ada/example/rest-api/art-institute/art-institute.php#L35-L43

example/rest-api/art-institute/art-institute.php
brookewp: >Does a block having supports_bulk = true on the selection query imply that the get query needs to be is_collection = true?

The answer is yes, but I'm not sure if it should be. So great question! 😄 

To use our existing setup, it needs to be `true` so we can handle multiple returned results, otherwise, we only expect a single result to be returned. 

But now I'm curious why we have`is_collection` in the first place. In trunk without multi-select, if I set the collection to `true` for this example, we can still display the single result without issue. 

To me, it seems safer to handle everything as an array so we can either map through one or more results and not need the `is_collection.`

inc/Editor/DataBinding/BlockBindings.php
brookewp: At the start of the function, I saw this helpful comment from Hew: 

https://github.com/Automattic/remote-data-blocks/blob/8ca83120bce3d91fc4238816d2464fd96bba4ada/inc/Editor/DataBinding/BlockBindings.php#L168-L177

From my understanding, when we are using standard WP methods for rendering blocks, we end up with a `WP_Block` instance. But if we are doing something custom (e.g., our hooks) before it's rendered in the editor or on the frontend, we end up with an array for more flexibility. But either way, it'll eventually become a `WP_Block` through the core process. So, I don't think it needs strict typing, I was just adhering to what was established initially in this function.

inc/Integrations/Airtable/AirtableIntegration.php
brookewp: >It's possible that some API only can do GET requests one at a time. Maybe we could implement an example of an API that doesn't support multi-select, but makes multiple requests internally in the get_query call.

You may have seen this already in [my earlier comment](https://github.com/Automattic/remote-data-blocks/pull/369#discussion_r1960570696) regarding `is_collection`, but I believe the boolean determines what is returned, so either a collection of results or a single result:

https://github.com/Automattic/remote-data-blocks/blob/b842c6fad317fa974c3df94625b19f0d8082477b/docs/extending/query.md?plain=1#L147

Expanding on my comment, I wonder why, instead of relying on the config/asking the user to tell us if it's a collection, we aren't checking for an array like this:

```diff
diff --git a/inc/Config/QueryRunner/QueryRunner.php b/inc/Config/QueryRunner/QueryRunner.php
index 436ef480..4f12998c 100644
--- a/inc/Config/QueryRunner/QueryRunner.php
+++ b/inc/Config/QueryRunner/QueryRunner.php
@@ -223,7 +223,7 @@ class QueryRunner implements QueryRunnerInterface {
 		// is_collection and unwrap if necessary.
 		$parser = new QueryResponseParser();
 		$results = $parser->parse( $response_data, $output_schema );
-		$results = $is_collection ? $results : [ $results ];
+		$results = is_array( $results ) ? $results : [ $results ];
 		$metadata = $this->get_response_metadata( $query, $raw_response_data['metadata'], $results );
 ```

Keeping the results to an array for consistency, but not limiting to the first item.

Comment (brookewp) [reply inc/Integrations/Airtable/AirtableIntegration.php:94]: As a side note to potential single API requests, I don't think we would want to connect multiple selections to multiple GET requests, as I think we'd run into rate limiting issues pretty quickly

File: inc/REST/RemoteDataController.php (hunk 6 above)
Comment (brookewp) [reply inc/REST/RemoteDataController.php]: Good catch! I initially removed it because it wasn't working with multi-selection, but I intended to circle back to it. I added it back here, with a check for an array to include multi-selection: [`a0ff2c9` (#369)](https://github.com/Automattic/remote-data-blocks/pull/369/commits/a0ff2c929142d12591e7e0cb9a216992702bcac7)

I'm not 100% sure what the benefit of it is, aside from making the request smaller. 🤔

**Comment by brookewp** [comment]
Thanks for all of the thoughtful comments @alecgeatches! 

>Items can only be multi-selected across the same page:
>When changing pages, the selection resets. I'm not sure if this is a data views limitation, but it seems likely users will need to select from items that appear across multiple pages.

Looks like a bug in DataViews 😞 

>Manual input on a multiselect still only asks for a single ID.
>Maybe we add below the "ART ID" label to include "Comma-separate multiple values" and handle that? Just seems useful to cover this case, but not super important.

To be honest, I didn't even think we'd include multi-select here 🤯 but it's a good point!

File: example/rest-api/art-institute/art-institute.php (hunk 2 above)
Comment (chriszarate) [reply example/rest-api/art-institute/art-institute.php:59]: > But now I'm curious why we have `is_collection` in the first place.

Short answer: To provide a consistent response shape from the `remote-data` REST endpoint. 

Longer answer involves the fact that JSONPath always returns an array, even for single values:

Screenshot 2025-02-19 at 10 02 20

We use `is_collection` when parsing response data to understand if the return value is a single item or, coincidentally a collection with only one member.

File: example/rest-api/art-institute/art-institute.php [hunk 7]
```diff
@@ -129,6 +157,7 @@ function register_aic_block(): void {
 			[
 				'query' => $search_art_query,
 				'type' => 'search',
+				'supports_bulk' => true,
```
chriszarate: I think it's extremely important for the queries themselves to declare their support for various UI features—not the config that surrounds the queries. We should give the input variable that accepts multiple IDs a special type, just as we have for search terms and pagination input variables.

Reasons:

1. It reduces the complexity of the block config, which is already complex.
2. We keep the configuration very close to the thing it is describing (the input variable), instead of far away.
3. We reduce the likelihood of having to repeat this configuration value in multiple places.
4. It matches the pattern we are already using.

inc/Editor/DataBinding/BlockBindings.php
chriszarate: That's correct, it's mainly to accommodate the Block Data API. It also happens to make testing easier.

inc/REST/RemoteDataController.php
chriszarate: > I'm not 100% sure what the benefit of it is, aside from making the request smaller.

Some APIs throw an error when they receive unexpected input. (Our own APIs do this.)

example/rest-api/art-institute/art-institute.php
chriszarate: After reading a bit more code, I think I'm unclear which approach you want to take:

a. Provide an array of _n_ IDs to a query and have it dispatch a single request for all of them. The query must support multiple IDs as input.
b. Given an array of _n_ IDs, execute the query _n_ times and use a "loop" block to render them.
c. Dynamically choose approach A or B, depending on whether the query supports multiple IDs as input.

Can you elaborate? If it's B or C, doesn't every query "support" bulk selection? Why would it even be an option?

brookewp: My approach was more toward option A, which allows you to enable bulk selection for the query (which would need to be updated to align with your earlier comment) if you have an API that accepts multiple inputs. Since you can submit a single input to these endpoints, I didn't think we'd need to have a specific input variable just for multiple selections. 

I also thought executing the query multiple times wouldn't be ideal, as someone may be blocked for making too many subsequent requests. 

But I'd like to hear your thoughts on which direction we should aim for. 🙂

example/rest-api/art-institute/art-institute.php
brookewp: Thanks for answering this! If the response always returns an array, even for single values, then couldn't we parse everything as an array and not have to rely on the config to know if it's a single item in an array or multiple items?

chriszarate: I think that might be possible if we merge the concept of single and loop blocks (which we probably should!). Right now, it just helps to prevent lots of inspection of the response to determine if its a collection or not.

example/rest-api/art-institute/art-institute.php
brookewp: @chriszarate I moved the flag to the input variable for the render/display queries in [`4df9b41` (#369)](https://github.com/Automattic/remote-data-blocks/pull/369/commits/4df9b410a5ae6693f15067b88d1d943c0d8e7516). Side notes: 
- I didn't make it a special type, because I thought we might still want to know the type, 'id' in this case (which I [use in ItemList](https://github.com/Automattic/remote-data-blocks/blob/bf0fc5cedf8423b80772e0d7da8e908e855778c5/src/blocks/remote-data-container/components/item-list/ItemList.tsx#L91-L94))
- We still need to communicate this capability to the selection queries, so I added a check in the config registry 
- I also limited bulk selection so only one input variable can be present, as I'm not certain how we want to factor

brookewp:
@alecgeatches I've implemented multi-select via comma separated IDs for the manual input. 

Also, it seems we can select across multiple pages by updating this to a controlled component. I haven't pushed those changes up yet, as I'm not sure how to deal with the incorrect number of selected items managed by DataViews. 🤔 The 'best' solution is probably adding the count into the header and hiding the existing one from DataViews. It just won't be the ideal placement.

brookewp:
For the bulk selection pagination issue, the problems: 
- navigating to other pages clears the selection
- when fixing this by using it as a controlled component, we still have issues where the buttons will disappear unless you go back to pages with selections or you make a new selection
- the total count at the bottom is per page, but I don't think it's entirely clear 

The solution I've come up with is adding a total count, and hiding the DataViews button, and adding our own within the Modal instead, so we can control when it's shown, and make sure it doesn't affect blocks that don't support bulk select. 

Screenshot 2025-02-21 at 1 25 46 PM

brookewp:
~https://github.com/user-attachments/assets/646e0105-41df-4a50-aa10-d3a0c6e37090~

~Maybe a 'Clear all' button could be helpful?~

Edited: doesn't apply with the new changes, with the custom buttons in the modal vs using the original from DataViews (where cancel would just clear the current page's selection)

alecgeatches (approved):
Tested again and this is working really well! I like that the multi-page selection is working and multi-selection by ID string. I'm excited to see if we can fully remove the idea of single/multiple/other queries into a combined set.

Thank you for your work here, looking great!

PR #370, opened 2025-02-13
Fix design mistake in nested config implementation

chriszarate:
This fixes a small design mistake in #367. The mistake is that it requires participants to extend `ArraySerializable` instead of to implement `ArraySerializableInterface`.

- As such, it renames the `__subclass` property to `__class` to reflect that subclass inheritance is not required.
- It also marks `ArraySerializable::from_array()` as `final` so that we can call it and have confidence that the `__class` property will be noticed and respected. If it is overridden, that promise could be broken.

### DIFF
--- a/inc/Validation/Validator.php
+++ b/inc/Validation/Validator.php
@@ -197,12 +197,14 @@ final class Validator implements ValidatorInterface {
 					return $this->create_error( 'Value does not provide a __class property', $class_ref );
 				}
 
+				$class_description = sprintf( 'Class %s specified by %s property', $subclass, ArraySerializableInterface::CLASS_REF_ATTRIBUTE );
+
 				if ( ! class_exists( $subclass ) ) {
-					return $this->create_error( 'Class specified by __class property does not exist', $subclass );
+					return $this->create_error( $class_description . ' does not exist', $subclass );
 				}
 
 				if ( $subclass !== $class_ref && ! is_subclass_of( $subclass, $class_ref, true ) ) {
-					return $this->create_error( 'Class specified by __class property must match or be a subclass of the target class', $subclass );
+					return $this->create_error( $class_description . ' must match or be a subclass of the target class', $subclass );
 				}
 
 				// Done with type validation, update the target class so we can validate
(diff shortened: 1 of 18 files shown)


inc/Validation/Validator.php
```diff
@@ -188,17 +188,27 @@ private function check_non_primitive_type( array $type, mixed $value ): bool|WP_
...
+					return $this->create_error( 'Value does not provide a __class property', $class_ref );
+				}
 
-					$class_ref = $subclass;
+				if ( ! class_exists( $subclass ) ) {
+					return $this->create_error( 'Class specified by __class property does not exist', $subclass );
 				}
 
+				if ( $subclass !== $class_ref && ! is_subclass_of( $subclass, $class_ref, true ) ) {
+					return $this->create_error( 'Class specified by __class property must match or be a subclass of the target class', $subclass );
```
alecgeatches: This would be a pretty abstract error to receive. Would it be possible to `sprintf()` what these classes should be? e.g.

```suggestion
					return $this->create_error( sprintf( 'Class specified by __class property "%s" must match or be a subclass of the target class "%s"', $subclass, $class_ref ), $subclass ) );
```

although I'm not sure if the string interpolation would work with these values, or if I've put them in the right order.

alecgeatches: Also, something similar for the error above, too:

```
Class specified by __class property "%s" does not exist
```

alecgeatches:
@chriszarate Could you add an integration test for a `__class`ed block? In #367 you added `testRegisterBlockWithNestedConfig()` which checks the registration, but it would be nice to see an example and proof-of-concept that uses results from nested configuration.

alecgeatches (approved):
Changes look good! Asked for a couple of error message improvements (if possible) and an integration test with nested configuration, but good to go otherwise.

chriszarate:
- Added integration test in https://github.com/Automattic/remote-data-blocks/pull/370/commits/6fbfdde966f4c54b0e7c3603a18f98b409af13ea
- Improved validation messages in https://github.com/Automattic/remote-data-blocks/pull/370/commits/1e273d718801f83c5664e664a664f73162700372

PR #372, opened 2025-02-14
Add Salesforce D2C as a data source

ingeniumed:
### Description

The purpose of this PR is to remove the old Salesforce B2C data source, and instead replace it with the Salesforce D2C data source. In addition, it also adds the ability to skip a search query in the event that the term itself is empty. I found that this was breaking the query search functionality and it felt like we shouldn't make the query if there is no search term to begin with.

I am cautiously putting this as a draft PR first to make sure the api calls are correct, first and foremost. Next, I'd also like to see if there are any other improvements I could make.

Search functionality

Screenshot 2025-02-14 at 4 00 44 pm

Full post with a product_id hardcoded, and the Energy search term being used as the query

Screenshot 2025-02-14 at 4 01 04 pm


inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -205,6 +205,18 @@ public function execute( HttpQueryInterface $query, array $input_variables ): ar
 			}
 		}
 
+		// If the query is a search query and the search term is empty, return a
+		// collection with no results. This is to exit early and avoid making an
+		// unnecessary API call.
+		if ( 'search' === $key && empty( $input_variables['search'] ) ) {
```
ingeniumed: Is there a built in way to do this already that I might have missed? I essentially want to skip the initial search call if the provided search term is empty. This happens when a block is first created and I don't think it should occur. Until a search term is provided, this should be skipped.

src/data-sources/salesforce-d2c/SalesforceD2CSettings.tsx
```diff
@@ -54,34 +54,36 @@ export const SalesforceB2CSettings = ( {
...
-				inputIcon={ SalesforceCommerceB2CIcon }
+				inputIcon={ SalesforceCommerceD2CIcon }
 				uuid={ uuid }
 			>
 				<TextControl
 					type="text"
-					label={ __( 'Merchant shortCode', 'remote-data-blocks' ) }
-					onChange={ shortCode => {
-						handleOnChange( 'shortcode', shortCode ?? '' );
+					label={ __( 'Instance URL', 'remote-data-blocks' ) }
```
ingeniumed: Is there a better way to get the instance url? Note that I haven't done any validation here to ensure it's `https` for example. I don't quite know the format of these urls, hence why I've given an example only.

inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -205,6 +206,18 @@ public function execute( HttpQueryInterface $query, array $input_variables ): ar
...
+		// unnecessary API call.
+		if ( ConfigRegistry::SEARCH_QUERY_KEY === $key && empty( $input_variables[ ConfigRegistry::SEARCH_QUERY_KEY ] ) ) {
+			return [
+				'is_collection' => true,
+				'metadata' => [],
+				'pagination' => [],
+				'results' => [],
+			];
+		}
+
```
chriszarate: Instead, the query should mark the input variable as `'required' => true` and the UI should be updated to implement this logic. Query runner should stay ignorant of implementation details like this.
```suggestion
```

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
```diff
@@ -0,0 +1,118 @@
...
+		string $client_secret
+	): string|WP_Error {
+		$saved_access_token = self::get_saved_access_token( $client_id );
+
+		if ( null !== $saved_access_token ) {
+			return $saved_access_token;
+		}
+
+		$access_token = self::get_token_using_client_credentials( $client_id, $client_secret, $endpoint );
+		return $access_token;
```
chriszarate: FYI we can use nullish coalescing for brevity
```suggestion
		return self::get_saved_access_token( $client_id ) ?? self::get_token_using_client_credentials( $client_id, $client_secret, $endpoint );
```

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	// Access token request using top-level credentials
```
chriszarate: proper docblock

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
```diff
@@ -0,0 +1,118 @@
...
+			);
+		}
+
+		$access_token = $response_data['access_token'];
+		self::save_access_token( $access_token, $client_id );
+
+		return $access_token;
+	}
+
+	// Access token cache management
```
chriszarate: proper docblock

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
```diff
@@ -0,0 +1,118 @@
...
+
+		$saved_access_token = wp_cache_get( $access_token_cache_key, 'oauth-tokens' );
+
+		if ( false === $saved_access_token ) {
+			return null;
+		}
+
+		$access_token = $saved_access_token['token'];
+
+		return $access_token;
```
chriszarate: for safety
```suggestion
		return $saved_access_token['token'] ?? null;
```

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
```diff
@@ -0,0 +1,118 @@
...
+	// Access token cache management
+
+	private static function save_access_token( string $access_token, string $client_id ): void {
+		$access_token_data = [
+			'token' => $access_token,
+		];
+
+		$access_token_cache_key = self::get_access_token_key( $client_id );
+
+		wp_cache_set(
```
chriszarate, line 175: Is this intentionally not setting a TTL? We had one before

src/data-sources/salesforce-d2c/SalesforceD2CSettings.tsx
chriszarate: > Is there a better way to get the instance url?

Do you mean in React / UI terms? Or in API terms? I don't understand the question, sorry.

src/data-sources/salesforce-d2c/SalesforceD2CSettings.tsx
```diff
@@ -2,44 +2,44 @@ import { TextControl } from '@wordpress/components';
...
 			__version: SERVICE_CONFIG_VERSION,
 			enable_blocks: true,
 		},
 	} );
 
 	const shouldAllowSubmit = useMemo( () => {
-		return state.shortcode && state.organization_id && state.client_id && state.client_secret;
-	}, [ state.shortcode, state.organization_id, state.client_id, state.client_secret ] );
+		return state.instance_url && state.store_id && state.client_id && state.client_secret;
+	}, [ state.instance_url, state.store_id, state.client_id, state.client_secret ] );
```
chriszarate: Useless memo, there is no reason to memoize a simple boolean check. Just compute it. (I know this is existing code, this is a pet peeve of mine.)

```
const shouldAllowSubmit = state.instance_url && state.store_id && state.client_id && state.client_secret;
```

chriszarate:
> The purpose of this PR is to remove the old Salesforce B2C data source, and instead replace it with the Salesforce D2C data source.

@ingeniumed @alecgeatches I don't even know what these letters stand for, but is there value in keeping both?

inc/Integrations/SalesforceD2C/SalesforceD2CDataSource.php
```diff
@@ -19,15 +19,15 @@ protected static function get_service_config_schema(): array {
...
+			'instance_url' => Types::string(),
+			'store_id' => Types::string(),
 		] );
 	}
 
 	protected static function map_service_config( array $service_config ): array {
 		return [
 			'display_name' => $service_config['display_name'],
-			'endpoint' => sprintf( 'https://%s.api.commercecloud.salesforce.com', $service_config['shortcode'] ),
+			'endpoint' => $service_config['instance_url'],
```
alecgeatches: This is an improvement, I think the full URL makes things easier.

inc/Integrations/SalesforceD2C/SalesforceD2CIntegration.php
```diff
@@ -105,9 +99,9 @@ private static function get_queries( SalesforceB2CDataSource $data_source ): arr
 				'data_source' => $data_source,
 				'endpoint' => function ( array $input_variables ) use ( $base_endpoint, $service_config ): string {
 					return sprintf(
-						'%s/search/shopper-search/v1/organizations/%s/product-search?siteId=RefArchGlobal&q=%s',
+						'%s/services/data/v63.0/commerce/webstores/%s/search/products?searchTerm=%s',
```
alecgeatches, line 102: Any insight on the `v63.0` URL here? I'm just curious if this is a rapidly evolving API with docs for each version, or something different.

src/data-sources/salesforce-d2c/SalesforceD2CSettings.tsx
alecgeatches: I think he's talking about validation for the "Instance URL" input:

image

I think the example is adequate and we don't need a lot of field validation. However, it would be great if we could have some sort of full-page validation on the bottom of this page to confirm all fields work with a test request, something like we have for the Google Sheets connection:

sheets-connection-confirmation

That way we can make sure there aren't any problems before we save the connection details.

ingeniumed: I didn't add that as a todo on the PR but I've that on my list before this PR is finalized.

@chriszarate  - I should have made that clearer that I did mean validation of the url once it's inputted by the user.

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
ingeniumed: I used to prefer this but I remember hearing in a lot of reviews that people found it hard to follow in terms of the logic. So I ended up trying to strike a balance as to when one is used over the other. 

That said, in this case happy to switch to using null coalescing as it's honestly a lot shorter.

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
ingeniumed: In this case there is no expiry on the token, that's why it's done that way. I do want to clarify this API to ensure I've done it correctly. That's why for now there's no TTL.

src/data-sources/salesforce-d2c/SalesforceD2CSettings.tsx
ingeniumed: Always better to follow the campground rule for existing code, so happy to change this up.

inc/Integrations/SalesforceD2C/SalesforceD2CDataSource.php
ingeniumed: Yip, 100% agreed as I realized this url can change depending on if you have a custom domain, etc

inc/Integrations/SalesforceD2C/SalesforceD2CIntegration.php
ingeniumed: As discussed, this is to be clarified but it seems like it's the minimum version in which this API was made available?

inc/Config/QueryRunner/QueryRunner.php
ingeniumed: I agree, that's why I commented this line to ensure it would start a discussion. This was a workaround for me to get this data source working, and by no means should be the way it's implemented for all the other data sources.

Realistically, the query runner should check to see if a required input variable is provided and error out if it's not provided. The UI should ensure that such a query is not run in the first place, and the APIs should 400 out. 

That being said, for this particular case I think the search on the UI should be updated to never execute a query if nothing is entered. There's nothing to search, so why execute that?

alecgeatches:
@ingeniumed I'm having trouble getting any products to load with a client id/secret combination, and I'm also not seeing any errors. It might make sense for us to walk through this together to troubleshoot what might be going wrong. In the meantime, I noticed some other UI updates we might need to make for setup:

1. On the "New Data Source" page for Salesforce D2C, some of the inputs no longer line up for me:

    Screenshot 2025-02-17 at 11 59 43 AM

    1. **Instance URL**: This one is a bit odd because I'm not sure where the `.my.salesforce.com` URL should come from. When I load my store products page, I have a URL that looks like this (anonymized a bit):

        Screenshot 2025-02-17 at 11 55 53 AM

        The left-hand arrow shows the `lightning.force.com` domain here rather than the one in the example URL. I'm assuming the one in the example is correct, but I wasn't sure how to get it other than modifying the current URL to append `.my.salesforce.com` instead. This might be tricky though, ideally we can support all of Salesforce's various links for a store and extract the right organization ID, or just give the user a specific link to pull an address from during setup.

    2. Also using the image above, the **Store ID** input appears to be the `channelId` from the URL. If that's correct we should rename this to "Channel ID" and maybe, as above, provide some sort of easy explanation to get that ID.

    3. The **Client ID** example `bc2991f1-eec8-...` no longer appears to be correct for the new API key type, which is a longer string with no dashes.

1. Minor, the Salesforce D2C text isn't centered. In WordPress under Admin -> Settings -> Remote Data Blocks -> "Connect New" button, I noticed the Salesforce text is a little off-center:

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
```diff
@@ -0,0 +1,114 @@
...
+	 * @param string $client_id The client ID (a version 4 UUID).
+	 * @param string $client_secret The client secret.
+	 * @return string|WP_Error The token or an error.
+	 */
+	public static function generate_token(
+		string $endpoint,
+		string $client_id,
+		string $client_secret
+	): string|WP_Error {
+		return /* self::get_saved_access_token( $client_id ) ??  */self::get_token_using_client_credentials( $client_id, $client_secret, $endpoint );
```
ingeniumed: I've noticed a weird bug where the token is invalid, and I suspect it's because the token expires. But there's no expiry time on it so it's weird

ingeniumed: Yip, we need to hit `/services/oauth2/introspect` in order to see expiry time is

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
ingeniumed: Fixed this by using their introspect endpoint that has the expiry time info

inc/Config/QueryRunner/QueryRunner.php
ingeniumed: Lemme know what you think of the latest change I pushed up for this

inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -203,6 +203,11 @@ public function execute( HttpQueryInterface $query, array $input_variables ): ar
 			if ( ! array_key_exists( $key, $input_variables ) && isset( $schema['default_value'] ) ) {
 				$input_variables[ $key ] = $schema['default_value'];
 			}
+
+			// If the input variable is required and not provided, return an error.
+			if ( ! array_key_exists( $key, $input_variables ) && isset( $schema['required'] ) && $schema['required'] ) {
+				return new WP_Error( 'remote-data-blocks-missing-required-input-variable', sprintf( 'Missing required input variable: %s', $key ) );
```
chriszarate, line 209: I'll be curious if this breaks any existing queries. I'd like to move towards full validation using our `Validator` class, but this will be a nice first step.

src/blocks/remote-data-container/hooks/useRemoteData.ts
```diff
@@ -134,6 +135,44 @@ export function useRemoteData( {
...
+					metadata: {},
+					queryInput,
+					resultId: '',
+					results: [],
+				};
+				resolvedUpdater( resolvedDataNew );
+				setLoading( false );
+				return;
+			}
+		}
```
chriszarate: I'm not sure I follow these changes. It seems like it is implementing input validation for a single input type (based on the slug name) and returning empty results if the validation fails? I think we should prefer either of these alternate approaches, but option 1 is probably the better one:

1. Use the existing `searchAllowsEmptyInput` boolean, which is correctly keyed to the input field type, to implement form validation—that's the intended purpose of `searchAllowsEmptyInput`. I'm not sure why this approach isn't working for you, but:
   - Maybe you want to suppress rendering the search field entirely based on this value?
   - Maybe the logic that defines `searchAllowsEmptyInput` needs to be improved?
   - Maybe you need to prop-drill to make `searchAllowsEmptyInput` accessible where you need it?
2. Alternatively, now that you've added a check for `required: true` in the QueryRunner, just let the query fail and surface the error to the user.

src/data-sources/salesforce-d2c/SalesforceD2CSettings.tsx
ingeniumed: Implemented this in

inc/Config/QueryRunner/QueryRunner.php
ingeniumed: This has been fixed

src/blocks/remote-data-container/hooks/useRemoteData.ts
ingeniumed: This has been fixed

ingeniumed:
Screenshot 2025-02-20 at 3 01 34 pm

The PR is ready for review once more. I have made a bunch of more changes which are:

- Made the data source configuration page a lot more simpler. It now pulls in the stores to make picking the store easier, and only requires the domain instead of the full URL. This way, there's less chances of mistakes before it's saved.
- The search functionality for data sources where empty searches aren't allowed has been fixed. In addition, the optimistic query fetches won't happen for queries where required input variables are missing. Previously, this would fail.
- A lot more documentation has been added in areas that I had to spend some time deciphering.

To Do - the instructions for the Salesforce data source setup has to be updated. I haven't done it yet and will do it in a separate PR.

inc/REST/AuthController.php
```diff
@@ -34,7 +35,21 @@ public function register_routes(): void {
...
-				'permission_callback' => [ $this, 'get_google_auth_token_permissions_check' ],
+				'permission_callback' => [ $this, 'permissions_check' ],
+			]
+		);
+
+		/**
+		 * API to get Salesforce D2C Stores using the client_credentials grant type
+		 * This is also meant to test the credentials provided by the user.
+		 */
+		register_rest_route(
```
ingeniumed, line 46: My reasoning for adding this here was to have this as a test for the salesforce credentials being valid or not. If they are valid, the stores are given back otherwise nothing is. This also helps to keep the token obscure and works around CORs problems that kept happening. It doesn't directly make sense as to why a stores call is here but its really because of this being an auth token check. It needs the token to get this, that's why.

Happy to have a debate on this.

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	/**
+	 * Get the webstores using the given endpoint, and token.
+	 *
+	 * @param string $endpoint The endpoint prefix URL for the data source.
+	 * @param string $token The token.
+	 * @return array|WP_Error The webstores or an error.
+	 */
+	public static function get_webstores(
```
ingeniumed, line 37: This is here for the same reason that the AuthController has this method - it's just for token verification. For the checkout API we will likely need another token so it won't use this. I also am reluctant to reveal the token outside of the server just for security sake.

inc/Integrations/SalesforceD2C/Auth/SalesforceD2CAuth.php
```diff
@@ -0,0 +1,200 @@
...
+		string $endpoint,
+		string $token,
+	): array|WP_Error {
+		$webstores_url = sprintf( '%s/services/data/v63.0/query/?q=SELECT+name,id+from+webstore', $endpoint );
+
+		/* phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.wp_remote_get_wp_remote_get -- We don't only work on VIP so we can't rely on that. That said, we should safely implement a wrapper that uses it when it's available. */
```
ingeniumed, line 43: I went back and forth on this. We could create a wrapper for this wherein if `vip_` version exists then use that or else fallback to the default one.

types/localized-block-data.d.ts
```diff
@@ -1,49 +1,92 @@
...
+	/** Whether the variable is required */
 	required: boolean;
+	/** The slug of the variable */
 	slug: string;
+	/** The type of the variable */
 	type: string;
 }
 
+/**
+ * This corresponds directly to the overrides defined for a data source.
```
chriszarate: Overrides are defined at the block level and apply to query input variables.

types/localized-block-data.d.ts
```diff
@@ -1,49 +1,92 @@
 type RemoteDataBinding = Pick< RemoteDataResultFields, 'name' | 'type' >;
 type AvailableBindings = Record< string, RemoteDataBinding >;
 
+/**
+ * This corresponds directly to the input schema defined for a data source.
```
chriszarate:
```suggestion
 * This corresponds directly to the input schema defined by a query.
```

types/localized-block-data.d.ts
```diff
@@ -1,49 +1,92 @@
...
 	help_text?: string;
+	/** The slug of the variable */
 	name: string;
 }
 
+/**
+ * This corresponds directly to the block config.
+ */
 interface BlockConfig {
+	/** The available bindings for the block */
```
chriszarate: Nitpick: I love code comments, but many of the added comments here (including this line) don't add much value and slow down the reader. They aren't adding any additional context, they just rearrange the existing semantic tokens into a sentence. "The {property name} for the {instance name}" is already obvious and inferable.

src/blocks/remote-data-container/hooks/useRemoteData.ts
```diff
@@ -92,10 +92,21 @@ export function useRemoteData( {
...
+
+	if ( ! query ) {
+		throw new Error( `Selector not found for query key: ${ queryKey }` );
+	}
+
+	const inputVariables = query.inputs;
+
+	const requiredInputVariables = inputVariables.filter( input => input.required );
+
 	const [ data, setData ] = useState< RemoteData >();
```
chriszarate: Nitpick: I find it helpful when hooks are kept at the top of the component. It allows you to quickly understand what is stateful.

inc/REST/AuthController.php
alecgeatches: I think this is a good place for this code! It's part of the auth flow and matches the file's description:

https://github.com/Automattic/remote-data-blocks/blob/0202a6b8ce6dd10c986700067890ad707163349e/inc/REST/AuthController.php#L14-L19

chriszarate:
@ingeniumed I found some state bugs with the work we paired on yesterday, and I pushed up some additional changes in https://github.com/Automattic/remote-data-blocks/pull/372/commits/eb56e831ee5ce2aa45a8a3d9d8db99d7fa04dc94. Can you please test them?

I think this resolves the conflicts between use cases that want "optimistic fetch" and those that don't. It also improves query input validation and provides error state.

inc/REST/AuthController.php
```diff
@@ -68,11 +83,44 @@ public function get_google_auth_token( WP_REST_Request $request ): WP_REST_Respo
...
+				__( 'Client ID, client secret and domain are required.', 'remote-data-blocks' ),
+				array( 'status' => 400 )
+			);
+		}
+
+		$endpoint = 'https://' . $domain . '.my.salesforce.com';
+
+		$token = SalesforceD2CAuth::generate_token( $endpoint, $client_id, $client_secret );
+		if ( is_wp_error( $token ) ) {
+			return rest_ensure_response( $token );
```
chriszarate: Are we sure these errors don't leak anything sensitive?

alecgeatches:
@ingeniumed I poked around a little with SOQL and I think I found a way for us to prefetch items from a store:

```
GET https://xxx.my.salesforce.com/services/data/v63.0/query/?q=SELECT+name,id+from+Product2+LIMIT+10

{
  "totalSize": 10,
  "done": true,
  "records": [
    {
      "attributes": {
        "type": "Product2",
        "url": "/services/data/v63.0/sobjects/Product2/01tWs000005QwjpIAC"
      },
      "Name": "Alpine Energy Drink Chai Post 0.5L PET - 6 pack",
      "Id": "01tWs000005QwjpIAC"
    },
    {
      "attributes": {
        "type": "Product2",
        "url": "/services/data/v63.0/sobjects/Product2/01tWs000005QwjqIAC"
      },
      "Name": "Alpine Energy Post Eco Pod, Chai",
      "Id": "01tWs000005QwjqIAC"
    },
   /* ... */
  ]
}
```

I'm not sure how to get the full size of the results for paging in the same request, but as a separate request this works:

```
GET https://xxx.my.salesforce.com/services/data/v63.0/query/?q=SELECT COUNT() FROM Product2

{
  "totalSize": 253,
  "done": true,
  "records": []
}
```

alecgeatches:
@ingeniumed Really happy you got this working! I love the improved setup flow.

I'm still doing some testing based on @chriszarate's additions, but here are some UI issues I noticed:

1. I pushed [a small change](https://github.com/Automattic/remote-data-blocks/pull/372/commits/0202a6b8ce6dd10c986700067890ad707163349e) to fix the "Salesforce D2C" connection text alignment issue.

2. During connection setup, if I put in the whole Salesforce `.lightning.force.com` URL in the "Domain" input I get an error that just says "An error occurred":

    Screenshot 2025-02-20 at 12 03 13 PM

    I know the intention here is that I'm just supposed to put `scomhello123usa456org` instead of an `https://` address, but this highlights a couple of possible changes:

    - Can we also accept `https://scomhello123usa456org.lightning.force.com` and just regex out the domain?
    - If not, in the event that the domain is in an obviously wrong format like above, we should return a more relevant error if we can.

3. Related, I think having the error and success states show twice is too many times. I think we can remove the success/error status below "Client ID" and just use the one at the bottom below "Client Secret".

4. The item selection screen only shows pictures, but I'd expect more details like the item title at least:

    Screenshot 2025-02-20 at 12 04 26 PM

    Is there any particular limitation causing this?

alecgeatches:
@ingeniumed Follow-up to [this comment above](https://github.com/Automattic/remote-data-blocks/pull/372#issuecomment-2672599900). None of these UI polish changes are necessary for this PR to go out, so feel free to punt any or all of those. The last item (only images when selecting products) is probably the most important if any are to be addressed, if it's easy. Thank you!

inc/REST/AuthController.php
ingeniumed: Perfect, I'll leave this in then. At some point, if we need to do this for more data sources we can shift it to its own controller.

inc/REST/AuthController.php
ingeniumed: This is something I meant to remove before the final PR as it was only in place for testing. I'll take it out and replace it with a generic message.

types/localized-block-data.d.ts
ingeniumed: That's a fair point. I wanted to use this as an opportunity to start the comments, and have them be built over time. Instead, I'll take out most of em and just leave the ones that make sense in place.

ingeniumed:
What I've pushed up:

- Added new validation for the domain in the data source configuration page. Now, it'll automatically pick out the domain name if the url is in the form of `.my.salesforce.com` or `.lightning.force.com`. Otherwise, it'll spit out a proper error message right below the client secret if it's invalid.
- I've also made minor tweaks to the warnings that were spit out related to the alignment between the fields.
- I've also adjusted the connection warnings to only show up below the client secret now.
- As for the item selection, I thought that was already fixed from a previous data view update but I guess it's intermittent. The `Name` field shows up as hidden actually so it is available. I'll dig into this in a separate PR as it might be a data views bug from last time.

PR #379, opened 2025-02-18
pin psalm version in github actions to 6.5.1

shekharnwagh:
Pins the Psalm version in github actions to same as the one in `composer.json` (6.5). The difference in versions was leading to difference in output.


chriszarate (approved):
thanks

PR #383, opened 2025-02-21
Support Salesforce product overrides by SKU

alecgeatches:
This PR does three things:

1. Fixes an issue with overrides in `useRemoteData()` that caused the UI checkbox for overrides to un-check itself when a block was modified. This was due to the `enabledOverrides` property always defaulting to an empty array. It's now been modified to pull overrides from `externallyManagedRemoteData`.
2. Modify the Salesforce connection to use product SKUs (e.g. `12345-1`) over internal Salesforce product IDs (e.g. `01tWs000005ccuWIAQ`). SKUs are also unique values, but should be more easy for customers to understand during use.
3. Improve an error message related to query output parameters not matching query input parameters. I encountered this error during the Salesforce SKU changes. Previously, the error was:

    ```
     Error registering block scom: Cannot map key "product_sku" from search query.
    ```

    This took me a while to understand and debug. I've expanded it, which hopefully explains the problem in more depth:

    ```
    Error registering block scom: Cannot map key "product_sku" from search query. The display query for this block requires a "product_sku" key as an input, but it is not present in the output schema for the search query. Try adding a "product_sku" mapping to the output schema for the search query.
    ```

## Testing

1. Register a Salesforce D2C data source with products

2. Create a new post and add a product using the auto-generated block. You will need to search for a product first, "drink" is a good term to use:

    salesforce-product-block

4. Next, enable data override for the block in the sidebar:

    data-override-checkbox

    Then save and publish the post.

5. On the front-end version of the post, pass in a `?sku=` parameter. Some available SKUs from our internal `d2cstore1` store include: `6010020`, `6010049`, `6010053`.
6. Verify the data on the page changes:

    SKU override example on frontend page

### DIFF
--- a/inc/Editor/BlockManagement/ConfigRegistry.php
+++ b/inc/Editor/BlockManagement/ConfigRegistry.php
@@ -96,7 +96,7 @@ class ConfigRegistry {
 
 			foreach ( array_keys( $to_query->get_input_schema() ) as $to ) {
 				if ( ! isset( $from_output_schema['type'][ $to ] ) ) {
-					return self::create_error( $block_title, sprintf( 'Cannot map key "%1$s" from %2$s query. The display query requires a "%1$s" key as an input, but it is not present in the output schema for the %2$s query. Try adding "%1$s" to the output schema for the %2$s query.', esc_html( $to ), $from_query_type, $from_query_type, esc_html( $to ) ) );
+					return self::create_error( $block_title, sprintf( 'Cannot map key "%1$s" from %2$s query. The display query for this block requires a "%1$s" key as an input, but it is not present in the output schema for the %2$s query. Try adding a "%1$s" mapping to the output schema for the %2$s query.', esc_html( $to ), $from_query_type ) );
 				}
 			}
 
--- a/inc/Editor/BlockManagement/ConfigRegistry.php
+++ b/inc/Editor/BlockManagement/ConfigRegistry.php
@@ -96,7 +96,7 @@ class ConfigRegistry {
 
 			foreach ( array_keys( $to_query->get_input_schema() ) as $to ) {
 				if ( ! isset( $from_output_schema['type'][ $to ] ) ) {
-					return self::create_error( $block_title, sprintf( 'Cannot map key "%s" from %s query', esc_html( $to ), $from_query_type ) );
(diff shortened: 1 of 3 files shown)


inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -96,7 +96,7 @@ public static function register_block( array $user_config = [] ): bool|WP_Error
 
 			foreach ( array_keys( $to_query->get_input_schema() ) as $to ) {
 				if ( ! isset( $from_output_schema['type'][ $to ] ) ) {
-					return self::create_error( $block_title, sprintf( 'Cannot map key "%s" from %s query', esc_html( $to ), $from_query_type ) );
+					return self::create_error( $block_title, sprintf( 'Cannot map key "%1$s" from %2$s query. The display query requires a "%1$s" key as an input, but it is not present in the output schema for the %2$s query. Try adding "%1$s" to the output schema for the %2$s query.', esc_html( $to ), $from_query_type, $from_query_type, esc_html( $to ) ) );
```
chriszarate: Do you need those last added args to `sprintf`?

src/blocks/remote-data-container/hooks/useRemoteData.ts
```diff
@@ -94,7 +94,6 @@ interface UseRemoteDataInput {
 // don't need an intermediate state update / re-render.
 export function useRemoteData( {
 	blockName,
-	enabledOverrides = [],
```
chriszarate, line 97: Remove from input type above, please

inc/Editor/BlockManagement/ConfigRegistry.php
alecgeatches: Good catch! Removed those.

src/blocks/remote-data-container/hooks/useRemoteData.ts
alecgeatches: Fixed in https://github.com/Automattic/remote-data-blocks/pull/383/commits/8acf0b23990b72c39477fec5fe8294f5fe3992c1

chriszarate (approved):

ingeniumed (approved):

PR #384, opened 2025-02-23
Add workflow for AI documentation generation

alecgeatches:
Use [`repomix`](https://github.com/yamadashy/repomix) to generate a markdown documentation file for AI consumption.

This PR currently uses the `docs/` and `examples/` folders for generation, for a total of `33,150 tokens`. Here's a comparison of the options I tried:

- `repomix --style=markdown docs/`: `19,032 tokens`
- `repomix --style=markdown docs/ example/`: `33,150 tokens`
- `repomix --style=markdown docs/ src/`: `104,933 tokens`
- `repomix --style=markdown docs/ example/ src/`: `119,051 tokens`
- `repomix --style=markdown docs/ example/ src/ tests/`: `170,777 tokens`

I've tested locally that having `docs/ai.md` in the `docs/` folder doesn't recurse and include itself, doubling the output tokens. The output file specified by `-o` is ignored from the input data.

## Testing

You can also experiment with this locally to test [other `repomix` options](https://github.com/yamadashy/repomix?tab=readme-ov-file#command-line-options):

```bash
# In the remote-data-blocks/ directory

npm install -g repomix
repomix --style=markdown -o docs/ai.md <folder1> <folder2> <...>
```


.github/workflows/ai-docs.yml
```diff
@@ -0,0 +1,45 @@
+name: Generate AI Documentation
+
+on:
+  pull_request:
```
ingeniumed: Would it be better to run this post merge? So it seamlessly adds the necessary documentation? I could see some cases where a conflict arises but it would be a bit more efficient imo.

ingeniumed (approved):

.github/workflows/ai-docs.yml
alecgeatches: Good thinking! It would require a lot of `git pull`s that we can avoid. Plus, the action already checks for changes prior to committing, so having it run on based on specific PR file changes is unnecessary.

PR #387, opened 2025-02-24
Add salesforce d2c APIs as a bruno collection

ingeniumed:
### Description

This is my Bruno collection that's designed to make querying the various Salesforce D2C APIs easier. I have added them here in the docs in order for it to be re-usable by anyone. I have removed all the PII info and left them as placeholders instead.

### Testing Instructions

- Import the collection into Bruno
- Replace the placeholders with actual values: salesforce url, client ID, client Secret and a generated auth token.


brookewp:
This is so helpful, thank you! Would it make sense to move this under tutorials and update the Salesforce tutorial with a section on how to test using this collection? Including your testing step you mentioned:

>Replace the placeholders with actual values: salesforce url, client ID, client Secret and a generated auth token.

The Salesforce tutorial needs to be updated anyway, since it still references B2C and has an empty section under `Salesforce Commerce B2C API Access`:

https://github.com/Automattic/remote-data-blocks/blob/trunk/docs/tutorials/salesforce-commerce.md

ingeniumed:
> This is so helpful, thank you! Would it make sense to move this under tutorials and update the Salesforce tutorial with a section on how to test using this collection? Including your testing step you mentioned:
> 
> > Replace the placeholders with actual values: salesforce url, client ID, client Secret and a generated auth token.
> 
> The Salesforce tutorial needs to be updated anyway, since it still references B2C and has an empty section under `Salesforce Commerce B2C API Access`:
> 
> https://github.com/Automattic/remote-data-blocks/blob/trunk/docs/tutorials/salesforce-commerce.md

That was my eventual plan, but I hesitated in case there were some differences in the steps we have followed and what the actual recommendation is. That said, considering we have dropped the old integration entirely it might make sense to do some of this at least. Lemme update this PR

ingeniumed:
I've moved the API collection under tutorials, and updated the Salesforce doc to reference this directly. I've also updated the AI docs based on this.

Once we confirm that our steps are right, I'll add detailed instructions on how to generate the client ID and client Secret as well.

PR #391, opened 2025-02-25
Add pagination to Shopify integration

brookewp:
This implements the pagination feature (@chriszarate added in #338) to our Shopify integration.

To get the total product count, we'd need access to `read_products` which requires authenticating with OAuth. To avoid this, I've made `total_items` optional and added `has_next_page`, so we can paginate through products one page at a time: 

https://github.com/user-attachments/assets/954d9fac-4d5a-4c23-9971-825bbb37fdc6

I also exposed `setPerPage` so the items per page can be updated, which makes going through the products a lot faster. This will also update for any sources with pagination set up (e.g. the art institute).

### DIFF
--- a/inc/Validation/ConfigSchemas.php
+++ b/inc/Validation/ConfigSchemas.php
@@ -288,9 +288,8 @@ final class ConfigSchemas {
 			'pagination_schema' => Types::nullable(
 				Types::object( [
 					// This field provides an integer representing the total number of
-					// items available in paginated results. This field must be defined
-					// in order present to enable pagination support of any type,
-					// including cursor-based pagination.
+					// items available in paginated results. Either this field or
+					// `has_next_page` must be defined in order to enable
 					'total_items' => Types::nullable(
 						Types::object( [
 							'name' => Types::nullable( Types::string() ),
--- a/inc/Validation/ConfigSchemas.php
+++ b/inc/Validation/ConfigSchemas.php
@@ -291,11 +291,13 @@ final class ConfigSchemas {
 					// items available in paginated results. This field must be defined
 					// in order present to enable pagination support of any type,
 					// including cursor-based pagination.
-					'total_items' => Types::object( [
+					'total_items' => Types::nullable(
+						Types::object( [
 						'name' => Types::nullable( Types::string() ),
 						'path' => Types::json_path(),
 						'type' => Types::enum( 'integer' ),
 					] ),
+					),
 					// This field provides a pagination cursor for the next page of
 					// paginated results, or a null value if there is no next page. This
 					// field must be defined in order to enable cursor-based pagination.
--- a/inc/Validation/ConfigSchemas.php
+++ b/inc/Validation/ConfigSchemas.php
@@ -289,7 +289,7 @@ final class ConfigSchemas {
 				Types::object( [
 					// This field provides an integer representing the total number of
 					// items available in paginated results. Either this field or
-					// `has_next_page` must be defined in order to enable
+					// `has_next_page` must be defined in order to enable pagination.
 					'total_items' => Types::nullable(
 						Types::object( [
 							'name' => Types::nullable( Types::string() ),
--- a/inc/Validation/ConfigSchemas.php
+++ b/inc/Validation/ConfigSchemas.php
@@ -293,10 +293,10 @@ final class ConfigSchemas {
 					// including cursor-based pagination.
 					'total_items' => Types::nullable(
 						Types::object( [
-						'name' => Types::nullable( Types::string() ),
-						'path' => Types::json_path(),
-						'type' => Types::enum( 'integer' ),
-					] ),
+							'name' => Types::nullable( Types::string() ),
+							'path' => Types::json_path(),
+							'type' => Types::enum( 'integer' ),
+						] ),
 					),
 					// This field provides a pagination cursor for the next page of
 					// paginated results, or a null value if there is no next page. This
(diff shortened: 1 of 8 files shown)


src/blocks/remote-data-container/hooks/usePaginationVariables.ts
```diff
@@ -35,7 +36,7 @@ export function usePaginationVariables( {
 	initialPerPage,
 	inputVariables,
 }: UsePaginationVariablesInput ): UsePaginationVariables {
-	const [ paginationData, setPaginationData ] = useState< RemoteDataPagination >();
+	const [ paginationData, setPaginationData ] = useState< RemoteDataPagination >( {} );
```
chriszarate: Why this change?

types/remote-data.d.ts
```diff
@@ -3,9 +3,10 @@ interface InnerBlockContext {
 }
 
 interface RemoteDataPagination {
-	cursorNext?: string;
-	cursorPrevious?: string;
-	totalItems: number;
+    hasNextPage?: boolean;
+    cursorNext?: string;
+    cursorPrevious?: string;
```
chriszarate:
```suggestion
	cursorNext?: string;
	cursorPrevious?: string;
	hasNextPage?: boolean;
```

inc/Validation/ConfigSchemas.php
```diff
@@ -291,11 +291,13 @@ private static function generate_http_query_config_schema(): array {
 					// items available in paginated results. This field must be defined
 					// in order present to enable pagination support of any type,
 					// including cursor-based pagination.
```
chriszarate:
```suggestion
					// items available in paginated results. Either this field or
					// `has_next_page` must be defined in order to enable
					// pagination.
```

src/blocks/remote-data-container/components/item-list/ItemList.tsx
```diff
@@ -157,7 +162,7 @@ export function ItemList( props: ItemListProps ) {
 			onChangeView={ onChangeView }
 			paginationInfo={ {
 				totalItems: totalItems ?? data.length,
-				totalPages: totalPages ?? 1,
+				totalPages: totalPages ?? ( hasNextPage ? page + 1 : page - 1 ),
```
chriszarate: If neither `totalPages` nor `hasNextPage` are defined, I think we still want to default to `1`

types/remote-data.d.ts
brookewp: The alphabet is hard sometimes

src/blocks/remote-data-container/hooks/usePaginationVariables.ts
brookewp: Good catch! Unintended change, courtesy of Cursor 😅

inc/Validation/ConfigSchemas.php
```diff
@@ -288,14 +288,15 @@ private static function generate_http_query_config_schema(): array {
...
-					// items available in paginated results. This field must be defined
-					// in order present to enable pagination support of any type,
-					// including cursor-based pagination.
-					'total_items' => Types::object( [
-						'name' => Types::nullable( Types::string() ),
-						'path' => Types::json_path(),
-						'type' => Types::enum( 'integer' ),
-					] ),
+					// items available in paginated results. Either this field or
+					// `has_next_page` must be defined in order to enable
```
chriszarate: anticipation
```suggestion
					// `has_next_page` must be defined in order to enable pagination.
```

src/blocks/remote-data-container/components/item-list/ItemList.tsx
```diff
@@ -157,7 +162,7 @@ export function ItemList( props: ItemListProps ) {
 			onChangeView={ onChangeView }
 			paginationInfo={ {
 				totalItems: totalItems ?? data.length,
-				totalPages: totalPages ?? 1,
+				totalPages: totalPages ?? ( hasNextPage ? page + 1 : page - 1 ) ?? 1,
```
chriszarate: This doesn't quite work. The part in parentheses will always return a non-null value, so the `1` is never reached

src/blocks/remote-data-container/hooks/usePaginationVariables.ts
```diff
@@ -88,6 +89,7 @@ export function usePaginationVariables( {
 		supportsCursorPagination || supportsPagePagination || supportsOffsetPagination;
 	const totalItems = paginationData?.totalItems;
 	const totalPages = totalItems && perPage ? Math.ceil( totalItems / perPage ) : undefined;
+	const hasNextPage = paginationData?.hasNextPage;
```
chriszarate, line 90: can you update the `supports*Pagination` variables above to take into account the statement we made in the schema? "either total_items or `has_next_page` must be defined"

src/blocks/remote-data-container/components/item-list/ItemList.tsx
brookewp: Right. What do you think of using `Math.max`?

```js
totalPages: totalPages ?? ( hasNextPage ? page + 1 : Math.max( 1, page ) ),
```

src/blocks/remote-data-container/hooks/usePaginationVariables.ts
brookewp: Something like this? 
```js
	const supportsPagination =
		supportsCursorPagination ||
		supportsPagePagination ||
		supportsOffsetPagination ||
		totalItems ||
		hasNextPage;
```

brookewp: @chriszarate I've added this here [`b5e7856` (#391)](https://github.com/Automattic/remote-data-blocks/pull/391/commits/b5e7856adf60918d891d852e1ec69f060a5d967c), but let me know if you were thinking of something else!

src/blocks/remote-data-container/components/item-list/ItemList.tsx
brookewp: I found another issue with the code you are referring to, where instead of navigating to the last result, it returned to the first page. The following commit (and suggestion in above comment) fixes the issue, so I'm curious about your thoughts! @chriszarate 

[`da6fcd1` (#391)](https://github.com/Automattic/remote-data-blocks/pull/391/commits/da6fcd1a830c094855138390169eaff69c06d228)

chriszarate (approved):
Nice!

PR #396, opened 2025-02-27
ItemList: Add check for empty fields

brookewp:
@alecgeatches found that fields weren't showing for Salesforce after initiating a search.

This is because there are no results on the initial render for Salesforce. So the field value was remaining as an empty array even after searching/updating the view. 

This adds a `useEffect` to make sure that fields are updated if there are changes.

| Before  | After |
|  Screenshot 2025-02-27 at 12 24 37 PM | Screenshot 2025-02-27 at 12 35 44 PM  |

### DIFF
--- a/src/blocks/remote-data-container/components/item-list/ItemList.tsx
+++ b/src/blocks/remote-data-container/components/item-list/ItemList.tsx
@@ -130,7 +130,7 @@ export function ItemList( props: ItemListProps ) {
 
 	// Only update if tableFields has content and view.fields is empty
 	useEffect( () => {
-		if ( tableFields.length > 0 && view.fields.length === 0 ) {
+		if ( tableFields.length > 0 && view.fields?.length === 0 ) {
 			setView( currentView => ( {
 				...currentView,
 				fields: tableFields,
--- a/src/blocks/remote-data-container/components/item-list/ItemList.tsx
+++ b/src/blocks/remote-data-container/components/item-list/ItemList.tsx
@@ -1,6 +1,6 @@
 import { useInstanceId } from '@wordpress/compose';
 import { Action, DataViews, View } from '@wordpress/dataviews/wp';
-import { useState } from '@wordpress/element';
+import { useState, useEffect } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 
 import { ItemListField } from '@/blocks/remote-data-container/components/item-list/ItemListField';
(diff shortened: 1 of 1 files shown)


alecgeatches (approved):
I've tested that this fixes the Salesforce connection case locally. Thank you!

src/blocks/remote-data-container/components/item-list/ItemList.tsx
```diff
@@ -128,10 +128,24 @@ export function ItemList( props: ItemListProps ) {
...
 
+	// Only update if tableFields has content and view.fields is empty
+	useEffect( () => {
+		if ( tableFields.length > 0 && view.fields?.length === 0 ) {
+			setView( currentView => ( {
+				...currentView,
+				fields: tableFields,
+			} ) );
+		}
+	}, [ tableFields ] );
```
chriszarate, line 139: The effect runs whenever `tableFields` changes. When does that happen? Or does it only run once?

brookewp: It runs whenever the view changes.

If we don't want that to happen, I can think of a few alternatives which would run less. The following would only run initially to check if the `tableFields` are defined and wouldn't run for each subsequent view change: 

```jsx
	const areTableFieldsDefined =
		! loading && data.length > 0 && tableFields.length > 0 && view.fields?.length === 0;

	useEffect( () => {
		if ( ! areTableFieldsDefined ) {
			return;
		}
		setView( currentView => ( {
			...currentView,
			fields: tableFields,
		} ) );
	}, [ areTableFieldsDefined ] );
```

But that would give us a type error for exhaustive deps. 

To avoid adding `tableFields` to the dependency, we could add `useRef` and then it would only run in the scenario needed here to update the fields once they have been loaded: 

```jsx
	const areTableFieldsDefined =
		! loading && data.length > 0 && tableFields.length > 0 && view.fields?.length === 0;

	// Store tableFields in a ref so we can access it in the effect without dependency
	const tableFieldsRef = useRef( tableFields );
	// Keep the ref current
	tableFieldsRef.current = tableFields;

	useEffect( () => {
		if ( ! areTableFieldsDefined ) {
			return;
		}

		setView( currentView => ( {
			...currentView,
			fields: tableFieldsRef.current,
		} ) );
	}, [ areTableFieldsDefined ] );
```

But not sure if either of these are preferred or if there are any other ways to achieve this. Curious on your thoughts, @chriszarate !

brookewp:
Chatted with @chriszarate and this will be resolved in #406 so will close this

PR #397, opened 2025-02-27
Add title type

brookewp:
After seeing the missing name field in #396, I thought it might make sense to add a 'title' type to the schema to set explicit titles. In DataViews, there are three special fields: 

> titleField: The id of the field representing the title of the record.
> mediaField: The id of the field representing the media of the record.
> descriptionField: The id of the field representing the description of the record.

When these are set, they get unique styling, which we are utilizing for `image_url` in the `mediaField`, which can be seen in the screenshots below. 

This way, we don't have to guess like we do for block patterns. However, I left the check under `string` for sources like Airtable and Google Sheets, where we aren't explicitly defining each field and its type. 

| Before  | After |
| Screenshot 2025-02-27 at 12 52 37 PM |  Screenshot 2025-02-27 at 12 54 31 PM | 

| Before  | After |
| Screenshot 2025-02-27 at 12 55 01 PM  | Screenshot 2025-02-27 at 12 54 39 PM |

### DIFF
--- a/inc/Editor/BlockPatterns/BlockPatterns.php
+++ b/inc/Editor/BlockPatterns/BlockPatterns.php
@@ -112,10 +112,14 @@ class BlockPatterns {
 						$bindings['heading']['content'] = [ $field, $name ];
 						break;
 					}
-
+					
 					$bindings['paragraphs'][] = [
 						'content' => [ $field, $name ],
 					];
+					break;
+
+				case 'title':
+					$bindings['heading']['content'] = [ $field, $name ];
 					break;
 
 				case 'image_alt':
--- a/inc/Sanitization/Sanitizer.php
+++ b/inc/Sanitization/Sanitizer.php
@@ -119,6 +119,9 @@ class Sanitizer implements SanitizerInterface {
 			case 'string':
 				return sanitize_text_field( strval( $value ) );
 
+			case 'title':
+				return sanitize_text_field( strval( $value ) );
+
 			case 'button_text':
 			case 'html':
 			case 'id':
--- a/inc/Sanitization/Sanitizer.php
+++ b/inc/Sanitization/Sanitizer.php
@@ -117,8 +117,6 @@ class Sanitizer implements SanitizerInterface {
 				return null;
 
 			case 'string':
-				return sanitize_text_field( strval( $value ) );
-
 			case 'title':
 				return sanitize_text_field( strval( $value ) );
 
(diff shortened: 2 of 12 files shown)


alecgeatches:
This is great, and would definitely help with auto-generation! What do you think about `string:title` or something like that? Like a type with hinting instead of a separate type. We could extend the same hints for other types like `id:primary`, `image_url:primary`, `string:description`, etc. I like the idea of adding information to the type rather than adding a different one, but it's subjective which is preferable.

inc/Sanitization/Sanitizer.php
```diff
@@ -119,6 +119,9 @@ public static function sanitize_primitive_type( string $type_name, mixed $value
 			case 'string':
 				return sanitize_text_field( strval( $value ) );
```
chriszarate:
```suggestion
```

inc/Editor/BlockPatterns/BlockPatterns.php
```diff
@@ -112,12 +112,16 @@ public static function register_default_block_pattern( string $block_name, strin
...
 						break;
 					}
-
+					
 					$bindings['paragraphs'][] = [
 						'content' => [ $field, $name ],
 					];
 					break;
 
+				case 'title':
```
chriszarate, line 121: Maybe we should remove the "autodetect" above and encourage use of this type?

chriszarate (approved):
I dig it too.

> What do you think about string:title or something like that? Like a type with hinting instead of a separate type. We could extend the same hints for other types like id:primary, image_url:primary, string:description, etc.

I like encoded it in the type as well. The only danger there is that you limit yourself in the future by tying it to another primitive type. For example, maybe we decided later that you wanted to indicate a level alongside the title, e.g. `'title' => [ 'level' => 'h1', 'content' => 'My title' ]`

I pushed up a small fix in https://github.com/Automattic/remote-data-blocks/pull/397/commits/311223f57ada9262d64108d87bb075ce97f0d8f8 to supply the missing type definition and to close a loophole that allowed you to provide invalid types to the `Validator`.

inc/Editor/BlockPatterns/BlockPatterns.php
brookewp: >I left the check under string for sources like Airtable and Google Sheets, where we aren't explicitly defining each field and its type
@chriszarate ^ This was tucked away in my PR description. Curious what you think in regards to this.

ingeniumed (approved):

inc/Editor/BlockPatterns/BlockPatterns.php
chriszarate: Makes sense, thanks for pointing me

PR #399, opened 2025-03-01
updating to work with WP Playground CLI

smithjw1:
This will work, but we lose query monitor, so we may not want to go this way.


chriszarate (approved):

chriszarate:
QM support will not land soon I don't think. I have an idea to surface logs differently when QM is not available.

PR #406, opened 2025-03-05
Support bulk selection by default for every block

chriszarate:
> [!NOTE]
> This pull request is large, mostly because the changes it makes affect server-side config schema, query execution, and block bindings as well as client-side rendering, data curation, and data fetching logic. I recommend reviewing by commit.

The initial motivation for this pull request was to revisit a few issues found when reviewing #369:

- The method used to render multiple results led to non-fatal query execution errors, because the child blocks were not "hydrated" in the expected way:

<p>Screenshot 2025-03-05 at 12 33 50</p>

- The `supports_bulk` property could only be successfully applied to ID fields, but there was no enforcement of this.
- Similarly, the bulk selection implementation within DataViews silently discarded any non-ID query inputs.

These issues were relatively minor, but as @brookewp and I discussed it, we realized that there were some architectural decisions that probably needed revisiting. Doing so would fix the above issues and allow bulk selection for any query type, not just those that opted-in. 

They also pave the way for removing "loop" blocks entirely, since both "single" and "loop" blocks are now rendering in the same fundamental way. That work is left for a future PR.

### Significant changes in this PR

- All remote data blocks are now [dynamic blocks](https://developer.wordpress.org/block-editor/getting-started/fundamentals/static-dynamic-rendering/#dynamic-rendering). Previously, only loop blocks were dynamic—so that they could duplicate their "template" for each result in the remote data set. Now, single blocks also need this capability—since the user can select one or more items. 
   - This also allows us to use the render callback function to "hydrate" the remote data in a consistent way, reducing redundant query executions. (Though these were resolved from in-memory cache.)
   - In the future, we might also implement fallback behavior via this render callback function.
- Allow batch query executions via the `QueryRunner` and REST API endpoint. This reduces boilerplate in both server-side and client-side code to loop over multiple requests.
   - Update block editor code to support batch queries.
- Allow multiple / bulk selection for every render query by default. If the query supports an `id:list` input variable type, it can be resolved by a single query. Otherwise the query is repeated for each selected item.
- Provide a UUID for each remote data result which can be used as an ID in contexts that require one, like DataViews.
- Use a consistent response shape for remote data in the block editor. Currently, we simplify and transform the shape of remote data to a flat object in order to make it friendlier for JavaScript code. However, this leads to confusion and loss of type safety. Instead, we can provide utility functions to work with this data.
- Remain backwards-compatible with blocks that were created prior to this pull request.

That's a lot! The result is more consistency and type safety and better capabilities around batch querying. Here's a short video of multi-select for a block that has done absolutely nothing to opt-in:

https://github.com/user-attachments/assets/f7a2f5f4-ae20-442c-a71f-d1790351abe5

### Known issues

- Multi-select doesn't currently work with synced patterns and this PR has not fixed that issue. I'll work to fix that separately.


brookewp:
Thanks for opening this and improving bulk selection! I'm still testing, but I just noticed that it's no longer possible to deselect: 

deselect

chriszarate:
> I just noticed that it's no longer possible to deselect

On it! Thanks

chriszarate:
> I just noticed that it's no longer possible to deselect

@brookewp This is fixed in https://github.com/Automattic/remote-data-blocks/pull/406/commits/3dd66d85aac555eac8b78010e538467b2f250d07, but DataViews does not seem to preserve selections across pages even though the underlying state is correct. Is that a known issue?

brookewp:
>DataViews does not seem to preserve selections across pages even though the underlying state is correct. Is that a known issue?

I have a few related comments about it in #369 after @alecgeatches pointed this out. This is the behaviour in the uncontrolled component, where I don't think the underlying state is correct? It only seems to be possible when using `selection` and `onChangeSelection` and controlling the state of the selections ourselves. 

But to me, it seems unintentional in the uncontrolled component, and it would be worth checking to see if that's the case or if it's a bug that needs to be reported. So I can follow up with that.

chriszarate:
> It only seems to be possible when using selection and onChangeSelection and controlling the state of the selections ourselves.
>
> But to me, it seems unintentional in the uncontrolled component, and it would be worth checking to see if that's the case or if it's a bug that needs to be reported.

This PR changes the implementation to be a controlled component, see `DataViewsModal`. The problem is still present, so I think it's an upstream bug.

brookewp:
>This PR changes the implementation to be a controlled component, see DataViewsModal. The problem is still present, so I think it's an upstream bug.

It is controlled in trunk / from #369, and the selections are preserved there: 

https://github.com/user-attachments/assets/1acfb621-5b5f-44fd-b01d-c7017f3a41f8

If this is what you are referring to. I can't replicate this in this branch, so I think some changes removed the behaviour.

chriszarate:
Silly me, this has nothing to do with DataViews. This is because the UUIDs attached to each result are changing on page navigation because the results are being re-fetched. I will add some memoization to the underlying fetch which will both fix this issue and improve performance. Thanks @brookewp for awesome feedback and detail as always!

chriszarate:
Now that the remote data results contain type information that was previously stripped out, @brookewp and I agreed that using the first result to introspect the DataViews columns instead of `availableBindings` made sense. 

I've made this change in https://github.com/Automattic/remote-data-blocks/pull/406/commits/14547fdb933b18343315b09e7c8f150165455743

This effectively means that we are using the output schema of the selection query to display data in DataViews instead of the output schema of the render query.

This also fixes #396 as a side-bonus.

brookewp:
It looks like it's not possible to 'Choose' a single item anymore, and results aren't being saved to the front end: 

https://github.com/user-attachments/assets/ef32aeac-0a93-4231-a918-b5cecb126fa1

I ran into the second issue when implementing bulk selection, but it was just when there were multiple items and not single items, too. I can look back at what happened there and share if it's helpful.

chriszarate:
> It looks like it's not possible to 'Choose' a single item anymore

Thanks! I introduced a regression in one of my fixes. Addressed in https://github.com/Automattic/remote-data-blocks/pull/406/commits/72bd6e4134c649292e892e90bd03af59524c8ca6

I also improved the logic so that the actions column is only displayed when relevant: https://github.com/Automattic/remote-data-blocks/pull/406/commits/465a183f95a6993e3cb2a443496cee535859fe95

brookewp:
I've just tested the most recent changes with the art institute example, airtable, and shopify, and all return this network error: 

```js
body
: 
{code: "rest_missing_callback_param", message: "Missing parameter(s): query_inputs",…}
code
: 
"rest_missing_callback_param"
data
: 
{status: 400, params: ["query_inputs"]}
params
: 
["query_inputs"]
0
: 
"query_inputs"
status
: 
400
message
: 
"Missing parameter(s): query_inputs"
```

chriszarate:
Oof, @brookewp, apologies for continually wasting your time! 

- Addressed some schema issues that weren't caught by existing tests. https://github.com/Automattic/remote-data-blocks/pull/406/commits/548dceff23190932e91616277dfce3c9e3233b1e
- Fixed a typo and added tests to cover these regressions https://github.com/Automattic/remote-data-blocks/pull/406/commits/7b84420e684fd62bc13c45f3b6fb6266a3769d84

brookewp (approved):
>apologies for continually wasting your time!

Not a waste of time! Manually testing these changes made me think about how I should have added some tests for ItemList before, since we've been making a lot of changes lately. So I've added a few tests to start here: https://github.com/Automattic/remote-data-blocks/pull/413

However, I did manually test your recent changes and everything looks good to me! ✅

brookewp:
@chriszarate I just noticed a warning in the console after the initial data is first rendered (before it's been cached) when opening the DataViewsModal:

```
react-dom.js?ver=18.3.1.1:73 Warning: The final argument passed to useEffect changed size between renders. The order and size of this array must remain constant.

Previous: [, ]
Incoming: [, , Name, Properties bought/interested in, Phone number, Email address] Error Component Stack
    at ItemList (ItemList.tsx:35:3)
    ```

chriszarate:
> Incoming: [, , Name, Properties bought/interested in, Phone number, Email address] Error Component Stack
>     at ItemList (ItemList.tsx:35:3)

That stack trace points to a weird place, but there's only one effect in that component so I think I got it. I'll raise a PR, thanks!

PR #413, opened 2025-03-10
Add tests for the ItemList component

brookewp:
There have been quite a few recent changes to the ItemList component (aka our implementation of DataViews), and there are more upcoming, so I thought it might be helpful to test expected behaviours for users regardless of implementation. 

I plan to add more tests, but we would need to consider some changes before those work. 

I've also added `@testing-library/user-event` which is [the preferred method over fireEvent by testing library ](https://testing-library.com/docs/user-event/intro/) for simulation of user actions. 

So I can follow up by replacing `fireEvent` for our other tests.


chriszarate (approved):
Preapproving this awesome PR!

brookewp:
>I plan to add more tests, but we would need to consider some changes before those work.

I'll merge this as the baseline. And then we can follow up with more tests as we go. 🙂

PR #414, opened 2025-03-11
Use string as effect dependency instead of array with indeterminate length

chriszarate:
Fixes the React warning noted in [this comment](https://github.com/Automattic/remote-data-blocks/pull/406#issuecomment-2711909636).


brookewp (approved):
Thanks for resolving this error!

PR #415, opened 2025-03-11
Fix migration of serialized data

chriszarate:
Another fix related to #406 and migration of previously serialized block data and field shortcodes.


brookewp:
~Thanks for looking into this! It looks like it's still happening. I can consistently replicate using Salesforce when loading/resetting the results~

Ignore this comment, grabbed the wrong branch name!

PR #416, opened 2025-03-11
Improve code organization and bundle size

chriszarate:
This PR does a few things:

- Prevents filters like `withBlockBinding`, `registerFormatType`, and other registrations from running multiple times (once per remote data block). Instead, actions that only need to run once should live and be enqueued separately. This has probably been a small performance drag, especially on sites with lots of registered remote data blocks.
- Does some corresponding code reorg to reflect this new separation. There is more we could do here, but I'll save it for future PRs.
- Splits `@wordpress/dataviews/wp` into a separate shared script dependency to reduce bundle sizes. This asset is huge (1.7 MB) so separating allows it to be reused instead of bundled three times (once for the plugin settings page, another for the container block, another for the field shortcode).

Previous build size:
```sh
> du -hd 1 build
1.9M	build/blocks
1.9M	build/settings
 40K	build/pattern-editor
3.8M	build
```

After:

```sh
> du -hd 1 build
196K	build/blocks
252K	build/settings
1.7M	build/dataviews
 76K	build/block-editor
 40K	build/pattern-editor
2.2M	build
```

This is still not great, but the bulk of this size is coming from DataViews.

There are no functional code changes, just moving and renaming.


src/dataviews/index.ts
```diff
@@ -0,0 +1,6 @@
+import { DataViews, filterSortAndPaginate } from '@wordpress/dataviews/wp';
+
+window.LockedPrivateDataViews = {
+	filterSortAndPaginate,
+	DataViews,
+};
```
maxschmeling, line 14: It would probably be good to put a comment in this file about why it's needed.

maxschmeling (approved):

src/dataviews/index.ts
chriszarate: Done in https://github.com/Automattic/remote-data-blocks/pull/416/commits/e7564f6b91bd80156be2dac3be42982505695098

PR #417, opened 2025-03-12
[WIP] split config schema

shekharnwagh:


inc/Config/DataSource/DataSourceQuery.php
```diff
@@ -0,0 +1,106 @@
...
+use RemoteDataBlocks\Config\ArraySerializable;
+use RemoteDataBlocks\Validation\ConfigSchemas;
+use RemoteDataBlocks\Validation\Validator;
+use WP_Error;
+
+/**
+ * Base class for data source query configurations.
+ * 
+ * This class represents the query-specific parameters for a data source,
+ * separate from the connection/authentication information.
```
chriszarate, line 14: Should we just merge this into the query, then? I don't think I understand the need for separation.

shekharnwagh: HttpDataSource and HttpQuery classes are different in that HttpDataSource only needs the config as input and nothing else, while for HttpQuery, those configs are just one of the inputs required amongst others.

The query configs (base/table for Airtable, spreadsheet/sheet for Google Sheets, etc.) currently don't map 1:1 with our Query/HttpQuery classes since one such query config would need multiple HttpQuery instances to register blocks. For example, selecting 2 tables in Airtable would need at least 2 queries - list and search queries for each of the tables. I tend to think of configs as minimum data required for us to generate the queries and register blocks which for Airtable would mean just knowing the tables and the schema for each and our integration code would use that to generate the HttpQuery class.

The new `DataSourceConnection` and `DataSourceQuery` classes can focus on just one thing - validating the configs. Having the config split would allow reusing the connections.

Slack [thread](https://a8c.slack.com/archives/C06HTDH3STW/p1740696890550449) for more context.

shekharnwagh: Open to suggestions if this sounds like its complicating things

chriszarate: > those configs are just one of the inputs required amongst others

What are the others?

shekharnwagh: `input_schema`, `output_schema`, `preprocess_response` which might not always be derived from the query config.

chriszarate: Could we provide defaults that could then be edited in the UI, if needed?

maxschmeling:
Closing this old WIP branch to keep the PR queue focused. Any config/schema split work that still matters should be reintroduced deliberately from current trunk.

PR #405, opened 2025-03-05
Add multiple queries to single block

brookewp:
WIP

Initial commit shows proof of concept with minimal changes... but not an ideal direction. 

TBC

Screenshot 2025-03-04 at 5 45 45 PM

PR #418, opened 2025-03-13
Expose inputs for loop blocks

brookewp:
This might be replaced by the work in #405, but it's a small step towards the goals there.

This adds a special type `ui:input` and `loop` type selector when `loop: true` so we can use that to show inputs when relevant. 

Returning all results filtered by an input (e.g., categories/collections) is a common use case, so examples of this have been added for Shopify and Salesforce. I also added an example in the Art Institute to set the number of results returned. A visual example of all three of these can be seen here: 

https://github.com/user-attachments/assets/8af847e7-f4dd-4c45-85fc-f04faab58d57

**Updated** to include input in the query panel to update any selection: 

https://github.com/user-attachments/assets/74cc5e50-a144-4256-af99-977abdfe4d8c

>[!NOTE]  
>I ran into a few things that we will want to consider when setting up queries, relating to the UI. One thing is a way to add text to the placeholder, as the generic text won't work in all cases and might not be the same as a block description. So for now, in this PR, I've added `instructions`,  but this could change with #405. Also open to alternative suggestions here!

## Testing steps

- Try new block(s) and add input
	- Salesforce: Products By Category with category ID 
	- Shopify: Products by Collection with collection ID 
	- Art Institute of Chicago Loop 

- Select the parent block and edit the ID in the block inspector panel
	- Edit multiple IDs (Shopify will have single inputs per ID, and Airtable/Art Insitute will have a separate list).


alecgeatches:
@brookewp This looks great! I'm excited for both the filter options and the easy retargeting a block to a new ID. Testing now.

alecgeatches:
@brookewp Noticed an odd bug that happens after you pick an item from the Art Institute block, and then pick another one on the same page:

https://github.com/user-attachments/assets/71bc0edf-8a33-45a3-b6df-d0e670e23aa6

## Reproduction steps

1. At the bottom of the root plugin file `remote-data-blocks.php` , add this line to enable the Art Institute example:

    ```
    require_once __DIR__ . '/example/rest-api/art-institute/art-institute.php';
    ```

2. Create a new post, insert the "Art Institute of Chicago" block.

3. Click the "Choose" button and select an art piece.
4. Click "Choose a pattern" and select the default pattern.
5. Delete the block, and re-insert an "Art Institute of Chicago" block.
6. Select an art piece.
7. Click "Choose a pattern", but see that the page is blank this time.

## Notes

- I can't seem to reproduce this in `trunk` (but I'm not 100% sure this is your problem)
- I can't reproduce this in the current branch (`try/loop-with-input`) with a Salesforce block, so it seems specific to the art institute block for now.
- There aren't any console logs to help

example/rest-api/art-institute/art-institute.php
```diff
@@ -156,6 +201,16 @@ function register_aic_block(): void {
 		],
 	]);
 
+	register_remote_data_block( [
+		'title' => 'Art Institute of Chicago Loop',
+		'icon' => 'art',
+		'instructions' => 'This block displays a set amount of artworks based on the provided limit.',
```
alecgeatches, line 207: I like this! Here's what just this looks like for anyone interested:

image

example/rest-api/art-institute/art-institute.php
```diff
@@ -89,6 +89,51 @@ function register_aic_block(): void {
...
+					'type' => 'id',
+				],
+				'artist_title' => [
+					'name' => 'Artist Title',
+					'type' => 'string',
+					'path' => '$.artist_title',
+				],
+				'title' => [
+					'name' => 'Title',
+					'type' => 'string',
```
alecgeatches: Might be a good block to add some [`title`-type](https://github.com/Automattic/remote-data-blocks/pull/397) supports?

src/blocks/remote-data-container/components/panels/QueryInputsPanel.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+					onUpdateQueryInputs( cleanedInputs );
+				} }
+			>
+				{ localInputs.map( ( input, index ) =>
+					Object.entries( input ).map( ( [ key, value ] ) => {
+						const displayValue = Array.isArray( value ) ? value.join( ',' ) : ( value as string );
+
+						return (
+							<TextControl
```
alecgeatches, line 40: If I'm understanding this correctly, this includes support for multiple inputs in the sidebar, awesome! Do we currently have any blocks that use multiple inputs?

brookewp: We don't have any examples with explicit multiple inputs, but you can see an example with multiple inputs when testing out bulk select with Shopify (or any data source that doesn't have `id:list` in their input schema: 

Screenshot 2025-03-20 at 12 46 42 PM

But interesting to think about... like if someone added 100 items: 

https://github.com/user-attachments/assets/ee00f65e-0095-4ab0-80c7-b03e17f92904

Maybe a good reason to have the Query Inputs panel below the Management one with Refresh/Reset 😅

example/rest-api/art-institute/art-institute.php
brookewp: Indeed 🥲 Nice catch.

src/blocks/remote-data-container/components/panels/QueryInputsPanel.tsx
alecgeatches: True! I think this is a strict improvement for even 100 items selected, but it'd be better to have those last, just in case.

alecgeatches (approved):
With the exception of [that one pattern issue](https://github.com/Automattic/remote-data-blocks/pull/418#issuecomment-2741387672) (that I'm not 100% sure is related to this feature), this works great! Feel free to merge if you fix the pattern issue or determine it to be unrelated.

brookewp:
Thanks @alecgeatches! I can replicate the issue you found in trunk after rebuilding my env. I'll open an issue/PR for that separately.

brookewp:
Just to note, I kept the art institute example but removed the extra test blocks from Shopify and Salesforce here: [`f1ae091` (#418)](https://github.com/Automattic/remote-data-blocks/pull/418/commits/f1ae0913d90a8326bd10917db340d34e2fddf4a1)

I plan to add info and examples to the docs instead.

PR #419, opened 2025-03-13
Bugfix: Sanitzer null check

chriszarate:
The Sanitizer nuked nullable boolean `false` values (like `'required' => false`) because false values are `empty()`. This fixes it and adds some additional tests.

### DIFF
--- a/tests/inc/Sanitization/SanitizerTest.php
+++ b/tests/inc/Sanitization/SanitizerTest.php
@@ -7,44 +7,121 @@ use RemoteDataBlocks\Sanitization\Sanitizer;
 use RemoteDataBlocks\Validation\Types;
 
 class SanitizerTest extends TestCase {
-	public function test_sanitize_string(): void {
+	public static function provideSanitizableStrings(): array {
+		return [
+			[ ' John Doe ', 'John Doe' ],
+			[ ' Jane Doe', 'Jane Doe' ],
+			[ 'Alice', 'Alice' ],
+			[ ' Bob ', 'Bob' ],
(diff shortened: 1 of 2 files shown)


tests/inc/Sanitization/SanitizerTest.php
```diff
@@ -7,44 +7,121 @@
...
-		$this->assertSame( 25, $result['age'] );
+		$this->assertSame( $expected, $result['age'] );
 	}
 
-	public function test_sanitize_boolean(): void {
+	public static function provideSanitizableBooleans(): array {
+		return [
+			[ true, true ],
+			[ 'true', true ],
+			[ 'false', true ],
```
maxschmeling, line 75: Is this what we want? I know you're just adding tests for existing functionality here. Wonder if 'false' should be `false` though.

maxschmeling (approved):

tests/inc/Sanitization/SanitizerTest.php
chriszarate: Not sure.

One on hand, the value should be fixed. Why should `'false'` be a magic string? What about `'False'`, `'FALSE'`, `'0'`, or `'__return_false'`? Where's the line?

On the other, maybe this comes up a lot? It's worth noting that in most scenarios, the data would have already gone through and failed validation.

maxschmeling: Yeah, I gave it some more thought and I think it's best as is. It's consistent with how these values work in JS at the very least.

PR #420, opened 2025-03-13
Bugfix: Don't cache error results

chriszarate:
Bugfix to prevent memoizeFn from caching Promise rejections.

### DIFF
--- a/tests/src/utils/function.test.ts
+++ b/tests/src/utils/function.test.ts
@@ -37,6 +37,25 @@ describe( 'function utils', () => {
 			expect( fn ).toHaveBeenCalledTimes( 3 );
 		} );
 
+		it( 'should cache a then-able promise if that is returned by the function', async () => {
+			const fn = vi.fn().mockImplementation( async ( arg: number ): Promise< string > => {
+				await Promise.resolve();
+				return `called with: ${ arg }`;
+			} );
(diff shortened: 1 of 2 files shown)


tests/src/utils/function.test.ts
```diff
@@ -18,5 +18,97 @@ describe( 'function utils', () => {
...
+			const memoized = memoizeFn< ( arg: number ) => Promise< string > >( fn );
+
+			expect( await memoized( 1 ) ).toEqual( 'called with: 1' );
+			expect( await memoized( 1 ) ).toEqual( 'called with: 1' );
+
+			const thenPromise = new Promise< string >( ( resolve, reject ) => {
+				memoized( 1 ).then( resolve ).catch( reject );
+			} );
+
+			expect( await thenPromise ).toEqual( 'called with: 1' );
```
maxschmeling, line 54: Does this provide useful extra assurance over using `await memoized( 1 )` twice?

maxschmeling (approved):

tests/src/utils/function.test.ts
chriszarate: Yes, if the original function returns a promise, it ensures that the value returned by the memoized function is also a promise (as opposed to the awaited value). This was a bug in my initial commit.

maxschmeling: Oh, I see. That makes sense now. 👍

PR #421, opened 2025-03-13
Surface query errors via error boundary

chriszarate:
https://github.com/user-attachments/assets/220623d6-a942-47f5-9799-ebef25619f80

The error duplication should be separately fixed by #416


alecgeatches (approved):
This is great and sorely needed! I noticed that error responses in DataViews can obscure the error until the view is exited:

https://github.com/user-attachments/assets/e0050cfd-8574-4272-809e-c1f221cc8dd3

Note: this request URL was modified to make it fail purposefully.

It'd be great if we had a way to also display errors in there, but either way this is an improvement as-is. Thank you!

chriszarate:
The `error` is returned by the hook, so it should be possible, but I don't see any [documented error handling in `DataViews`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-dataviews/). @brookewp Anything come to mind?

brookewp:
Ya, there isn't anything built into DataViews for this. How about adding a Notice as a child of our `DataViewsModal` component?

https://github.com/user-attachments/assets/c46506d8-7ef1-4fb3-b4b5-2ffca522a03e

jarekmorawski:
These are all great ideas. Thanks for looking into them! The notice works well because it's very prominent, but what if the user is currently editing five different blocks? The error doesn't specify which block is affected, and while we could make it smart and actionable, I think we'll be better off simplifying the UI and displaying the error in the block placeholder.

- The error state is shown in the block,
- The error title will be the same for all connection errors (= "Cannot connect") unless there are any other error types that may require customized copy and CTA,
- When the block is not focused, the block has a distinctive placeholder with an error icon,
- When the user clicks the Show error details button, the placeholder expands and reveals a code block with the API response. The user can select its content and copy it like any other text string,
- Clicking "Try again" will attempt to reconnect to the source,
- The sidebar is unaffected and still displays the block's details without the settings and customization options.

Could this work?

error_state

chriszarate:
> Could this work?

Thanks so much for looking @jarekmorawski! I'll give it a shot.

brookewp:
@chriszarate I've created a new `PlaceholderError` component based on @jarekmorawski's design above. I can directly commit the changes if preferred, but I figured I'd share a diff in case you want to take a look first. 

I'm curious about your thoughts on implementing this. I would have thought to add it to `edit.tsx`, but the hook there is just looking at the display query key so it wouldn't catch all errors. 

Here is a demo of what the component looks like:

https://github.com/user-attachments/assets/57371c38-062c-42b6-8488-deaef5c1c3b4

```diff
diff --git a/src/blocks/remote-data-container/components/placeholders/PlaceholderError.tsx b/src/blocks/remote-data-container/components/placeholders/PlaceholderError.tsx
new file mode 100644
index 00000000..f3209a6b
--- /dev/null
+++ b/src/blocks/remote-data-container/components/placeholders/PlaceholderError.tsx
@@ -0,0 +1,48 @@
+import { Button, Placeholder, __experimentalHStack as HStack, Icon } from '@wordpress/components';
+import { useState } from '@wordpress/element';
+import { __, sprintf } from '@wordpress/i18n';
+import { caution } from '@wordpress/icons';
+
+import { RemoteDataFetchError } from '../../hooks/useRemoteData';
+
+interface PlaceholderErrorProps {
+	blockName: string;
+	error: Error;
+	onTryAgain: () => void;
+}
+
+export function PlaceholderError( { blockName, error, onTryAgain }: PlaceholderErrorProps ) {
+	const [ showErrorDetails, setShowErrorDetails ] = useState( false );
+
+	let message = error.message;
+	if ( error instanceof RemoteDataFetchError && error.cause instanceof Error ) {
+		message = `${ message }: ${ error.cause.message }`;
+	}
+
+	return (
+		<Placeholder
+			className="remote-data-container-error"
+			icon={ caution }
+			label={ sprintf( __( 'Cannot connect to remote data source: (%s)' ), blockName ) }
+		>
+			<Icon className="remote-data-container-error__icon-overlay" icon={ caution } />
+			<HStack justify="flex-start">
+				<Button variant="secondary" onClick={ onTryAgain }>
+					{ __( 'Try again' ) }
+				</Button>
+				<Button variant="tertiary" onClick={ () => setShowErrorDetails( ! showErrorDetails ) }>
+					{ sprintf( __( '%s error details' ), showErrorDetails ? 'Hide' : 'Show' ) }
+				</Button>
+			</HStack>
+			{ showErrorDetails && (
+				<code
+					className={ `remote-data-container-error__content is-${
+						showErrorDetails ? 'open' : 'closed'
+					}` }
+				>
+					{ message }
+				</code>
+			) }
+		</Placeholder>
+	);
+}
diff --git a/src/blocks/remote-data-container/editor.scss b/src/blocks/remote-data-container/editor.scss
index 238fc2d4..90185279 100644
--- a/src/blocks/remote-data-container/editor.scss
+++ b/src/blocks/remote-data-container/editor.scss
@@ -4,6 +4,69 @@
 
 @import url(@wordpress/dataviews/build-style/style.css);
 
+// selected error placeholder
+.rdb-container.is-selected .components-placeholder.remote-data-container-error {
+
+	.components-placeholder__label > svg {
+		fill: var(--Gutenberg-Alert-Red, #cc1818);
+	}
+
+	.remote-data-container-error__content {
+		width: auto;
+		background: var(--Gutenberg-Gray-100, #f0f0f0);
+		color: var(--Gutenberg-Gray-900, #1e1e1e);
+		line-height: 1.5;
+		padding: 12px 16px;
+		border-radius: 2px;
+	}
+
+	// hide overlay icon
+	.remote-data-container-error__icon-overlay {
+		display: none;
+	}
+}
+
+// unselected error placeholder
+.rdb-container:not(.is-selected) .components-placeholder.remote-data-container-error {
+	background: var(--Gutenberg-Gray-100, #f0f0f0);
+	box-shadow: inset 0 0 1px 1px var(--Gutenberg-Alert-Red, #cc1818);
+
+	// hide content when block is not selected
+	*:not(.remote-data-container-error__icon-overlay):not(.remote-data-container-error__icon-overlay *) {
+		visibility: hidden;
+	}
+
+	// hide open error details when block is not selected for consistent sizing
+	.remote-data-container-error__content.is-open {
+		display: none;
+	}
+
+	// add strikethrough across placeholder
+	&::before {
+		content: "";
+		position: absolute;
+		top: 0;
+		left: 0;
+		width: 100%;
+		height: 100%;
+		background: var(--Gutenberg-Gray-200, #e0e0e0);
+		clip-path: polygon(1% 1%, 2% 1%, 99% 99%, 98% 99%);
+	}
+
+	// show overlay icon when block is not selected
+	.remote-data-container-error__icon-overlay {
+		visibility: visible;
+		width: 48px;
+		height: 48px;
+		background: var(--Gutenberg-Gray-100, #f0f0f0);
+		fill: var(--Gutenberg-Gray-200, #e0e0e0);
+		position: absolute;
+		top: 50%;
+		left: 50%;
+		transform: translate(-50%, -50%);
+	}
+}
+
 .remote-data-blocks-loading-overlay {
 	align-items: center;
 	background-color: rgba(255, 255, 255, 0.25);

```

chriszarate:
@brookewp Push it up with thanks! I think `edit.tsx` can destructure `error` and pass it down

brookewp:
>Push it up with thanks!

Done! [`23c9de0` (#421)](https://github.com/Automattic/remote-data-blocks/pull/421/commits/23c9de0819bbf07eccf7a9b5769da8d0ef567264)

>I think edit.tsx can destructure error and pass it down

When destructuring `error` here, I was seeing `error` was `undefined` (I broke the search query endpoint to trigger an error):

https://github.com/Automattic/remote-data-blocks/blob/500df719ec9bc0e5d3f4a03f09ea4248bca108a7/src/blocks/remote-data-container/edit.tsx#L42-L47

chriszarate:
`error` is simple React state, can you trace why `setError` is not being called? Or can you push up your current state in a branch off this one?

brookewp:
>error is simple React state, can you trace why setError is not being called? Or can you push up your current state in a branch off this one?

I have a test branch here: `test/improve/block-editor-query-notices`

I'm looking into it, but I'm sure you'll be quicker than me.

brookewp:
Btw, if it's helpful, `setError` is being called, but only the display query is passed to `useRemoteData` in `edit.tsx`. And the error I was testing was for the search query. Since the hook is called in the DataViewsModal, we see the console error.  

**Edit:** I had added comments in my test branch to demonstrate this, but not sure if they stand out so wanted to mention:
https://github.com/Automattic/remote-data-blocks/blob/7486060906dad61863480ef616c4713a7a6b29d5/src/blocks/remote-data-container/edit.tsx#L47-L50

So I'm not sure if we'd want modify edit to look at the other keys, or adjust how the error works in `useRemoteData`, or maybe use context... 🤔

chriszarate:
Ah, I see. So we have the "render" usage of `useRemoteData` at the top level, then another use of `useRemoteData` in `DataViewsModal` further down the component tree. We need shared state between them to capture errors for all uses of useRemoteData.

We could use context. What about throwing and using an error boundary? Are we able to recover / try again?

brookewp:
>We could use context. What about throwing and using an error boundary? Are we able to recover / try again?

I can give that a try and report back

brookewp:
Sorry for the delay on my report 😄 

It worked to add an error boundary [`56dd8a2` (#421)](https://github.com/Automattic/remote-data-blocks/pull/421/commits/56dd8a26472b4349d4d909414e3d0ec8682f2c0e). However, I'm not super familiar with those/class components (which I believe is still the preferred option in order to use `getDerivedStateFromError`).

When testing an error, I see:

```
Uncaught RemoteDataFetchError: Client error: `GET https://api.artic.edu/api!/v1/artworks?limit=10&page=1` resulted in a `404 Not Found` response:
{"status":404,"error":"Sorry, something went wrong.","detail":"An unrecognized exception was thrown. Our developers have (truncated...)

    at useRemoteData.ts:27:9
    at Generator.next (<anonymous>)
    at fulfilled (usePatterns.ts:113:1)
hook.js:608 The above error occurred in the <DataViewsModal> component:
```

Is it expected to see an additional error where the error occurred?

chriszarate:
I have revived this PR, benefiting from the wonderful work done by @brookewp, who we dearly miss. The main update was moving the error boundary so that it could catch errors thrown via the hooks called in the `Edit` component. I take no credit for this, just pushing forward the existing work by Brooke.

This video simulates some errors via the built-in example API:


https://github.com/user-attachments/assets/b848dd7a-1f92-4f06-a7fd-dedb487bd051

src/blocks/remote-data-container/hooks/useRemoteData.ts
```diff
@@ -224,9 +217,10 @@ export function useRemoteData( {
 		onSuccess?.();
 	}
 
-	function reset(): void {
+	function reset( err?: Error ): void {
+		console.log( 'reset', err );
```
ingeniumed: Should this be removed?

src/blocks/remote-data-container/components/placeholders/PlaceholderError.scss
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	// hide overlay icon
+	.remote-data-container-error__icon-overlay {
+		display: none;
+	}
+}
+
+// unselected error placeholder
+.rdb-container:not(.is-selected) .components-placeholder.remote-data-container-error {
```
ingeniumed, line 25: I wonder if it would be better to at least some have text or UI showing when it's not selected?

ingeniumed (approved):
Works really well with the error block as well 👍🏾 

I had two comments, one of which is minor and the other doesn't have to be a blocker for this PR.

src/blocks/remote-data-container/components/placeholders/PlaceholderError.scss
chriszarate: Not sure I follow. Do you mean a test to ensure that the selected state is applied? That is handled by core so I'm not sure how we would unit test it. Or do you mean some other UI state should be applied?

ingeniumed: I meant to say text, that was a typo apologies. I've corrected it to say text, so that one would be able to tell that the block is in an error state which is why it appears that way. It won't render when the post is viewed, so this would be just for the block editor

chriszarate: We are closely following @jarekmorawski's design, so I'll defer to him on this suggestion. It does bring up a general a11y concern to make sure that the error is discoverable / announceable.

src/blocks/remote-data-container/hooks/useRemoteData.ts
chriszarate: Done in https://github.com/Automattic/remote-data-blocks/pull/421/commits/05a96bdd8f9b2a730b275d4ed7bbf4ba869860c1, thanks!

PR #426, opened 2025-03-18
Use getValues for bindings

maxschmeling:
Uses the `getValues` argument of `registerBlockBindingsSource` to retrieve binding values async with core data.


src/blocks/remote-data-container/index.ts
```diff
@@ -52,11 +56,109 @@ addFilter(
...
+const queryDataStateKey = (queryKey: string, blockName: string, queryInputs: Record<string, string>) => `${blockName}:${queryKey}:${JSON.stringify(queryInputs ?? {})}`;
+
+const remoteDataBlocksStoreConfig: ReduxStoreConfig< State, Actions, Selectors > = {
+	reducer: ( state = {}, action ) => {
+		switch ( action.type ) {
+			case 'RECEIVE_REMOTE_DATA':
+				// console.log('store: RECEIVED REMOTE DATA')
+				// console.log({action});
+				// console.log(`key: ${queryDataStateKey(action.queryKey, action.blockName, action.queryInput)}`);
+				return { ...state, [ queryDataStateKey(action.queryKey, action.blockName, action.queryInputs) ]: action.data };
```
chriszarate, line 76: Each query response contains a UUID in the `resultId` property. Maybe we should use that?

chriszarate: Nevermind, that won't work since the UUIDs will be different on subsequent page loads. 👎

src/blocks/remote-data-container/index.ts
```diff
@@ -52,11 +56,109 @@ addFilter(
...
+
+register( remoteDataBlocksStore );
+
 registerBlockBindingsSource( {
 	name: 'remote-data/binding',
 	label: 'Remote Data Binding',
 	usesContext: [ 'remote-data-blocks/remoteData' ],
-	getValues() {
-		return {};
+	getValues( { context, clientId, bindings, select, ...other } ) {
```
chriszarate, line 124: Should we be using `setValues` too? Or should we be dispatching actions from `useRemoteData()`?

maxschmeling:
Closing this as obsolete. The getValues-based binding approach landed later in #654 (`src/block-editor/binding-sources/remote-data-binding.ts`), and this draft is now both out of date and superseded by the current implementation.

PR #428, opened 2025-03-21
Remote Data Template block

chriszarate:
As a prerequisite for pagination, this PR introduces a Remote Data Template block that serves to encapsulate the blocks that render remote data:

https://github.com/user-attachments/assets/6d512855-26df-4536-9604-97608a245828

This block's function (and implementation) is very similar to Core's [Post Template block](https://wordpress.org/documentation/article/post-template-block/): It allows us to configure a single representation of remote data in the editor, then use it as a template during render.

We had a half-measure towards this functionality in the existing `LoopTemplate` component; however, without a dedicated, registered block, we are unable to access the block and its contents during render. It is also needed so that we can insert other blocks (like pagination blocks) that are explicitly not included in the template—and therefore not repeated for each item.

While working on this, the existing "loop" blocks became superfluous. We already effectively perform loop rendering in non-loop blocks via multi-select. The hardcoded "Load items" button was the only remaining functionality keyed to the injected `loop` property, and it was easily replaced by a new "selector" type. This can be further improved by Brooke's work but for now it simply preserves the current functionality.

This change is **backwards-compatible** and will not break existing remote data blocks that were saved without a template block.

Major changes:

- Introduce a Remote Data Template block that is automatically inserted whenever remote data is selected.
- Remove "loop" blocks and move existing template functionality into the new template block.
- Add `render.php` to the `remote-data-container` block to make its dynamic rendering explicit.
- Update the rendering code in `BlockBindings` to look for and use a template block, when present.


brookewp:
@chriszarate there were a lot of small conflicts after I merged #418, so I resolved those here: [`865dea2` (#428)](https://github.com/Automattic/remote-data-blocks/pull/428/commits/865dea2d4feb9d5bbdace6a4b861dbd7de932fea). 

_Disclaimer: Nothing was changed/removed from this PR._

inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -78,14 +76,33 @@ public static function register_block( array $user_config = [] ): bool|WP_Error
...
+				[
+					'image_url' => $display_query->get_image_url(),
+					'inputs' => [],
+					'name' => 'Load collection',
+					'query_key' => self::DISPLAY_QUERY_KEY,
+					'type' => 'load-collection',
+				]
+			);
+		}
+
```
brookewp: I think instead of adding another selector, we should add a conditional check to the single selector to see if it's type `input` or `load-collection` (can we rename it to `collection`)? So it behaves the same as before. Otherwise, we will get the 'Load collection' button and 'Provide manual input' buttons, like this: 

An example of a collection with no inputs: 
Screenshot 2025-03-25 at 2 47 31 PM

And the built-in Airtable loop block:
Screenshot 2025-03-25 at 3 29 01 PM

brookewp: I borrowed the logic here for a related fix: https://github.com/Automattic/remote-data-blocks/pull/435

But I added it as a conditional in the `render_query` as I mentioned above. But if there's a reason it should be a selector, we can use the version here instead!

chriszarate:
Thanks for the improvements, @brookewp! I fixed a couple more conflicts in https://github.com/Automattic/remote-data-blocks/pull/428/commits/82cb060c323f75bdbe561bdb240a2295493aa51f

chriszarate:
> I think instead of adding another selector, we should add a conditional check to the single selector to see if it's type input or load-collection (can we rename it to collection)?

@brookewp We separately need to check if the query is loading a collection or not. It could be loading a single item. I've made this change in https://github.com/Automattic/remote-data-blocks/pull/428/commits/3ba760afcc419718230012e0e6a8ddaf35a28f08

brookewp (approved):
This looks great and is working well! 🎉 

A small non-blocker, I wonder if we can hide the 'Remote Data' Panel when it's empty as it might be confusing?

Screenshot 2025-04-01 at 10 00 22 AM

chriszarate:
> A small non-blocker, I wonder if we can hide the 'Remote Data' Panel when it's empty as it might be confusing?

Good catch, thanks! Will tackle in a separate PR.

PR #434, opened 2025-03-25
Fix: pattern selection missing after making duplicate selection

brookewp:
@alecgeatches [found an issue](https://github.com/Automattic/remote-data-blocks/pull/418#issuecomment-2741387672) after selecting the same item in succession. When doing so, no patterns are available to select in the modal: 

https://github.com/user-attachments/assets/22158ccf-fcdc-4235-92cc-365f64c42620

This appears to be due to a brief moment where `__experimentalGetAllowedPatterns` returns an empty array, and its state isn't updated. I believe this happened because `__experimentalGetAllowedPatterns` was extracted from `useSelect` as a function reference rather than tracking its return value.

## Solution

In this PR, `__experimentalGetAllowedPatterns(rootClientId)` is now inside the selector function, to ensure its result is included in `useSelect`'s state tracking. 

https://github.com/user-attachments/assets/fd78a682-7b7e-4b51-b44a-7baac154d8b4

## Testing

After selecting an item in a block (replicated in the art institute and airtable blocks), add another block and make the same selection. You should see patterns to select (ensure custom patterns are present in addition to default).

### DIFF
--- a/src/blocks/remote-data-container/hooks/usePatterns.ts
+++ b/src/blocks/remote-data-container/hooks/usePatterns.ts
@@ -35,13 +35,27 @@ export function cloneBlockWithAttributes(
 export function usePatterns( remoteDataBlockName: string, rootClientId: string = '' ) {
 	const { patterns } = getBlockConfig( remoteDataBlockName ) ?? {};
 	const { replaceInnerBlocks } = useDispatch< BlockEditorStoreActions >( blockEditorStore );
-	const { getBlocks, getPatternsByBlockTypes, __experimentalGetAllowedPatterns } =
-		useSelect< BlockEditorStoreSelectors >( blockEditorStore, [
-			remoteDataBlockName,
-			[ remoteDataBlockName, rootClientId ],
-		] );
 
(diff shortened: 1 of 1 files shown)


chriszarate:
> I believe this happened because __experimentalGetAllowedPatterns was extracted from useSelect as a function reference rather than tracking its return value.

This is correct, and my fault because I don't think I read [the docs](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#useselect) closely:

> When data is only used in an event callback, the data should not be retrieved
on render, so it may be useful to get the selectors function instead.
>
> Don’t use useSelect this way when calling the selectors in the render
function because your component won’t re-render on a data change.

I will do a review to find other spots where I made this mistake. 

Note I pushed up a type improvement in https://github.com/Automattic/remote-data-blocks/pull/434/commits/ad920afd274727a938c9d359abdf2f7bbc1c9e55.

chriszarate (approved):

alecgeatches (approved):
Thank you!

PR #436, opened 2025-03-28
add hook to signal plugin loaded

shekharnwagh:
### Description

This PR adds a new action hook to signal the plugin is loaded and adds an early return to prevent multiple instances of the plugin from being loaded.

- Added a check to prevent multiple instances of the Remote Data Blocks plugin from being loaded by returning early if the plugin is already defined ( [`remote-data-blocks.php`](diffhunk://#diff-5f9e69557260c9fd4c96c84baf735c2a61e39819b8c406f626c19d4ef64d71aaR19-R24) ).
- Added a `do_action` call to fire the `remote_data_blocks_loaded` action, indicating that the plugin is fully loaded and ready for use. This would be useful for plugins that need to defer their initialization until remote data blocks is ready ( [`remote-data-blocks.php`](diffhunk://#diff-5f9e69557260c9fd4c96c84baf735c2a61e39819b8c406f626c19d4ef64d71aaR60-R61) ).

### DIFF
--- a/docs/extending/hooks.md
+++ b/docs/extending/hooks.md
@@ -10,6 +10,25 @@ There are two types of hooks: Actions and Filters. To use either, you need to wr
 
 Actions allow you to add data or change how WordPress operates. Actions will run at a specific point in the execution of plugin. Callback functions for an Action do not return anything back to the calling Action hook.
 
+### remote_data_blocks_loaded
+
+This action fires when Remote Data Blocks is fully loaded and ready for use. Plugins that depend on Remote Data Blocks should use this hook to defer their initialization until Remote Data Blocks is fully loaded.
+
+```php
+function my_plugin_init() {
+	// Initialize your plugin that depends on Remote Data Blocks here
+	// All Remote Data Blocks classes and functionality are now available
+}
+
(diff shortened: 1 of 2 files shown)


alecgeatches (approved):
Looks good to me! We do something similar in [the Block Data API](https://github.com/Automattic/vip-block-data-api/blob/e3fd40c4c441c45fd355f50767783c95a4910fcb/vip-block-data-api.php#L20) and [governance plugin](https://github.com/Automattic/vip-governance-plugin/blob/9202d9d263dd44c12d946e4f32e7905db2685908/vip-governance.php#L20), but we use `<plugin-name>_LOADED` instead of the version. Either is fine, though. I like the `_loaded` hook too, useful!

shekharnwagh:
> in [the Block Data API](https://github.com/Automattic/vip-block-data-api/blob/e3fd40c4c441c45fd355f50767783c95a4910fcb/vip-block-data-api.php#L20) and [governance plugin](https://github.com/Automattic/vip-governance-plugin/blob/9202d9d263dd44c12d946e4f32e7905db2685908/vip-governance.php#L20), but we use <plugin-name>_LOADED instead of the version

Updated to use the constant `REMOTE_DATA_BLOCKS__LOADED` so that it's obvious what it denotes and makes it consistent with other VIP Plugins.

PR #439, opened 2025-03-31
Remove salesforce as a data source

ingeniumed:
### Description

The purpose of this PR is to remove Salesforce as a data source in the public facing version of this plugin. Any existing salesforce data source would appear as an HTML data source and when attempting to edit will show that it's an unsupported service.

There still is the pending feature of allowing the ability to insert data sources dynamically via other methods outside of the plugin. That should be a follow up PR given how complex that will be.


chriszarate:
> Any existing salesforce data source would appear as an HTML data source and when attempting to edit will show that it's an unsupported service.

Shouldn't it be completely disabled in some way and marked as such?

alecgeatches (changes requested):
This works great! My only requested changes are in the vein of [Chris' suggestion above](https://github.com/Automattic/remote-data-blocks/pull/439#issuecomment-2767352759). Here's a quick runthrough of the different options available on my Salesforce data source ("sf-tag") after I updated the plugin:

https://github.com/user-attachments/assets/ead1b255-0184-4494-99f4-5957236748bb

- **Edit**: Shows a "Service not (yet) supported." page. Ideally we don't show this button at all if it doesn't go anywhere.
- **Copy UUID**: This still works, which seems fine to me if you're trying to find a deactivated source in code or something. I think we could remove or keep this option, either is fine.
- **Delete**: Works as expected.
- **Duplicate**: Fails with an error, we should also remove this one.
- **View Code**: Fails with console errors on a blank page. Remove this as well.

I think that deactivated sources should potentially show just "Copy UUID" and "Delete" actions. It may also be good to add a warning to the table page, maybe "Data source no longer available" in the "META" column, or even an error icon somewhere.

The process of removing a built-in source may happen again in the future, or happen with custom sources, so it would be good to have nice UI defaults for that case. Thank you!

ingeniumed:
This was partially why I had kept it in draft status - I figured there would be something I might have missed. This is 100% something that could be improved so the experience isn't the way it is right now.

Lemme dig into it, especially the options you mentioned @alecgeatches and see what I come up with.

Note: I should have marked that as a ToDo or a pending item in my PR description to flag that.

ingeniumed:
In light of what I said above, I have put it back into draft status.

inc/WpdbStorage/DataSourceCrud.php
```diff
@@ -36,8 +36,33 @@ public static function get_config_by_uuid( string $uuid ): array|WP_Error {
 		return new WP_Error( 'data_source_not_found', __( 'Data source not found', 'remote-data-blocks' ), [ 'status' => 404 ] );
 	}
 
-	public static function get_configs(): array {
-		return self::get_all_configs();
+	public static function get_configs(): array|WP_Error {
```
ingeniumed: @alecgeatches / @chriszarate - This is the fix I'm proposing for what you two raised.

The logic I had in my head was that, this should follow the same logic as when you request an item that's disabled or soft deleted. It wouldn't be returned back under any circumstance. What that means in practice for us is,

- Get the configs
- Inflate each one against the list of known classes that we have
- If it doesn't inflate, then it means that's not valid
- Only give back all the valid configs

As part of making the UI extendable, we would def tackle the point Alec made about custom data sources. Right now, data sources injected via a constant follow this exact logic. So this is bringing it in line with that logic. This fits with the rest way of working imo, as why would want our UI to reveal that a data source is unsupported when it's technically an item that should not even be revealed to the client?

I have added one extra step here which is removal of unsupported data sources. This is set to false by default, as it felt like a destructive operation that we don't want to enable it yet. If it's set to true, it'll remove all the invalid configs and cleanup the DB. I also have to test it after we have enabled support for data sources that could be injected. I don't want to have them interfere and cause bad consequences. I'm happy to remove this if you two prefer.

What I think this doesn't do is add the ability to mark a data source as disabled, and then be able to re-enable it. That is a different feature imo, for when a data source is supported and one chooses to enable/ disable it. This would be when a data source isn't even supported in the first place.

ingeniumed: Note: Previously, we never inflated the configs when fetching all of em. So this is a good addition to have in place.

chriszarate: > The logic I had in my head was that, this should follow the same logic as when you request an item that's disabled or soft deleted. It wouldn't be returned back under any circumstance. What that means in practice for us is,
>
> - Get the configs
> - Inflate each one against the list of known classes that we have
> - If it doesn't inflate, then it means that's not valid
> - Only give back all the valid configs

I don't agree with this from a UX / DevEx perspective. I think an invalid data source should be surfaced in the data sources list, with a clear UI indicator that it is disabled because it is invalid. Ideally, we also surface any associated error messages. It doesn't have to be accomplished via this class method, but it needs to be possible.

Silently removing invalid data sources will lead to confusion. And we definitely don't want to auto-remove them from the persisted storage (option).

> What I think this doesn't do is add the ability to mark a data source as disabled, and then be able to re-enable it.

This has not been requested AFAIK.

alecgeatches: Agree with Chris above! I think we want to show a disconnected/unavailable source for a couple of reasons:

- If a data source type has been removed, this is a good place to communicate that something is messed up. Silently ignoring makes it trickier to find
- If there's an unexpected configuration issue, we don't want to drop data sources accidentally

ingeniumed: I get what you two are saying - just show it as disabled with the option to delete it (via the UI that is) and view any error messages necessary for this. I had been approaching this from a slightly different mindset, which was to think about all the possible cases where a data source could be disabled. I saw 4 possible cases:

- The data source has auth or other config problems. An example would if the client ID was wrong, or the auth type has been deprecated and this is an old data source.
- The data source has failed to migrate, and there is an error in this data source.
- The data source has been manually disabled.
- The data source's class itself is no longer supported. This is the case we are solving right now.

To me, that 4 case stood out as an odd ball compared to the other 3 cases. This is because, to me this is like bad data in that there is no reason to show it in the UI save to delete it. The user cannot do anything with that info, save for acking it or delete it. In addition, it does add a bit of barrier if you leave our platform that any data source which we don't support will not be shown in the data sources list.

Hence, I did it the way that I did it. I did add the auto-removal code in there but disabled it for the exact reason you said. I was 50/50 on it, more leaning towards no than yes tbh.

All this to say, that I wanted to give some more background into why I did it the way I did. What I'll do next is to try and disable this data source in the way suggested above. I'm hoping to do it in a way that other data sources can take advantage of it as well, for the other use cases I outlined above. If it's proving too challenging, I'll stick to just the one use case.

inc/WpdbStorage/DataSourceCrud.php
```diff
@@ -36,8 +36,20 @@ public static function get_config_by_uuid( string $uuid ): array|WP_Error {
...
+		$configs = self::get_all_configs();
+		$valid_configs = [];
+
+		foreach ( $configs as $config ) {
+			$instance = self::inflate_config( $config );
+
+			// If the data source is valid, set a transient field called enabled to true or false otherwise.
+			$config['enabled'] = ! is_wp_error( $instance );
```
ingeniumed: I've added this new transient field whose sole purpose is to reflect if that data source should be enabled or not. Given that this method is called anytime a get is requested, regardless of it being a get all or get one (we should look to optimize this using our learnings from the workflow plugin), this will always be set.

I didn't add this in the `migrate` method as the output of the inflated config isn't returned and that method wouldn't be called for configs that aren't tied to a supported data source.

src/data-sources/DataSourceList.tsx
```diff
@@ -152,8 +149,18 @@ const DataSourceList = () => {
 		table: {},
 	};
 
-	const isItemEligibleForActions = ( item: DataSourceConfig ) => {
-		return item.config_source === ConfigSource.STORAGE;
+	const isItemEligibleForActions = ( item: DataSourceConfig, action: string ) => {
```
ingeniumed, line 152: What I have done is, using the enabled status to only allow delete and copy actions. Everything else is disabled.

src/data-sources/DataSourceList.tsx
```diff
@@ -152,8 +149,19 @@ const DataSourceList = () => {
...
+		if ( item.config_source !== ConfigSource.STORAGE ) {
+			return false;
+		}
+
+		// This is meant to limit actions on disabled items. For an item to be disabled, it must have the enabled field set to false.
+		// The only actions allowed on disabled items are delete and copy as they won't cause any unintended side effects.
+		const actionsAllowedForDisabledItems = [ 'delete', 'copy' ];
```
ingeniumed, line 159: This is the new check on the UI that limits the operations on disabled items. Note that I am explicitly looking for the enabled flag to be set to false, so that in case a data source is given back without it it won't lead to bugs (that shouldn't happen but this catches it).

I've also added some explanation for the logic here as well.

inc/WpdbStorage/DataSourceCrud.php
ingeniumed: I've modified the code to now do the following:

- Set the enabled flag on each data source. This will be true for classes we support, and false otherwise.
- The settings UI will use the enabled flag to limit operations on disabled items to just copy and delete.

This will also leave the capacity to surface error messages within the config, if we wanted. It'll also allow us to add a filter or some other mechanism to allow third party data sources to have this flag be set too. So those can re-use this logic as well.

Hopefully I have covered my bases on this.

alecgeatches:
@ingeniumed I tried adding a SF data source on `trunk`, and then switching to this branch, and my data source is no longer present. Here were my steps:

1. In `trunk`, add a new Salesforce data source.
2. Run `git checkout remove/salesforce-data-source`, `git pull`, and then `npm run build`.
3. Go back to the Remote Data Blocks settings page, and see that the Salesforce data source is no longer visible.

From my understanding above I should be able to see a source, just with limited options. Is there something missing there?

inc/WpdbStorage/DataSourceCrud.php
```diff
@@ -37,7 +37,19 @@ public static function get_config_by_uuid( string $uuid ): array|WP_Error {
 	}
 
 	public static function get_configs(): array {
-		return self::get_all_configs();
+		$configs = self::get_all_configs();
+		$valid_configs = [];
+
+		foreach ( $configs as $config ) {
+			$instance = self::inflate_config( $config );
```
alecgeatches: I see a couple of things I don't understand:

- Before we returned an array of uninflated `$config` arrays, but now we return an array of instantiated classes.
- Nit, we're adding all configurations into the `$valid_configs` array, although the name to me implies it's just the valid ones.

Can we use a simpler check to see if a configuration is valid instead of inflating classes? Maybe

```php
public static function get_configs(): array {
	return array_map(
		function ( array $config ) {
			$data_source_class = REMOTE_DATA_BLOCKS__DATA_SOURCE_CLASSMAP[ $config['service'] ] ?? null;

			if ( null === $data_source_class ) {
				$config['enabled'] = false;
			} else {
				$config['enabled'] = true;
			}

			return $config;
		},
		self::get_all_configs()
	);
}
```

or if inflation is necessary to tell if a source is invalid, we should still modify `$configs` and return that instead, rather than changing the output type here.

alecgeatches (approved):
Requested one change above related to how we check if sources are valid and to avoid data-type changes (unless they were intentional). However other than that, testing worked great and I think this is good to go.

inc/WpdbStorage/DataSourceCrud.php
ingeniumed: It would mean we aren't truly validating if it's a valid config or not. We are doing a quick check to see if its supported, without validating the whole thing against the data source it's for. This is the reason I had kept it in.

I could see an argument for the performance being impacted if there are a large enough number of data sources, and if the inflation is expensive but I don't think that'd be the case. I'd lean towards keeping it in.

In regards to the output type being changes, I haven't changed the output type. It's still giving back the configs, albeit with an enabled field added in there. The inflation of the config is to decide if they are valid or not, and has no bearing on the return value. The only reason I had `validConfigs` in there was because previously some configs were being dropped. Happy to do it in one go, with the `array_map` approach.

ingeniumed: Also note: I haven't changed the output of this function. It's still giving back the configs, albeit with an enabled field added in there. The inflation of the config is to decide if they are valid or not, and has no bearing on the return value.

ingeniumed: @chriszarate  - would wanna get your take on if full inflation should be dropped in favour or just the classmap lookup. I'm leaning towards keeping it in as the safety of it outweighs the perf impact it could have imo for a list that may not be that long.

alecgeatches: > In regards to the output type being changes, I haven't changed the output type. It's still giving back the configs, albeit with an enabled field added in there.

True! You're right, the instance wasn't actually used here. There's still a benefit that there's no extra instantiation of a class, but much more minor then what I thought.

inc/WpdbStorage/DataSourceCrud.php
```diff
@@ -37,7 +37,19 @@ public static function get_config_by_uuid( string $uuid ): array|WP_Error {
...
 
 	public static function get_configs(): array {
-		return self::get_all_configs();
+		return array_map(
+			function ( array $config ) {
+				// Inflate the config to check if it's valid.
+				$instance = self::inflate_config( $config );
+
+				// If the data source is valid, set a transient field called enabled to true or false otherwise.
+				$config['enabled'] = ! is_wp_error( $instance );
```
chriszarate: I see any easy path to displaying errors:

```suggestion
				$config['errors'] = is_wp_error( $instance ) ? [ $instance ] : [];
```

src/data-sources/DataSourceList.tsx
```diff
@@ -152,8 +149,19 @@ const DataSourceList = () => {
...
+		if ( item?.enabled === false && ! actionsAllowedForDisabledItems.includes( action ) ) {
```
chriszarate:
```suggestion
		if ( item.errors?.length && ! actionsAllowedForDisabledItems.includes( action ) ) {
```

chriszarate (commented):
Suggestion to use `errors` instead of `enabled`, but later PR is ok

inc/WpdbStorage/DataSourceCrud.php
ingeniumed: I had gone back and forth between doing it this way, or having both `enabled` as well as `errors`. I went for the latter, with `errors` to be added in a later PR that also shows it as a notice on the settings screen UI.

I'll change it up in this PR itself.

ingeniumed:
@chriszarate - I have switched enabled to errors, though I have tweaked the way the errors are stored. I realized, by keeping the errors as is what would happen is that you'd have another errors field in the errors. So on the UI, we'd have to access an errors field in every error item just to get to the error messages and error codes. You can see in this screenshot below how it looks like:

CleanShot 2025-04-07 at 12 04 18

By tweaking it to store the errors under the error, it keeps as a key-value array where the key is the code and the value is the error message instead. The screenshot below shows that:

CleanShot 2025-04-07 at 12 04 07

The downside to this is that we'd need to tweak the code to store the additional data field (`errorData`), if we do it down the line.

I like this more than just the `WP_Error` as is, as it makes it easier to consume. Lemme know what you think.

PR #442, opened 2025-04-02
Improve cache debugging and remove footguns

chriszarate:
I originally set out to add some in-memory caching to our existing cache strategy. However, I realized that, since we delegate to `wp_cache_get`, this is done for us. You can confirm by reading `WP_Object_Cache` source code or by monitoring Redis `GET`s to confirm that they are not repeated within the same request lifecycle:

```sh
$ redis-cli monitor | grep -i '"GET" "wp:remote-data-blocks:'
```

Instead, this PR removes some risky configurability and improves logging:

1. Remove `remote_data_blocks_bypass_cache` filter. We have not identified a use case for this filter; caching behavior is separately controllable on queries via `ttl`. This filter can be considered a footgun since it completely disables the ability to cache _any_ request, even in-memory.
   - Currently, since block bindings are repeated for each block in a remote data block, and potentially repeated again for each item in a collection query, we often end up executing the same query many times—sometimes hundreds of times. Disabling caching globally is a bad idea. If it is needed in narrow circumstances, it can be done at the query level.
2. Simplify class instantiation by removing configurability that is unused or unneeded. We can always add filtering in the future, if requested. 
3. Fix non-static instantiation of cache middleware. This doesn't fix any known bugs, but someone who used a custom caching implementation might have been impacted.
4. Improve cache logging with consistent output and additional information like `age` and `ttl`.

Example cache logging:

Screenshot 2025-04-02 at 13 09 27

### DIFF
--- a/docs/ai.md
+++ b/docs/ai.md
@@ -760,14 +760,6 @@ function custom_query_response_metadata( array $metadata, HttpQueryInterface $qu
 add_filter( 'remote_data_blocks_query_response_metadata', 'custom_query_response_metadata', 10, 3 );
 ```
 
-### remote_data_blocks_bypass_cache
-
-Filter to bypass the cache for a specific request (default: `false`).
-
-```php
-add_filter( 'remote_data_blocks_bypass_cache', '__return_true' );
-```
-
 ### remote_data_blocks_http_client_retry_delay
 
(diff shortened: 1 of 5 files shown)


ingeniumed (approved):

ingeniumed:
Minor: looks like we can't setup data sources that may run into CORs issues in WP Playground. I'll look into this

chriszarate:
> CORs issues in WP Playground. I'll look into this

Chat with the team / search the channel, progress is blocked right now.

PR #443, opened 2025-04-03
Add more info about the block for the block bindings

ingeniumed:
### Description

The goal of this PR is to make it easier to spot when a block binding hasn't been mapped to the right remote data. Now, in the side panel under a block binding the remote data block's name will show up so as to spot any errors quickly.

### DIFF
--- a/src/block-editor/filters/withBlockBinding.tsx
+++ b/src/block-editor/filters/withBlockBinding.tsx
@@ -67,7 +67,7 @@ function BoundBlockEdit( props: BoundBlockEditProps ) {
 	return (
 		<>
 			<InspectorControls>
-				<PanelBody title={ sprintf( __( 'Remote Data Block: %s', 'remote-data-blocks' ), remoteDataName ) }>
+				<PanelBody title={ __( sprintf ( 'Remote Data Block: %s', remoteDataName ), 'remote-data-blocks' ) }>
 					<BlockBindingControls
 						attributes={ attributes }
 						availableBindings={ availableBindings }
--- a/src/block-editor/filters/withBlockBinding.tsx
+++ b/src/block-editor/filters/withBlockBinding.tsx
@@ -68,7 +68,7 @@ function BoundBlockEdit( props: BoundBlockEditProps ) {
 		<>
 			<InspectorControls>
 				<PanelBody
-					title={ sprintf( __( 'Remote Data Block: %s' ), remoteDataName ), 'remote-data-blocks' ) }
+					title={ sprintf( __( 'Remote Data Block: %s', 'remote-data-blocks' ), remoteDataName ) }
 				>
 					<BlockBindingControls
 						attributes={ attributes }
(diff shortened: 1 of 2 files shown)


chriszarate (commented):
Fine with the code, but needs product / design review. I am not sure the block name (slug) is the best label here. It's possible that content authors have never even seen a block name.

src/block-editor/filters/withBlockBinding.tsx
```diff
@@ -67,7 +67,9 @@ function BoundBlockEdit( props: BoundBlockEditProps ) {
 	return (
 		<>
 			<InspectorControls>
-				<PanelBody title={ __( 'Remote Data', 'remote-data-blocks' ) }>
+				<PanelBody
+					title={ __( sprintf( 'Remote Data Block: %s', remoteDataName ), 'remote-data-blocks' ) }
```
chriszarate: This string is not really localizable since it includes a highly variable element

alecgeatches: Flipping the order around works, though!

```suggestion
					title={ sprintf( __( 'Remote Data Block: %s' ), remoteDataName ), 'remote-data-blocks' ) }
```

ingeniumed: Whoops, you know what - this was how it was before. I flipped it while going through it as it looked like the wrong order. Good catch

ingeniumed:
> Fine with the code, but needs product / design review. I am not sure the block name (slug) is the best label here. It's possible that content authors have never even seen a block name.

Happy to start that, and leave this PR open. I can add my thoughts there.

src/utils/localized-block-data.ts
```diff
@@ -14,6 +14,14 @@ export function getBlockDataSourceType( blockName?: string ): string {
 	return getBlockConfig( blockName )?.dataSourceType ?? '';
 }
 
+export function getBlockTitle( blockName?: string ): string {
```
ingeniumed, line 26: As discussed, I have to tried to access the block's title from the global config so the prefix `remote-data-blocks` goes away. Lemme know what you think @chriszarate about this.

src/utils/localized-block-data.ts
```diff
@@ -14,6 +14,14 @@ export function getBlockDataSourceType( blockName?: string ): string {
...
+	if ( ! blockName ) {
+		return '';
```
chriszarate, line 28: Minor, but maybe fallback to the block name (with prefix removed) or a string like `unknown`).

chriszarate (approved):

src/utils/localized-block-data.ts
chriszarate: This could happen if the block stopped being registered, for example. So maybe block name is better.

ingeniumed: Good point, happy to do that. It'd keep it consistent with how it was before, wherein the blockName was used.

ingeniumed: I've set it so that, if the title is not found in the config then the block name is used (with the prefix removed). If no block name is provided, then its an empty string which is consistent with the other methods (from what I can tell).

PR #445, opened 2025-04-04
Port over the performance fixes for the plugin

ingeniumed:
### Description

This is meant to port over the performance fixes that were prototyped, to the actual plugin. I've skipped some of the fixes in order to actually focus on the ones that matter in the plugin. I've added in comments to the relevant lines so as to explain them better.

Since I've gotten rid of the more risky ones, what's left is a lot easier to test and safer imo. The best way to test it would be to create a page with a lot of block bindings on them. I created a single block binding with 20 products under it, and then duplicated it 4 times. That way I could see the caching in action, etc.

### DIFF
--- a/inc/Config/QueryRunner/QueryRunner.php
+++ b/inc/Config/QueryRunner/QueryRunner.php
@@ -40,15 +40,15 @@ class QueryRunner implements QueryRunnerInterface {
 			return $headers;
 		}
 
-		$method = $query->get_request_method();
-		$body = $query->get_request_body( $input_variables );
 		$endpoint = $query->get_endpoint( $input_variables );
-		$cache_ttl = $query->get_cache_ttl( $input_variables );
 		$parsed_url = wp_parse_url( $endpoint );
 
 		if ( false === $parsed_url ) {
 			return new WP_Error( 'Unable to parse endpoint URL' );
 		}
+
+		$scheme = $parsed_url['scheme'] ?? '';
+		$host = $parsed_url['host'] ?? '';
 
 		/**
 		 * Filters the allowed URL schemes for this request.
@@ -59,23 +59,26 @@ class QueryRunner implements QueryRunnerInterface {
 		 */
 		$allowed_url_schemes = apply_filters( 'remote_data_blocks_allowed_url_schemes', [ 'https' ], $query );
 
-		if ( empty( $parsed_url['scheme'] ?? '' ) || ! in_array( $parsed_url['scheme'], $allowed_url_schemes, true ) ) {
+		if ( empty( $scheme ) || ! in_array( $scheme, $allowed_url_schemes, true ) ) {
 			return new WP_Error( 'Invalid endpoint URL scheme' );
 		}
 
-		if ( empty( $parsed_url['host'] ?? '' ) ) {
+		if ( empty( $host ) ) {
 			return new WP_Error( 'Invalid endpoint URL host' );
 		}
 
-		$scheme = $parsed_url['scheme'];
-		$host = $parsed_url['host'];
-		$user = $parsed_url['user'] ?? '';
-		$path = $parsed_url['path'] ?? '';
-
+		$method = $query->get_request_method();
+		$body = $query->get_request_body( $input_variables );
+		$cache_ttl = $query->get_cache_ttl( $input_variables );
 		$query = ! empty( $parsed_url['query'] ?? '' ) ? '?' . $parsed_url['query'] : '';
 		$port = ! empty( $parsed_url['port'] ?? '' ) ? ':' . $parsed_url['port'] : '';
-		$pass = ! empty( $parsed_url['pass'] ?? '' ) ? ':' . $parsed_url['pass'] : '';
-		$pass = ( $user || $pass ) ? $pass . '@' : '';
+		$path = $parsed_url['path'] ?? '';
+		$user = $parsed_url['user'] ?? '';
+		$pass = $parsed_url['pass'] ?? '';
+		$auth = '';
+		if ( $user || $pass ) {
+			$auth = $user . ( $pass ? ':' . $pass : '' ) . '@';
+		}
 
 		$request_details = [
 			'method' => $method,
@@ -83,7 +86,7 @@ class QueryRunner implements QueryRunnerInterface {
 				RequestOptions::HEADERS => $headers,
 				RequestOptions::JSON => $body,
 			],
-			'origin' => sprintf( '%s://%s%s%s%s', $scheme, $user, $pass, $host, $port ),
+			'origin' => $scheme . '://' . $auth . $host . $port,
 			'ttl' => $cache_ttl,
 			'uri' => sprintf( '%s%s', $path, $query ),
 		];
@@ -166,6 +169,7 @@ class QueryRunner implements QueryRunnerInterface {
 	protected function get_response_metadata( HttpQueryInterface $query, array $response_metadata, array $query_results ): array {
 		$age = intval( $response_metadata['age'] ?? 0 );
 		$time = time() - $age;
+		$count = count( $query_results );
 
 		$query_response_metadata = [
 			'last_updated' => [
@@ -176,7 +180,7 @@ class QueryRunner implements QueryRunnerInterface {
 			'total_count' => [
 				'name' => 'Total count',
 				'type' => 'integer',
-				'value' => count( $query_results ),
+				'value' => $count,
 			],
 		];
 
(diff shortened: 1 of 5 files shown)


inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -40,16 +40,16 @@ protected function get_request_details( HttpQueryInterface $query, array $input_
 			return $headers;
 		}
 
-		$method = $query->get_request_method();
```
ingeniumed, line 43: This just moves things around to exit early, and only do url construction work after those initial checks pass.

inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -166,6 +169,7 @@ protected function get_raw_response_data( HttpQueryInterface $query, array $inpu
 	protected function get_response_metadata( HttpQueryInterface $query, array $response_metadata, array $query_results ): array {
 		$age = intval( $response_metadata['age'] ?? 0 );
 		$time = time() - $age;
+		$count = count( $query_results );
```
ingeniumed, line 172: This was a minor change I had made. It's impact isn't that much, but every bit counts.

inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -204,13 +208,12 @@ public function execute( HttpQueryInterface $query, array $input_variables ): ar
...
 		// Set default input variables.
 		foreach ( $input_schema as $key => $schema ) {
-			if ( ! array_key_exists( $key, $input_variables ) && isset( $schema['default_value'] ) ) {
-				$input_variables[ $key ] = $schema['default_value'];
-			}
-
-			// If the input variable is required and not provided, return an error.
-			if ( ! array_key_exists( $key, $input_variables ) && isset( $schema['required'] ) && $schema['required'] ) {
-				return new WP_Error( 'remote-data-blocks-missing-required-input-variable', sprintf( 'Missing required input variable: %s', $key ) );
+			if ( ! array_key_exists( $key, $input_variables ) ) {
```
ingeniumed, line 211: This is meant to simplify this if check.

inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -272,17 +275,10 @@ public function execute_batch( HttpQueryInterface $query, array $array_of_input_
...
-				function ( array $carry, mixed $item ): array {
-					if ( is_array( $item ) ) {
-						return array_merge( $carry, $item );
-					}
-
-					return array_merge( $carry, [ $item ] );
-				},
-				[]
-			);
+			$ids = [];
```
ingeniumed: Replacing the `array_reduce` had a minor impact on the execution of the queries, hence why I kept this in.

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -223,7 +225,24 @@ public static function get_value( array $source_args, WP_Block|array $block, str
 			return $fallback_content;
 		}
 
-		$query_response = self::execute_queries( $block_context, $source_args, $field_name );
+		// Generate cache key based on context and source args that affect the query
```
ingeniumed: This targets the case where the block with the same set of queries is duplicated. It sped up the single request time by around 10-20% depending on how much had to be loaded.

inc/HttpClient/RdbCacheStrategy.php
```diff
@@ -14,7 +14,7 @@
 
 class RdbCacheStrategy extends GreedyCacheStrategy {
 	private const CACHE_INVALIDATING_REQUEST_HEADERS = [ 'Authorization', 'Cache-Control' ];
-	private const FALLBACK_CACHE_TTL_IN_SECONDS = 60;
+	private const FALLBACK_CACHE_TTL_IN_SECONDS = 120;
```
ingeniumed: Originally, I had this at 10 mins but that felt like a lot. I've made this 2 mins as that accounted for the case where a query could take quite a while to come back, and we may have several of em. It helps subsequent requests. 

Happy to revert this, just found that on a complicated site this helped to delay when subsequent requests could take a while to load due to the cache expiring.

inc/Validation/Types.php
```diff
@@ -177,20 +177,21 @@ public static function object( array $properties ): array {
 	}
 
 	public static function one_of( array ...$member_types ): array {
+		$allowed_non_primitive_member_types = [
```
ingeniumed, line 180: Moving the check outside the loop helped to reduce the time just a touch.

inc/Validation/Types.php
```diff
@@ -315,11 +316,15 @@ public static function is_serializable( array $type ): bool {
 	}
 
 	private static function is_type( string $type_name, array ...$types_to_check ): bool {
-		return array_reduce( $types_to_check, function ( bool $carry, array $type ) use ( $type_name ): bool {
+		foreach ( $types_to_check as $type ) {
```
ingeniumed, line 319: Same modification as what was done above.  It has a minor impact but every bit matters.

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: Noticed a bug in this that the tests caught. I'm trying to fix it. If I'm not able to, then I'm going to remove it from this PR.

inc/Config/QueryRunner/QueryRunner.php
alecgeatches: I'm not sure how this has a performance impact, but it won't make it worse.

inc/Config/QueryRunner/QueryRunner.php
alecgeatches: I'm surprised this had a performance impact, how could you tell? It seems like this is only used in specific circumstances where `id:list` is available on a query, and there are multiple queries on the page, and even then the code looks equivalent.

Again, I don't think this hurts, and I think the imperative style looks cleaner here. But I struggle to see how performance is changed here.

inc/Editor/DataBinding/BlockBindings.php
alecgeatches: I'll wait to review/test this until you've figured out that bug. @chriszarate You mentioned in a prior PR https://github.com/Automattic/remote-data-blocks/pull/442 that `wp_cache_get()` handles some of the memoizing internally. Do you know if in-memory caching here will definitely or definitely not have an impact?

inc/HttpClient/RdbCacheStrategy.php
alecgeatches: I think any default in the 1-5 minute range is fine, but curious what @chriszarate has to say here.

inc/Validation/Types.php
alecgeatches: This one makes sense to me, we get the benefit of an early return here.

inc/Editor/DataBinding/BlockBindings.php
chriszarate: It will not have an impact, except in the rare circumstances where caching is disabled at the query level (i.e., when the query sets `TTL=-1`). We probably want to address this, but we should do so in the caching middleware, not here in the calling code. I have a PR that does this using a Guzzle utility class. However, I don't think it's a big priority; disabling caching is probably more of a debugging option than a realistic production scenario.

inc/HttpClient/RdbCacheStrategy.php
chriszarate: > I've made this 2 mins as that accounted for the case where a query could take quite a while to come back, and we may have several of em

I'm not sure I understand this statement. Certainly increasing the default TTL will help make cache misses less frequent (for queries that use the default TTL). Is there another effect I'm missing?

chriszarate:
We would strongly benefit from atomic PRs here, with one improvement at a time. This would help us observe the benefit, rank the impact, have a better historical record, and have clear examples that we could apply to other code.

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: In trying to figure out this bug, I have come to the same conclusion. This isn't worth doing, so I'm going to revert it.

inc/HttpClient/RdbCacheStrategy.php
ingeniumed: Nope, that's really it. Given we were able to make other optimizations internally, that led us to significantly improving the overall query time I'm going to revert this. We can revisit it if need be.

inc/Config/QueryRunner/QueryRunner.php
ingeniumed: I collected the timings of the minor impact changes together, and used that as the basis for keeping em tbh. I dropped a lot of the other minor ones as they weren't worth doing. This one I kept, as you said, it makes it cleaner.

That said, I'm going to revert it. I'm just going over all the other performance improvements and dropping em if it's not really needed.

ingeniumed:
> We would strongly benefit from atomic PRs here, with one improvement at a time. This would help us observe the benefit, rank the impact, have a better historical record, and have clear examples that we could apply to other code.

Tbh, the biggest reason I didn't do that was because I wasn't 100% certain if some of the high impact improvements might go in. I didn't want to open up 5-6 PRs (some of which might be very small) and then close them all down as they weren't worth doing. That's why I kept this in draft status, so I could discuss this and only once we were 100% certain on which improvements to keep I'd see if the PR had to be split further.

After the comments/discussions above, I've reverted a good chunk of the performance improvements and what's left is a lot smaller now. It focuses really on two areas:

- Simple readability change
- Early returns

They should be a lot easier to review and their impact will be minor. Lemme know what you think.

inc/Config/QueryRunner/QueryRunner.php
chriszarate: The comment was deleted here, can you restore / improve it?

inc/Validation/Types.php
chriszarate: Ok, but this function is only called in two places. Once the array has a single member, the other it has two members

chriszarate:
I'm having trouble seeing the impact, given the low call counts for each of the improvement sites. But I would love to learn something if these are measurable improvements. How did you benchmark? Can you share if the results are significant?

ingeniumed:
The results are insignificant tbh, there's no measurable difference. The improvements were significant when all the performance improvements were present. Since I've taken 90% of em out, now it's not significant.

I'm going to close this PR off, as it's not worth it to get in in hind sight.

PR #449, opened 2025-04-08
Front-end pagination support

chriszarate:
This PR adds a pagination helper block that enables pagination on the front-end (render). It loosely follows the pattern of the Core query pagination block, but fully encapsulates the pagination links. This pagination block is automatically inserted whenever a collection that supports pagination is inserted.

The biggest change is moving logic from the `usePaginationVariables` hook to PHP, so that both server-side and client-side code can reuse the logic. This logic is encapsulated in the new `Pagination` class and consumed by `QueryRunner` and `BlockBindings`.

Pagination links are implemented via a single query variable (`rdb-pagination`) that envelopes all possible pagination parameters (cursor, offset, page, etc.) and encodes them in base64. Example: `?rdb-pagination=eyI3MDkzOWFmYWZmNDE4MWYwODI5MTUzMTk4Nzg1ODE0ZSI6eyJwYWdlIjozfX0`.

For now, we will only support navigating a single block's pagination thread at any time, but the implementation leaves room to support multiple threads in the future, should that surface as an important use case.

### Automatic pagination block insertion

https://github.com/user-attachments/assets/3c286105-a977-468f-a857-069bdf323fb6

### Front-end pagination

https://github.com/user-attachments/assets/a083e277-3dfb-45bd-bae5-49947807fa02

### DIFF
--- a/inc/Config/ArraySerializable.php
+++ b/inc/Config/ArraySerializable.php
@@ -13,7 +13,15 @@ defined( 'ABSPATH' ) || exit();
  * ArraySerializable class
  */
 abstract class ArraySerializable implements ArraySerializableInterface {
-	final private function __construct( protected array $config ) {}
+	protected string $id;
+
+	final private function __construct( protected array $config ) {
+		$this->id = md5( wp_json_encode( [ get_called_class(), $config ] ) );
+	}
+
+	final public function get_id(): string {
+		return $this->id;
+	}
 
 	protected function get_or_call_from_config( string $property_name, mixed ...$callable_args ): mixed {
 		$config_value = $this->config[ $property_name ] ?? null;
--- a/inc/Editor/DataBinding/Pagination.php
+++ b/inc/Editor/DataBinding/Pagination.php
@@ -11,10 +11,20 @@ use function get_query_var;
 defined( 'ABSPATH' ) || exit();
 
 class Pagination {
-	private static $variable_name = 'rdb-pagination';
+	private static string $variable_name = 'rdb-pagination';
 
 	public static function init(): void {
 		add_filter( 'query_vars', [ __CLASS__, 'register_query_var' ], 10, 1 );
+
+		/**
+		 * Filter the name of the query variable used for pagination.
+		 *
+		 * @param string $query_var_name The name of the query variable.
+		 */
+		self::$variable_name = apply_filters(
+			'remote_data_blocks_pagination_query_var_name',
+			'rdb-pagination'
+		);
 	}
 
 	public static function create_query_var( string $query_id, array $pagination_input_variables ): string {
--- a/inc/Editor/DataBinding/Pagination.php
+++ b/inc/Editor/DataBinding/Pagination.php
@@ -15,9 +15,9 @@ class Pagination {
 		add_filter( 'query_vars', [ __CLASS__, 'register_query_var' ], 10, 1 );
 	}
 
-	public static function create_query_var( string $block_id, array $pagination_input_variables ): string {
+	public static function create_query_var( string $query_id, array $pagination_input_variables ): string {
 		$value = [
-			$block_id => $pagination_input_variables,
+			$query_id => $pagination_input_variables,
 		];
 
 		return add_query_arg( self::$variable_name, self::encode_query_var( $value ) );
@@ -31,7 +31,7 @@ class Pagination {
 		return base64_encode( wp_json_encode( $query_var_value ) );
 	}
 
-	public static function get_pagination_input_variables_for_current_request( string $block_id ): array {
+	public static function get_pagination_input_variables_for_current_request( string $query_id ): array {
 		$value = self::decode_query_var( get_query_var( self::$variable_name, '' ) );
 
 		if ( empty( $value ) ) {
@@ -44,7 +44,7 @@ class Pagination {
 		// We only expect a single key => value pair, but in the future we may
 		// decide to support more than one. This would allow us to control the
 		// pagination of multiple remote data blocks independently.
-		return $value[ $block_id ] ?? [];
+		return $value[ $query_id ] ?? [];
 	}
 
 	public static function format_pagination_data_for_query_response( array|null $pagination_data, array $query_input_schema, array $input_variables ): array {
--- a/inc/Editor/DataBinding/Pagination.php
+++ b/inc/Editor/DataBinding/Pagination.php
@@ -1,0 +1,171 @@
+<?php declare(strict_types = 1);
+
+namespace RemoteDataBlocks\Editor\DataBinding;
+
+use function add_filter;
+use function add_query_arg;
+use function get_query_var;
+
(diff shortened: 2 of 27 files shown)


smithjw1:
I've had a look this morning and was able to do some shenanigans with input variables. These are bad-faith user inputs.

https://github.com/user-attachments/assets/6f03f606-8255-4df0-8471-7a2186755410

I have other thoughts about labels/UX that I can return to later in the day.

@chriszarate, could you also stub out some documentation on this? Looking through the code, I think I know how it works, but it would be helpful for my review if we had a bit of documentation for me to check against. I'm also fine hearing that it's too early for that.

chriszarate:
> I've had a look this morning and was able to do some shenanigans with input variables. These are bad-faith user inputs.

I've opened an issue for this (VIPCMS-357); these inputs are generic and provided for input variables of any type (not just pagination input variables). This PR does not intersect with these inputs at all.

chriszarate:
> could you also stub out some documentation on this? Looking through the code, I think I know how it works, but it would be helpful for my review if we had a bit of documentation for me to check against. I'm also fine hearing that it's too early for that.

I've updated the existing documentation for how pagination variables are defined and they are complete now. We don't have any documentation for the editor experience or the front-end render experience, so I haven't attempted any scoped to pagination. But I can if you think they should exist standalone.

smithjw1:
> We don't have any documentation for the editor experience or the front-end render experience, so I haven't attempted any scoped to pagination. But I can if you think they should exist standalone.

Great! I didn't bother to check the current documentation to see what was there, so I missed that we already had it. Sorry about that.

I opened a Draft PR with some suggestions, but I'll add two more bigger questions. I think both of these are unrelated to this PR, but input variables in general, but I wanted to start here, and then I can create issues for them separately.

## Hidden input variables

@stoph and I were taking a look at this and weren't sure when a customer would want to set a starting page other than "1" when setting up a loop block. 

I thought about this some more and can imagine situations where developers don't want to allow editorial users to adjust input variables beyond what's defined in code, e.g. when the API has a maximum number that can be returned.

Do we think adding a `hidden` boolean is useful? Our examples can express smart defaults, but we can leave the implementation up to customers.

## Input variable names in the sidebar

Screenshot 2025-04-09 at 14 00 43

We use the key as the label for the input, not the name defined in the config. Could we switch that to the name when it exists?

inc/Config/ArraySerializable.php
```diff
@@ -13,7 +13,15 @@
  * ArraySerializable class
  */
 abstract class ArraySerializable implements ArraySerializableInterface {
-	final private function __construct( protected array $config ) {}
+	protected string $id;
+
+	final private function __construct( protected array $config ) {
+		$this->id = md5( wp_json_encode( [ get_called_class(), $config ] ) );
```
ingeniumed, line 19: One of the thoughts I had was to see if we could leverage the UUID from the config instead of doing it this way. But, I see that the problem is this isn't always available like for code based sources (Art Institute being an example of this). 

Is there any benefit in changing that, so that all data source configs do have a UUID assigned to them? Then, we could use that UUID here and wherever else we might need to in the future?

inc/Editor/DataBinding/Pagination.php
```diff
@@ -0,0 +1,212 @@
...
+		$pagination_type = 'NONE';
+		$total_items = $pagination_data['total_items'] ?? null;
+
+		$pagination_input_variable_targets = [
+			'offset' => $offset_variable,
+			'page' => $page_variable,
+			'per_page' => $per_page_variable,
+		];
+
+		if ( $cursor_variable ) {
```
ingeniumed, line 150: Could this logic be documented just above this if statement, or as a function comment?

inc/Config/ArraySerializable.php
ingeniumed: All good if the answer is no, or that it be punted to a future PR. This is just a thought that occurred to me when I tried to experiment with changing it to the UUID, and not seeing it reflect since the Art Institute didn't have one.

ingeniumed (approved):
Neither of my comments block the merge of this PR. It looks good to me!

inc/Config/ArraySerializable.php
chriszarate: > Is there any benefit in changing that, so that all data source configs do have a UUID assigned to them? Then, we could use that UUID here and wherever else we might need to in the future?

It's a good question. We've discussed both sides. The major downsides of universally requiring UUIDs for all entities are:

1. You have to delegate UUID creation to the user, and they can mess it up. We cannot create stable UUIDs on their behalf without persistence (which code-generated entities don't have). We can validate that a string _looks_ like a UUID, but we can't really validate that it is unique or stable.
2. There isn't really much benefit to the user that justifies making this a requirement. Code-generated UUIDs are not persisted and there is no central store, so there is no concept of "get by UUID." Even if there was, users have already created an instance of their entity and can freely pass it around in their code—which is probably easier and clearer than passing around a UUID string.

Another downside of UUIDs in this use case is that they have no functional value. They are just identifiers. The ID introduced in this PR, by contrast, is deterministic based on the inputs that influence its behavior. This is useful since we are using it to identify external attributes that can be applied to this entity (pagination variables). If the input changes, the ID changes, and it makes sense that the attributes might no longer apply. Meanwhile, a UUID might change for any number of reasons that have nothing to do with the entity's behavior.

chriszarate: All that said, the main reason that this ID is useful at all is because block instances don't have stable identifiers. If they did, we would use them.

chriszarate:
> I thought about this some more and can imagine situations where developers don't want to allow editorial users to adjust input variables beyond what's defined in code, e.g. when the API has a maximum number that can be returned.
>
> Do we think adding a hidden boolean is useful? Our examples can express smart defaults, but we can leave the implementation up to customers.

@smithjw1 I see the utility, but it feels like injecting the default value (done in your PR) also solves for this in a limited way. 

My instinct is that the issue you're describing is much more complex and speaks to a desire for developers to have more control over the selection UI—and the ensuing validation and user feedback. I don't think adding configuration toggles is a scalable approach for this; we need to provide some hooks for developers to inject their own UI.

> We use the key as the label for the input, not the name defined in the config. Could we switch that to the name when it exists?

Yes, it's a bit more invasive of a change and unrelated to pagination, so I'll tackle it in a separate PR and pull in your improvements as well.

inc/Editor/DataBinding/Pagination.php
chriszarate: Done in https://github.com/Automattic/remote-data-blocks/pull/449/commits/1465512f7006658a78925c02cf23b5a99e51b1d9

chriszarate:
Added a filter for the variable name in https://github.com/Automattic/remote-data-blocks/pull/449/commits/0acd7721976ac3101574a27c2359e4a0b2993c0a

PR #452, opened 2025-04-08
Use new Telemetry class from VIP mu-plugins

chriszarate:
Use the new `Telemetry` class from VIP mu-plugins. This class abstracts away many of the checks that were being performed in this plugin, and as a result we can remove a large amount of code. 

I also renamed `TracksTelemetry` to `Telemetry` since it is no longer pinned to that service. Note that much of the code, including `track_remote_data_blocks_usage` is unchanged.

I've also added an integration test to cover this work.


inc/Telemetry/Telemetry.php
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	private function setup_tracking_via_hooks(): void {
+		// WordPress hooks.
+		add_action( 'activated_plugin', [ $this, 'track_plugin_activation' ], 10, 1 );
+		add_action( 'deactivated_plugin', [ $this, 'track_plugin_deactivation' ], 10, 1 );
+		add_action( 'save_post', [ $this, 'track_remote_data_blocks_usage' ], 10, 2 );
+
+		// Custom hook to allow other plugin code to track events
+		add_action( 'remote_data_blocks_track_event', [ $this, 'record_event' ], 10, 2 );
```
maxschmeling, line 59: We're inconsistent between the action and the `EVENT_PREFIX` value (`remote_data_blocks` vs `remotedatablocks`)

Any reason we can't make those consistent?

inc/Telemetry/Telemetry.php
```diff
@@ -0,0 +1,153 @@
...
+			$key = $data_source_type . '_data_source_count';
+			$track_props[ $key ] = ( $track_props[ $key ] ?? 0 ) + 1;
+			$track_props['remote_data_blocks_total_count'] = ( $track_props['remote_data_blocks_total_count'] ?? 0 ) + 1;
+		}
+
+		$this->record_event( 'blocks_usage_stats', $track_props );
+	}
+
+	/**
+	 * Record an event with Tracks.
```
maxschmeling:
```suggestion
	 * Record a telemetry event.
```

inc/Telemetry/Telemetry.php
```diff
@@ -0,0 +1,153 @@
...
+	 *
+	 * @param string $event_name The name of the event.
+	 * @param array  $props      The properties to send with the event.
+	 *
+	 * @return bool True if the event was recorded, false otherwise.
```
maxschmeling: No return value now

maxschmeling (approved):

inc/Telemetry/Telemetry.php
chriszarate: Most of the namespacing we've done in the WordPress plugin uses `remote_data_blocks`. However, Tracks is external to WordPress and the event names were already registered months ago using `remotedatablocks` as the prefix. It's a bit painful to change, otherwise I would have done it.

inc/Telemetry/Telemetry.php
chriszarate:
```suggestion
```

chriszarate: Done in https://github.com/Automattic/remote-data-blocks/pull/452/commits/9a721d39f4260dd2a36341e3a8dacb62a7530717

PR #453, opened 2025-04-09
Jacob's Review

smithjw1:
I've added a few pieces of code to provide specifics to my review. There are some things I didn't want to attempt and will comment on the original PR about.

Feel free to push back on any of these and improve them.

### DIFF
--- a/inc/Editor/BlockManagement/ConfigRegistry.php
+++ b/inc/Editor/BlockManagement/ConfigRegistry.php
@@ -82,6 +82,7 @@ class ConfigRegistry {
 							'required' => $input_var['required'] ?? true,
 							'slug' => $slug,
 							'type' => $input_var['type'] ?? 'string',
+							'default_value' => $input_var['default_value'] ?? null,
 						];
 					}, array_keys( $input_schema ), array_values( $input_schema ) ),
 					'name' => $has_required_variables ? 'Manual input' : ( $is_collection ? 'Load collection' : 'Load item' ),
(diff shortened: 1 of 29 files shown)


inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -82,6 +82,7 @@ public static function register_block( array $user_config = [] ): bool|WP_Error
 							'required' => $input_var['required'] ?? true,
 							'slug' => $slug,
 							'type' => $input_var['type'] ?? 'string',
+							'default_value' => $input_var['default_value'] ?? null,
```
smithjw1, line 85: This is the first change to bring the default value into the modal that pops up when you customize the inputs.

Since these values will be used if you enter nothing, I wanted them to appear.

inc/Editor/DataBinding/Pagination.php
```diff
@@ -209,4 +209,13 @@ public static function register_query_var( array $query_vars ): array {
...
 		return $query_vars;
 	}
+
+	/**
+	 * Get the query variable name used for pagination.
+	 * 
+	 * @return string The query variable name
+	 */
+	public static function get_query_var_name(): string {
+		return self::$variable_name;
```
smithjw1, line 219: I'm not wild about the query var name, specifically including `rdb` but I understand the desire to namespace it.

I'd be interested to hear other's thoughts about something more generic like "pagination"

What would it take to make this configurable via filter?

chriszarate: `apply_filters( 'remote_data_blocks_pagination_query_var_name', self::$variable_name );`

chriszarate: Done in original PR

PR #455, opened 2025-04-11
Use a label to highlight the connected block

ingeniumed:
### Description

This PR is meant to incorporate the feedback from @jarekmorawski, to build on #443. It does the following:

- Revert the title back to `Remote Data`
- Add in a paragraph that mirrors the Gutenberg helper text to highlight what the underlying block binding is.

CleanShot 2025-04-10 at 11 49 44 (1)

### DIFF
--- a/src/block-editor/filters/withBlockBinding.tsx
+++ b/src/block-editor/filters/withBlockBinding.tsx
@@ -76,7 +76,7 @@ function BoundBlockEdit( props: BoundBlockEditProps ) {
 		<>
 			<InspectorControls>
 				<PanelBody title={ __( 'Remote Data', 'remote-data-blocks' ) }>
-					<p style={ { fontSize: 'small', color: 'rgb(117, 117, 117)' } }>
+					<p className="rdb-block-helper-text">
 						{ sprintf( __( 'Connected to %s', 'remote-data-blocks' ), remoteDataTitle ) }
 					</p>
 					<BlockBindingControls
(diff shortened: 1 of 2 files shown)


src/block-editor/filters/withBlockBinding.tsx
```diff
@@ -75,9 +75,10 @@ function BoundBlockEdit( props: BoundBlockEditProps ) {
 	return (
 		<>
 			<InspectorControls>
-				<PanelBody
-					title={ sprintf( __( 'Remote Data Block: %s', 'remote-data-blocks' ), remoteDataTitle ) }
-				>
+				<PanelBody title={ __( 'Remote Data', 'remote-data-blocks' ) }>
+					<p style={ { fontSize: 'small', color: 'rgb(117, 117, 117)' } }>
```
ingeniumed: The fontSize and colour are used to replicate the Gutenberg helper text shown on a lot of their components. Unfortunately, `PanelBody` doesn't have `help` as an option on it so this is the only way to show it.

I have set the fontSize to be small, and set the colour to be the same within the helper text. If anyone has a better suggestion, I'm happy to update this.

chriszarate: Can this be targeted with a className instead of inline style?

chriszarate (approved):
One question, otherwise 👍

src/block-editor/filters/withBlockBinding.tsx
ingeniumed: Done in the latest commit 👍🏾

ingeniumed:
CleanShot 2025-04-14 at 10 02 06

PR #462, opened 2025-04-15
ignore node_modules for webpack watch

shekharnwagh:
The command to start local server was failing with EMFILE errors. Ignoring node_modules and increasing the aggregateTimeout in watchOptions fixed the issue for now. The files in node_modules won't be changed during development without any dependencies changing, so it's safe to ignore them for watching.

### DIFF
--- a/webpack.utils.js
+++ b/webpack.utils.js
@@ -7,7 +7,7 @@ const path = require( 'path' );
 // This function modernizes the configuration object to support TypeScript. It
 // also allows for additional scripts to be added to the entry point. Blocks are
 // included by default, so this is only needed for non-block scripts.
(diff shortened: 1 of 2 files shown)


webpack.utils.js
```diff
@@ -37,6 +37,7 @@ function modernize( config, additionalScripts = {}, additionalPlugins = [] ) {
 				'@': path.resolve( __dirname, 'src/' ),
 			},
 		},
+		watchOptions: Object.keys( watchOptions ).length ? watchOptions : config.watchOptions,
```
chriszarate: merge instead of override, consistent with how the other params work
```suggestion
		watchOptions: { ...config.watchOptions, ...watchOptions },
```

chriszarate (approved):

PR #464, opened 2025-04-15
Adding authentication specific help text.

smithjw1:
The generic help text left me with questions, so I added a constant to contain help text appropriate for each authenticating type.

Before:
before

After:
Screenshot 2025-04-15 at 13 59 52

### DIFF
--- a/src/data-sources/components/HttpAuthSettingsInput.tsx
+++ b/src/data-sources/components/HttpAuthSettingsInput.tsx
@@ -2,10 +2,12 @@ import { SelectControl, TextControl } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { ChangeEvent } from 'react';
 
+import { HttpAuthTypes } from '../http/types';
 import PasswordInputControl from '@/data-sources/components/PasswordInputControl';
 import {
 	HTTP_SOURCE_AUTH_TYPE_SELECT_OPTIONS,
 	HTTP_SOURCE_ADD_TO_SELECT_OPTIONS,
+	HTTP_SOURCE_AUTH_HELP_TEXT,
 } from '@/data-sources/constants';
 import { HttpConfig } from '@/data-sources/types';
 
@@ -77,10 +79,7 @@ export const HttpAuthSettingsInput: React.FC< HttpAuthSettingsInputProps > = ( {
 					value={ auth?.value ?? '' }
 					onChange={ value => onChange( 'value', value ) }
 					__next40pxDefaultSize
-					help={ __(
-						'The authentication value to use for the HTTP endpoint. When using Basic Auth, this is "username:password" string.',
-						'remote-data-blocks'
-					) }
+					help={ HTTP_SOURCE_AUTH_HELP_TEXT[ ( auth?.type ?? 'none' ) as HttpAuthTypes ] }
 				/>
 			) }
 		</>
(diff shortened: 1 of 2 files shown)


src/data-sources/components/HttpAuthSettingsInput.tsx
```diff
@@ -2,10 +2,12 @@ import { SelectControl, TextControl } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { ChangeEvent } from 'react';
 
+import { HttpAuthTypes } from '../http/types';
```
smithjw1: I'm not sure about this and the type import on line 10, but this was necessary to get Typescript to stop complaining.

src/data-sources/constants.ts
```diff
@@ -44,6 +44,13 @@ export const HTTP_SOURCE_ADD_TO_SELECT_OPTIONS: SelectOption< HttpApiKeyDestinat
 	{ label: __( 'Query Params', 'remote-data-blocks' ), value: 'queryparams' },
 ];
 
+export const HTTP_SOURCE_AUTH_HELP_TEXT: Record< HttpAuthTypes, string > = {
+	none: __( 'No authentication is required.', 'remote-data-blocks' ),
+	bearer: __( 'A string like: "Bearer {token}".', 'remote-data-blocks' ),
```
smithjw1, line 49: AI says this is true. If we auto-add "Bearer " it couldn't figure it out, and I couldn't either.

shekharnwagh: I may not be understanding it, but if the type is bearer, won't it be obvious we'll be using the prefix `Bearer ` for the auth header ? In Postman when you select the type as bearer token, you only need to enter the `{token}` and Postman passes it as `Bearer {token}` in header automatically.

src/data-sources/components/HttpAuthSettingsInput.tsx
shekharnwagh: I tried removing it from here and [place where its used](https://github.com/Automattic/remote-data-blocks/pull/464/files#diff-5a6b6f4b66b1008a168c76a6cb88b6586db40da9de3c125cbcb4dc0d1f98e1aaR81) in this file and TS did not throw any error. 

I've pushed a commit 4e7f944 to remove that, but feel free to revert that if you are able to reproduce it.

src/data-sources/constants.ts
smithjw1: It isn't obvious to me (or AI when I ask). I can't figure out how to test it to know for sure.

shekharnwagh: Unlike other source types, HTTP type hasn't been fleshed out yet, so we don't have code which consumes this config and uses auth yet. We have the [zip code api example](https://github.com/Automattic/remote-data-blocks/tree/trunk/example/rest-api/zip-code), but that doesn't use auth. My understanding is that when users add something in code to use this config, they have some check on if `auth.type = bearer then header.authorization = Bearer {auth.value}`. Users could also use it the way the message says here. Users might use it the way we hint it in the docs. So if think the value should include `Bearer` string we can add something about that [here](https://github.com/Automattic/remote-data-blocks/blob/trunk/docs/tutorials/http.md)

smithjw1: Ok I get it now. There is no magic the plugin does, so these are just different config variables that get stored and the end-user has to figure out exactly how to use them.

I'm going to close this PR and come at this problem a different way.

PR #465, opened 2025-04-16
Custom query monitor panel, improve validation issue reporting

chriszarate:
This PR introduces a custom Query Monitor panel for Remote Data Blocks, with accompanying improvements to how we report validation errors.

Currently, we pipe these errors to the default Logs panel. We include more information than is helpful and it is very difficult to parse and read. It looks like this:

<p>Screenshot 2025-04-16 at 16 17 32</p>

With this PR, we construct a descriptive path using PHP syntax that more directly points to the issue. We also withhold context that is of little value in favor of a more direct stack trace:

<p>Screenshot 2025-04-16 at 16 24 43</p>

These issues are also reported in their own dedicated panel. In future PRs, we can add additional panels dedicated to other kinds of issues or logs (query executions, caching, etc).


smithjw1:
This is awesome! I didn't know we had this logging and it would have saved me some time.

I don't see us mention query monitor in the docs. Should we add something to the Local Development doc?  And we could be me if folks want to move on to other development tasks.

chriszarate:
> I don't see us mention query monitor in the docs.

We mention it here, but could definitely cross-link and amplify.

https://github.com/Automattic/remote-data-blocks/blob/trunk/docs/troubleshooting.md#query-monitor

We also call it out in an admin notice when we detect issues:

https://github.com/Automattic/remote-data-blocks/blob/5c1cd08c3ce8d381d19a5e74a65e4e5ef1d9da7e/inc/Editor/AdminNotices/AdminNotices.php#L65

smithjw1:
> We mention it here, but could definitely cross-link and amplify.

That will teach me to leave case-sensitive search enabled.

inc/Validation/Validator.php
```diff
@@ -40,14 +33,14 @@ public function validate( mixed $data ): bool|WP_Error {
 	 * @param mixed $value The value to validate.
```
ingeniumed, line 34: Should add `$path` here as a param

inc/Logging/QueryMonitor/QueryMonitor.php
```diff
@@ -0,0 +1,52 @@
...
+
+class QueryMonitor {
+	public static function init(): void {
+		add_filter( 'qm/collectors', [ __CLASS__, 'add_collectors' ], 90, 1 );
+		add_filter( 'qm/outputter/html', [ __CLASS__, 'add_outputters' ], 90, 1 );
+	}
+
+	public static function add_collectors( array $collectors ): array {
+		$collector_classes = [
+			'RemoteDataBlocks\Logging\QueryMonitor\RdbHttpRequestCollector',
```
ingeniumed: Was this, and `RdbHttpRequestOutputHtml` supposed to be pushed up as well? I'm not able to find them in the plugin

ingeniumed (commented):
I really like this way of doing the logs. I wonder if we could have all the logs go to the new `Remote Data Blocks` section by default? As we add new panels in, those can get triaged into that panel. The one that I'd really like to see is the caching panel as those cache logs do take up a lot of debugging output.

inc/Logging/QueryMonitor/QueryMonitor.php
chriszarate: Oops, yeah, not intentionally. That class will be in my next PR. No harm since there's a `class_exists` check below, but I'll fix.

chriszarate:
>  I wonder if we could have all the logs go to the new Remote Data Blocks section by default? As we add new panels in, those can get triaged into that panel.

Yes, that's the goal!

> The one that I'd really like to see is the caching panel as those cache logs do take up a lot of debugging output.

Up next! :)

ingeniumed (approved):

PR #467, opened 2025-04-18
Simplify HttpClient to a simple singleton pattern and use TTL request header

chriszarate:
This PR does a few related things that are necessary to unblock improved performance and logging.

1. Make `HttpClient` a true singleton. Originally, `HttpClient` was not implemented as a singleton, but over time we realized that we needed a single instance of our caching middleware to preserve in-memory state. However, it's much simpler and clearer to implement a singleton pattern for `HttpClient` itself.
2. The `HttpClient->init()` method was flawed, because it injected per-request options into shared state. In theory, this could allow the options of one request to overwrite those of another. With a singleton pattern, we can pass all options via the `request` method.
3. Setting a custom TTL did not actually work because of the static caching middleware. We should use the same strategy as the vendor cache middleware that we are extending: Set a custom request header, read it, then remove it before sending the request. This is all handled by the vendor code.
4. Remove unused methods. Early exploration exposed methods for queuing and parallel execution, but they proved incompatible with how block bindings are executed. We can remove them and their tests. Similarly, we remove the unused shorthand `get` and `post` methods in favor of `request`.

### DIFF
--- a/inc/Config/QueryRunner/QueryRunner.php
+++ b/inc/Config/QueryRunner/QueryRunner.php
@@ -83,6 +83,8 @@ class QueryRunner implements QueryRunnerInterface {
 		$pass = ! empty( $parsed_url['pass'] ?? '' ) ? ':' . $parsed_url['pass'] : '';
 		$pass = ( $user || $pass ) ? $pass . '@' : '';
 
+		$origin = sprintf( '%s://%s%s%s%s', $scheme, $user, $pass, $host, $port );
+
 		$request_details = [
 			'method' => $method,
 			'options' => [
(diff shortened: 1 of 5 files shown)


inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -77,16 +82,24 @@ protected function get_request_details( HttpQueryInterface $query, array $input_
...
 		$pass = ( $user || $pass ) ? $pass . '@' : '';
+		$origin = sprintf( '%s://%s%s%s%s', $scheme, $user, $pass, $host, $port );
+
+		$cache_headers = [];
+		if ( intval( $cache_ttl ) > 0 ) {
+			$cache_headers[ RdbCacheStrategy::CACHE_TTL_REQUEST_HEADER ] = $cache_ttl;
+		}
+		if ( intval( $cache_ttl ) < 0 ) {
+			$cache_headers['Cache-Control'] = 'no-store';
+		}
```
maxschmeling, line 93: Just confirming that adding no cache header if `$cache_ttl = 0` is what we intend?

maxschmeling (approved):

inc/Config/QueryRunner/QueryRunner.php
chriszarate: We might revisit, but `0` has no documented behavior right now:

> - Return a positive integer to set a custom TTL in seconds.
> - Return -1 to disable caching.
>  - Return null to use the global default cache TTL (60 seconds).

PR #468, opened 2025-04-18
Remove constant targeting Playground CORS proxy

chriszarate:
Remove constant targeting Playground CORS proxy now that it is enabled by default


maxschmeling:
For context: https://github.com/WordPress/wordpress-playground/pull/2076

maxschmeling (approved):

PR #473, opened 2025-04-21
Add new Queries panel to Query Monitor

chriszarate:
This PR builds on #465 to provide visibility into remote data blocks queries and cache disposition in a dedicated, filterable panel:

https://github.com/user-attachments/assets/3d99ecb8-e2f3-4f03-9e85-d5afff1079f5

It does via a new logging Guzzle middleware.


inc/Logging/QueryMonitor/RdbHttpRequestOutputHtml.php
```diff
@@ -0,0 +1,180 @@
...
+					);
+
+					/* Cache status */
+					echo '<td class="qm-has-toggle qm-ltr qm-nowrap">';
+					echo self::build_toggler(); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+					echo '<ol>';
+					echo '<li>' . esc_html( $row['cache_status'] ) . '</li>';
+					echo '<div class="qm-toggled qm-info qm-supplemental">';
+
+					if ( '' !== $row['cache_age'] ) {
```
ingeniumed, line 104: Love this change. It makes that whole row so much more readable now, as you just need to click into it to see the relevant metadata like age, etc.

ingeniumed (approved):
I vote this as the PR of the year because it makes the QM panel so much more readable.

PR #474, opened 2025-04-21
Cache error responses for a short time

chriszarate:
As discussed in https://github.com/Automattic/remote-data-blocks/pull/461#pullrequestreview-2768267664, remote API downtime or a sustained error rate can threaten the stability of a site. In-memory caching cannot completely mitigate this risk, since a high volume of requests across a high number of workers could still result in a high failure rate.

To address this, we can cache errors in object cache for a short time (30 seconds in this PR). This is very similar to the behavior of our edge cache, which caches 404s for a short time.

We can make this value filterable, but I'd like to see that request presented by a real-world user so that we can tailor the filter to their needs.


ingeniumed (approved):

ingeniumed:
Forgot to add this as a comment - should add this change in the docs so it's known.

PR #475, opened 2025-04-21
Add in-memory caching to `QueryRunner`

chriszarate:
This is a performance and developer experience improvement to remove redundant query executions during a single request. It implements a simple in-memory cache in `QueryRunner` so that repeated identical queries are quickly resolved from cache.

## Why is in-memory caching beneficial?

Mainly, because block bindings with identical queries are often repeated within a remote data block. For example, this remote data block has 7 children core blocks with block bindings, each of which needs the same data from the exact same query:

Screenshot 2025-04-21 at 15 24 30

Currently, for this example, our block binding resolver function executes the exact same query 7 times. Via our Guzzle client, the redundant query executions resolve from object cache. On top of this, WP Core implements its own in-memory cache for object cache, which means that there is only one network request to object cache.

So why do we need this additional caching? Two main reasons:

1. There is additional post-processing that happens after the query is executed. The response is parsed as JSON and traversed recursively by `JsonPath` in order to extract the values defined by the query's output schema.
   - In aggregate across many blocks (or across many posts on an archive page), this can be expensive. 
   - Additionally, the plugin user has the opportunity to define their own callback functions to generate or transform this output, which might contain expensive code. They might have the (very reasonable) expectation that these callback functions would only be called once.
2. While we will soon cache error responses in object cache (#474), some errors unavoidably manifest as exceptions that must be caught, and therefore cannot be cached by Guzzle middleware. We do not want to repeat these requests many times over within the same request.

It has a small side benefit of removing the _appearance_ of multiple calls in our logging and debugging tools:

| Before  | After |
| Screenshot 2025-04-21 at 15 30 07  | Screenshot 2025-04-21 at 15 29 46  |

## Why implement in-memory caching _here_?

The `QueryRunner` implements query execution and response parsing, so it feels like a natural location. We could implement caching at the block binding level, but use cases that executed queries directly (e.g., custom code, REST API, interactivity API) would not benefit.

Another option is to inject the post-processing logic into our Guzzle middleware and cache the result, but that could result in breakage when processing code is updated. It's much safer to cache the raw API response.

### DIFF
--- a/inc/Config/QueryRunner/QueryRunner.php
+++ b/inc/Config/QueryRunner/QueryRunner.php
@@ -18,13 +18,20 @@ defined( 'ABSPATH' ) || exit();
  * Class that executes queries.
  */
 class QueryRunner implements QueryRunnerInterface {
+	/**
+	 * @var array<string, array|WP_Error>
+	 */
+	protected static array $in_memory_cache;
+
 	protected HttpClient $http_client;
 
 	/**
 	 * @param HttpClient|null $http_client The HTTP client used to make HTTP requests.
+	 * @param array|null $in_memory_cache The in-memory cache used to store query results.
 	 */
-	public function __construct( ?HttpClient $http_client = null ) {
+	public function __construct( ?HttpClient $http_client = null, ?array $in_memory_cache = null ) {
 		$this->http_client = $http_client ?? HttpClient::instance();
+		self::$in_memory_cache = $in_memory_cache ?? self::$in_memory_cache ?? [];
 	}
 
 	/**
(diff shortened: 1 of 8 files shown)


inc/Config/QueryRunner/QueryRunner.php
```diff
@@ -152,7 +153,6 @@ protected function get_raw_response_data( HttpQueryInterface $query, array $inpu
 
 		return [
 			'metadata' => [
-				'age' => intval( $response->getHeaderLine( 'Age' ) ),
```
ingeniumed, line 155: I noticed that this is still being referenced in `QueryRunner#get_response_metadata`, just need to remove that.

ingeniumed (approved):
Just one item that needs to be tweaked, otherwise this is good with me.

inc/Config/QueryRunner/QueryRunner.php
chriszarate: Opted to restore it for now in https://github.com/Automattic/remote-data-blocks/pull/475/commits/64549e95e4e01c8b2faa936ef19c35167a1fc354

ingeniumed (approved):

PR #477, opened 2025-04-22
Bugfix: Avoid repeated render of dynamic blocks

chriszarate:
In a block tree, the deepest descendant blocks are rendered first. This makes intuitive sense, since the ancestor blocks cannot be rendered without rendered descendants. This means that our dynamic template block does not need to be rerendered if there is only one result.

Similarly, our dynamic container block does not need to be rerendered if it contains a template block, since that block has already been rendered.

Depends on #478

### DIFF
--- a/inc/Config/QueryRunner/QueryResponseParser.php
+++ b/inc/Config/QueryRunner/QueryResponseParser.php
@@ -55,7 +55,8 @@ final class QueryResponseParser {
 	 */
 	public function parse( mixed $data, array $schema ): mixed {
 		$json_obj = $data instanceof JsonObject ? $data : new JsonObject( $data );
-		$value = $json_obj->get( $schema['path'] ?? '$' );
+		$default_path = $schema['is_collection'] ?? false ? '$[*]' : '$';
+		$value = $json_obj->get( $schema['path'] ?? $default_path );
 
 		if ( is_array( $schema['type'] ?? null ) ) {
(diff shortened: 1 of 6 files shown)


inc/Config/QueryRunner/QueryResponseParser.php
```diff
@@ -55,7 +55,8 @@ private function get_field_value( mixed $field_value, string $type_name, mixed $
 	 */
 	public function parse( mixed $data, array $schema ): mixed {
 		$json_obj = $data instanceof JsonObject ? $data : new JsonObject( $data );
-		$value = $json_obj->get( $schema['path'] ?? '$' );
+		$default_path = $schema['is_collection'] ?? false ? '$[*]' : '$';
```
maxschmeling: The `?? false` just reads poorly to me in this line, though it functions correctly. Does this avoid a warning or something?

We could check `isset` or `array_key_exists` or something, but that's longer.

Maybe just parens around it to make it read better?

```suggestion
		$default_path = ( $schema['is_collection'] ?? false ) ? '$[*]' : '$';
```

maxschmeling (approved):

inc/Config/QueryRunner/QueryResponseParser.php
chriszarate: Done in https://github.com/Automattic/remote-data-blocks/pull/477/commits/1f21745b46881e9036a168a419b4f15a74e0bd9d

PR #479, opened 2025-04-22
Add block binding Query Monitor panel

chriszarate:
Add a panel showing details about block binding that were resolved during the current render, including any related to "field shortcodes."

https://github.com/user-attachments/assets/77dbce6d-bfbb-4bab-b9fc-b60807a2cb0f

Also remove some logging calls that are already surfaced via other channels (like validation errors).

### DIFF
--- a/inc/Logging/QueryMonitor/RdbBlockBindingOutputHtml.php
+++ b/inc/Logging/QueryMonitor/RdbBlockBindingOutputHtml.php
@@ -1,0 +1,189 @@
+<?php declare(strict_types = 1);
+
+namespace RemoteDataBlocks\Logging\QueryMonitor;
+
+use QueryMonitor as QM;
+use QM_Output_Html_Logger;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( class_exists( 'QM_Output_Html_Logger' ) ) {
+	class RdbBlockBindingOutputHtml extends QM_Output_Html_Logger {
+		public static string $collector_id = 'remote-data-blocks-block-binding';
+
+		public function output(): void {
+			/** @var QM_Data_Logger $data */
+			$data = $this->collector->get_data();
+
+			if ( ! empty( $data->logs ) ) {
+				$block_names = array_unique( array_column( $data->logs, 'block_name' ) );
+				$levels = array_unique( array_column( $data->logs, 'level' ) );
+				$remote_data_block_names = array_unique( array_column( $data->logs, 'remote_data_block_name' ) );
+
+				$this->before_tabular_output();
+
+				echo '<thead>';
+				echo '<tr>';
+				echo '<th scope="col" class="qm-filterable-column">';
+				echo $this->build_filter( 'result', $levels, __( 'Level', 'remote-data-blocks' ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+				echo '</th>';
+				echo '<th scope="col">' . esc_html__( 'Message', 'remote-data-blocks' ) . '</th>';
+				echo '<th scope="col" class="qm-filterable-column">';
+				echo $this->build_filter( 'component', $remote_data_block_names, __( 'Remote data block', 'remote-data-blocks' ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+				echo '</th>';
+				echo '<th scope="col" class="qm-filterable-column">';
+				echo $this->build_filter( 'type', $block_names, __( 'Block', 'remote-data-blocks' ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+				echo '</th>';
+				echo '<th scope="col">' . esc_html__( 'Post ID', 'remote-data-blocks' ) . '</th>';
+				echo '<th scope="col">' . esc_html__( 'Caller', 'remote-data-blocks' ) . '</th>';
+				echo '</tr>';
+				echo '</thead>';
+
+				echo '<tbody>';
+
+				foreach ( $data->logs as $row ) {
+					$row_attr = [];
+					$class = '';
+
+					$is_warning = in_array( $row['level'], $this->collector->get_warning_levels(), true );
+
+					if ( $is_warning ) {
+						$class = 'qm-warn';
+					}
+
+					$stack = array();
+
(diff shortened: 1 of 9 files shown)


inc/Logging/QueryMonitor/RdbBlockBindingOutputHtml.php
```diff
@@ -0,0 +1,189 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( class_exists( 'QM_Output_Html_Logger' ) ) {
+	class RdbBlockBindingOutputHtml extends QM_Output_Html_Logger {
+		public static string $collector_id = 'remote-data-blocks-block-binding';
+
+		public function output(): void {
+			/** @var QM_Data_Logger $data */
```
maxschmeling, line 17: Maybe it's overkill, but I would love to define a little function that can handle making the elements and opening/closing. Used AI to generate this:
```
<?php
function el($tag, $attributes = null, $content = null) {
    // Start the element with the tag name
    $html = "<$tag";
    
    // Add attributes if provided
    if (is_array($attributes)) {
        foreach ($attributes as $key => $value) {
            $html .= " " . htmlspecialchars($key) . '="' . htmlspecialchars($value) . '"';
        }
    }

    // Close the opening tag
    $html .= ">";

    // Add content if provided
    if (is_array($content)) {
        foreach ($content as $child) {
            $html .= $child;
        }
    } else {
        $html .= $content;
    }

    // Add the closing tag
    $html .= "</$tag>";

    return $html;
}
```
It would just be easier to read, nest with whitespace, and keep from having accidental malformed HTML when changes are made.
```
echo el('th', ['scope' => 'col'], [
    el('p', null, 'Some Text')
]);
```

Just a thought. Take it or leave it.

maxschmeling: AI tells me this is the code updated to use that function

```
if (!empty($data->logs)) {
    $block_names = array_unique(array_column($data->logs, 'block_name'));
    $levels = array_unique(array_column($data->logs, 'level'));
    $remote_data_block_names = array_unique(array_column($data->logs, 'remote_data_block_name'));

    $this->before_tabular_output();

    echo el('thead', null, [
        el('tr', null, [
            el('th', ['scope' => 'col', 'class' => 'qm-filterable-column'], $this->build_filter('result', $levels, __('Level', 'remote-data-blocks'))),
            el('th', ['scope' => 'col'], esc_html__('Message', 'remote-data-blocks')),
            el('th', ['scope' => 'col', 'class' => 'qm-filterable-column'], $this->build_filter('component', $remote_data_block_names, __('Remote data block', 'remote-data-blocks'))),
            el('th', ['scope' => 'col', 'class' => 'qm-filterable-column'], $this->build_filter('type', $block_names, __('Block', 'remote-data-blocks'))),
            el('th', ['scope' => 'col'], esc_html__('Post ID', 'remote-data-blocks')),
            el('th', ['scope' => 'col'], esc_html__('Caller', 'remote-data-blocks')),
        ])
    ]);

    echo el('tbody', null, array_map(function ($row) {
        $row_attr = [];
        $class = in_array($row['level'], $this->collector->get_warning_levels(), true) ? 'qm-warn' : '';
        $stack = array_map(function ($frame) {
            return self::output_filename($frame['display'], $frame['calling_file'], $frame['calling_line']);
        }, $row['filtered_trace']);

        $row_attr['data-qm-component'] = $row['remote_data_block_name'];
        $row_attr['data-qm-result'] = $row['level'];
        $row_attr['data-qm-type'] = $row['block_name'];

        return el('tr', $row_attr + ['class' => $class], [
            el('td', ['class' => 'qm-nowrap'], [
                $is_warning ? QM::icon('warning') : QM::icon('blank'),
                esc_html(ucfirst($row['level']))
            ]),
            el('td', null, esc_html($row['message'])),
            el('td', null, esc_html($row['remote_data_block_name'])),
            el('td', ['class' => 'qm-has-toggle qm-ltr qm-wrap'], [
                self::build_toggler(),
                esc_html($row['block_name']),
                el('div', ['class' => 'qm-toggled qm-info qm-supplemental'], [
                    el('table', null, [
                        el('tbody', null, array_map(function ($key, $value) {
                            return el('tr', null, [
                                el('td', null, esc_html($key)),
                                el('td', null, el('pre', null, el('code', null, esc_html(wp_json_encode($value)))))
                            ]);
                        }, array_keys($row['block_info']), $row['block_info']))
                    ])
                ])
            ]),
            el('td', null, esc_html(strval($row['post_id']))),
            el('td', ['class' => 'qm-has-toggle qm-nowrap qm-ltr'], [
                !empty($stack) ? self::build_toggler() : '',
                el('ol', null, [
                    el('li', null, $caller = array_shift($stack)),
                    !empty($stack) ? el('div', ['class' => 'qm-toggled'], el('li', null, implode('</li><li>', $stack))) : ''
                ])
            ])
        ]);
    }, $data->logs));

    $this->after_tabular_output();
} else {
    $this->before_non_tabular_output();
    $notice = __('No block bindings.', 'remote-data-blocks');
    echo $this->build_notice($notice);
    $this->after_non_tabular_output();
}
```

maxschmeling (approved):

inc/Logging/QueryMonitor/RdbBlockBindingOutputHtml.php
chriszarate: I dig it. The counterpoint is that we are extending classes from Query Monitor ([like this one](https://github.com/johnbillion/query-monitor/blob/9d666542c508dad471127bd20b4b444f323d4e29/output/html/logger.php#L37)) so copying their procedural code keeps things aligned and makes it possible to compare the two when debugging.

maxschmeling: Yeah, I get keeping it similar. Though the part that changes is just simple string generation...

I really don't care either way though. I have an aversion to unstructured string concatenation for generating code like that. But keeping consistency there isn't bad and it's not a huge problem... especially since the output isn't something that is used in the actual site output.

Feel free to leave it as is.

chriszarate: I'll do this in a separate PR across all the classes. 👍 It's kind of dodging the linter in a way that could backfire. I want to be careful since technically we're working with untrusted input.

ingeniumed (approved):

PR #480, opened 2025-04-23
Add support for rendering empty results

ingeniumed:
### Description

This PR is meant to introduce a new block that is specifically for rendering empty results in the event that a query comes back empty. This has been done in a way that makes the message customisable without blindly outputting html. It won't render this in the event that the block has been accidentally placed outside a remote data block, or if the query gives back results.

## This is how the block layout looks like in the editor

CleanShot 2025-04-23 at 15 03 13

## This is how the empty block looks like in the editor

CleanShot 2025-04-23 at 07 04 48

## This is how the post would look like when it's rendered

CleanShot 2025-04-23 at 15 02 57

There's another bug fix I made which was when the rows that have already been mapped in a remote data block for a Google Sheets data source are deleted. This causes the plugin to error out and not fall back to the stale data.

### Testing

- Create a new data source with google sheets
- Ensure the sheet is empty
- This should render this block


inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -198,6 +198,32 @@ private static function execute_queries( array $block_context, array $source_arg
 		}
 	}
 
+	public static function should_render_empty_result( WP_Block $block ): bool {
```
ingeniumed, line 203: This has been done this way so that the calls are centralized to this class just like the pagination block.

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -198,6 +198,32 @@ private static function execute_queries( array $block_context, array $source_arg
...
+		$block_context = $block->context[ self::$context_name ] ?? [];
+		$empty_message = __( 'No results found.', 'remote-data-blocks' );
+
+		/**
+		 * Filter the empty result message for a block binding.
+		 *
+		 * @param string $empty_message The original empty result message.
+		 * @param array $block_context The block context.
+		 */
+		$empty_message = apply_filters(
```
ingeniumed: This allows the empty message to be customized, in case users want to tweak this. Note that, this is meant to be just a string rather than an HTML that could create problems for us.

src/blocks/remote-data-container/hooks/usePatterns.ts
```diff
@@ -76,6 +76,8 @@ export function usePatterns( remoteDataBlockName: string, rootClientId: string =
 			innerBlocks.push( createBlock( 'remote-data-blocks/pagination' ) );
 		}
 
+		innerBlocks.push( createBlock( 'remote-data-blocks/empty-result' ) );
```
ingeniumed: This ensures it's always available, and we can decide when to render it. If there's another place to insert this, let me know.

src/blocks/remote-data-empty-result/edit.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+import './editor.scss';
+
+export function Edit( props: BlockEditProps< RemoteDataEmptyResultBlockAttributes > ): JSX.Element {
+	const { context } = props;
+	const blockProps = useBlockProps();
+
+	const { remoteData } = useRemoteDataContext( context );
+
+	if ( ! remoteData?.results ) {
```
ingeniumed: Placeholder in the event that results don't exist, same as pagination

src/blocks/remote-data-empty-result/edit.tsx
```diff
@@ -0,0 +1,43 @@
...
+					label={ __( 'Remote Data Empty Result' ) }
+					instructions={ __(
+						'This block only works when placed inside a remote data block when there are no results. This block will be ignored as currently configured.'
+					) }
+				/>
+			</div>
+		);
+	}
+
+	if ( remoteData?.results.length === 0 ) {
```
ingeniumed: Only render this block if the results are empty, otherwise it's not shown.

src/blocks/remote-data-empty-result/render.php
```diff
@@ -0,0 +1,24 @@
...
+use RemoteDataBlocks\Editor\DataBinding\BlockBindings;
+// Global variables provided by WordPress for block rendering:
+// $attributes (array): The block attributes.
+// $content (string): The block default content.
+// $block (WP_Block): The block instance.
+
+$should_render_empty_result = BlockBindings::should_render_empty_result( $block );
+
+// Skip the rendering if the block's results are not empty.
+if ( ! $should_render_empty_result ) {
```
ingeniumed: Same thing as the JS, skip if the results are not empty otherwise show the empty message which can be customized

src/blocks/remote-data-pagination/style.scss
```diff
@@ -3,7 +3,7 @@
  */
 
 .remote-data-pagination {
-	color: var(--wp--custom--remote-data-blocks--paginatin-link-disabled, #999);
+	color: var(--wp--custom--remote-data-blocks--pagination-link-disabled, #999);
```
ingeniumed, line 6: Noticed this was a typo

inc/Integrations/Google/Sheets/GoogleSheetsDataSource.php
```diff
@@ -100,6 +100,14 @@ public static function preprocess_get_response( array $response_data, array $inp
 		if ( isset( $response_data['values'] ) && is_array( $response_data['values'] ) ) {
 			$values = $response_data['values'];
 			$columns = array_shift( $values ); // Get column names from first row
+
+			// if the values are now empty, give back $selected_row as null
+			// This can happen if rows are deleted in the sheet after they
+			// have been used in a remote data block.
+			if ( empty( $values ) ) {
```
ingeniumed, line 107: I noticed that when you delete a row in a google sheet that's already been mapped, it causes the whole plugin to error out due to an out of bounds check here. I ran into it while trying to test out what would happen if I did this case. It will render the stale data now.

chriszarate:
Instead of a fixed block with (filterable) placeholder message, I expected a block that accepts children blocks (with a default paragraph block on insertion). This is how the [core query-loop-no-results block](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/query-no-results) works. That provides a lot more flexibility with about the same amount of code.

src/blocks/remote-data-empty-result/block.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "https://schemas.wp.org/trunk/block.json",
+  "apiVersion": 3,
+  "name": "remote-data-blocks/empty-result",
+  "version": "0.1.0",
+  "usesContext": [ "remote-data-blocks/remoteData" ],
+  "title": "Remote Data Empty Result",
```
chriszarate: "Empty result" doesn't quite parse to me. We want to display content when there are no results. Can we follow the pattern of the core block and call it "No Remote Data Results"?

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -198,6 +198,32 @@ private static function execute_queries( array $block_context, array $source_arg
 		}
 	}
 
+	public static function should_render_empty_result( WP_Block $block ): bool {
+		$block_context = $block->context[ self::$context_name ] ?? [];
+
+		// Only give back true if there are no results
+		return isset( $block_context['results'] ) && empty( $block_context['results'] );
```
chriszarate: You are relying on the serialized results from the last block save, not the real-time query results. You will need to execute the query again.

src/blocks/remote-data-container/hooks/usePatterns.ts
chriszarate: This is the right place. 👍

src/blocks/remote-data-empty-result/edit.tsx
chriszarate: On the edit side, it should always be rendered / editable.

ingeniumed: Yip, I didn't mean it won't be rendered at all. It is always rendered but just with a default div when there isn't a need to show it. It'll be a placeholder, or the empty message otherwise. 

Hopefully that clears it up - I should have said only render the empty message instead of "only render this block".

src/blocks/remote-data-empty-result/block.json
ingeniumed: Agreed, I'll make that change.

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: Good catch. I hadn't done that, as I wanted to let the remote data block handle fetching the new data. Thinking about how the dynamic block would work, we would need to get the real-time query result here.

src/blocks/remote-data-no-result/edit.tsx
```diff
@@ -0,0 +1,52 @@
...
+];
+
+export function Edit( props: BlockEditProps< RemoteDataNoResultBlockAttributes > ): JSX.Element {
+	const { context } = props;
+	const { remoteData } = useRemoteDataContext( context );
+	const blockProps = useBlockProps();
+
+	// This is mirroring the query-no-results block from Gutenberg.
+	// https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/query-no-results/edit.js
+	let templateToUse = NO_RESULTS_PLACEHOLDER_TEMPLATE;
```
ingeniumed: This is done the same way that the query-no-results block is done with the use of templates. The idea is that the template represents the initial state of the block which in our case is either a placeholder, no results message or nothing at all. In the case of nothing at all, there are no inner blocks to render so it's just an empty block.

src/blocks/remote-data-no-result/render.php
```diff
@@ -0,0 +1,16 @@
...
+// $block (WP_Block): The block instance.
+
+$should_render_empty_result = BlockBindings::should_render_empty_result( $block );
+
+// Skip the rendering if the block's results are not empty.
+if ( ! $should_render_empty_result ) {
+	return;
+}
+
+echo wp_kses_post( $content );
```
ingeniumed: Only if no results are found would the content of the no results block be rendered.

src/blocks/remote-data-template/components/loop-template/LoopTemplate.tsx
```diff
@@ -30,8 +29,10 @@ export function LoopTemplate( props: LoopTemplateProps ) {
 		[ clientId ]
 	);
 
+	// Don't render anything if there are no results.
+	// Leave it to the no results block to handle this.
 	if ( ! remoteData.results.length ) {
-		return <p>{ __( 'No results found.' ) }</p>;
+		return null;
```
ingeniumed: This stops double "no results found" by ensuring that only the no results block handles this case

ingeniumed:
I have changed the no results block to now be block based, so it can be customized to one's needs. I have also removed the double "No results found" problem by not having the LoopTemplate handle it.

CleanShot 2025-04-24 at 13 02 38

CleanShot 2025-04-24 at 13 02 49

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -198,6 +198,14 @@ private static function execute_queries( array $block_context, array $source_arg
 		}
 	}
 
+	public static function should_render_empty_result( WP_Block $block ): bool {
+		$block_context = $block->context[ self::$context_name ] ?? [];
+		$query_response = self::execute_queries( $block_context, [], 'remote_data_block_check_for_no_results' );
```
chriszarate: need to merge `trunk` I think
```suggestion
		$query_response = self::execute_queries( $block_context, [] );
```

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -198,6 +198,14 @@ private static function execute_queries( array $block_context, array $source_arg
...
+
+		// Only give back true if there are no results
+		return isset( $query_response['results'] ) && empty( $query_response['results'] );
```
chriszarate, line 214: We might want to think about whether we want this block to render in error conditions—or only if there is a successful response and no results. The core "No Results" block doesn't really have this dilemma since WP_Query is realistically always going to succeed.

src/blocks/remote-data-no-result/block.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "https://schemas.wp.org/trunk/block.json",
+  "apiVersion": 3,
+  "name": "remote-data-blocks/no-result",
+  "version": "0.1.0",
+  "usesContext": [ "remote-data-blocks/remoteData" ],
+  "title": "No Remote Data Results",
```
chriszarate: Results or result? Should be consistent with block name

src/blocks/remote-data-no-result/edit.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+	if ( remoteData?.results?.length === 0 ) {
+		templateToUse = NO_RESULTS_TEMPLATE;
+	} else if ( remoteData?.results ) {
+		templateToUse = [];
+	}
+
```
chriszarate: I don't think this logic is useful. Similar to the core block, we always want the editor to see and edit the content that _would_ display if there are no results, regardless of whether there are _currently_ any results.

```suggestion
```

src/blocks/remote-data-no-result/render.php
```diff
@@ -0,0 +1,16 @@
...
+// Global variables provided by WordPress for block rendering:
+// $attributes (array): The block attributes.
+// $content (string): The block default content.
+// $block (WP_Block): The block instance.
+
+$should_render_empty_result = BlockBindings::should_render_empty_result( $block );
+
+// Skip the rendering if the block's results are not empty.
+if ( ! $should_render_empty_result ) {
+	return;
```
chriszarate: This is a kind of esoteric use of `return` outside a function, but has the intended effect. I might amend to `return null;` since IIRC it is coerced to null anyway. 

https://www.php.net/manual/en/function.return.php

```suggestion
	return null;
```

src/blocks/remote-data-no-result/render.php
chriszarate: I like all these explanatory comments! Would only say if you think they are useful for PR review then they would also be useful as code comments.

src/blocks/remote-data-template/components/loop-template/LoopTemplate.tsx
chriszarate: Since this is just a function of a prop, it's usually better to hoist this logic into the parent component (`Edit` in this case). Mentioned in [React docs](https://react.dev/learn/conditional-rendering#conditionally-returning-nothing-with-null)

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: I thought about that, and I was leaning towards making a separate block for errors. My main motivation for that was, there could be different error states based on the type of error that's encountered. When we don't have any results it's easy to decide that we want to show something that says, no results are present. Now this is assuming that we want to hand over that kind of control to the editor, and don't want to just have the single error state.

src/blocks/remote-data-no-result/edit.tsx
ingeniumed: Fair enough, it is a state that we want them to configure so it should always be visible.

src/blocks/remote-data-no-result/block.json
ingeniumed: Changed it to result

src/blocks/remote-data-no-result/render.php
ingeniumed: I did add a bunch of em already, but I'll go through and add more based on my latest commit

src/blocks/remote-data-no-result/edit.tsx
```diff
@@ -0,0 +1,48 @@
...
+		},
+	],
+];
+
+const NO_RESULTS_PLACEHOLDER_TEMPLATE: Template[] = [
+	[
+		'core/paragraph',
+		{
+			placeholder: __(
+				'This block only works when placed inside a remote data block when there are no results. This block will be ignored as currently configured.'
```
chriszarate: Slight clarification since in the editor it doesn't matter if there are no results currently.

```suggestion
				'This block must be placed inside a remote data block. This block will be ignored as currently configured.'
```

src/blocks/remote-data-no-result/edit.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export function Edit( props: BlockEditProps< RemoteDataNoResultBlockAttributes > ): JSX.Element {
+	const { context } = props;
+	const { remoteData } = useRemoteDataContext( context );
+	const blockProps = useBlockProps();
+
+	// This is mirroring the query-no-results block from Gutenberg.
+	// https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/query-no-results/edit.js
+	const templateToUse = ! remoteData?.blockName
+		? NO_RESULTS_PLACEHOLDER_TEMPLATE
```
chriszarate: When the placeholder is used, I think we might want to return a "real" placeholder using the core UI component instead of a template.

src/blocks/remote-data-pagination/edit.tsx
```diff
@@ -12,20 +13,17 @@ import './editor.scss';
...
 		return (
-			<div { ...blockProps }>
-				<Placeholder
-					label={ __( 'Remote Data Pagination' ) }
-					instructions={ __(
-						'This block only works when placed inside a remote data block using data that supports pagination. This block will be ignored as currently configured.'
-					) }
-				/>
-			</div>
+			<Placeholder
```
chriszarate, line 22: This should be added above the existing check for pagination data, not replace it

ingeniumed: Do you mean it should be placed as a seperate check to see if its outside the remote data block vs if pagination is configured or not?

Shouldn't this work just like the empty results block where a customer can configure it always in the editor, and then it should only render if there is pagination data? I have added the check where it won't render if there are not pagination links.

chriszarate: > Shouldn't this work just like the empty results block where a customer can configure it always in the editor, and then it should only render if there is pagination data?

The key difference is that pagination is at the query schema level, so it's known in advance and not a run-time discovery. So we can let the user know it will never work.

ingeniumed: Fair point. I'll update this to have two separate checks then.

chriszarate (approved):

PR #492, opened 2025-05-01
apply http data source service config to request

shekharnwagh:
The generic HTTP data source supports authentication configuration in the service config, including Bearer Token, Basic Auth, and API Key. Authentication headers are now properly applied to outgoing requests based on the configured authentication type -

- Bearer Token: Adds to `Authorization` header with value `Bearer {token}`
- Basic Auth: Adds to `Authorization` header with value `Basic {base64_encoded_credentials}`
- API Key: Adds to `Authorization` header if `add_to` is `header`, or as a query parameter if `add_to` is `queryparams`

### DIFF
--- a/inc/Config/DataSource/HttpDataSource.php
+++ b/inc/Config/DataSource/HttpDataSource.php
@@ -98,11 +98,51 @@ class HttpDataSource extends ArraySerializable implements HttpDataSourceInterfac
 		return ConfigSchemas::get_http_data_source_service_config_schema();
 	}
 
+	private static function get_auth_to_request_headers( array $service_config ): array {
+		if ( ! isset( $service_config['auth'] ) ) {
+			return [];
+		}
+
+		$auth_headers = [];
+		$auth = $service_config['auth'];
+
+		if ( 'bearer' === $auth['type'] ) {
+			$auth_headers['Authorization'] = 'Bearer ' . $auth['value'];
+		} elseif ( 'basic' === $auth['type'] ) {
+			$auth_headers['Authorization'] = 'Basic ' . base64_encode( $auth['value'] );
+		} elseif ( 'api-key' === $auth['type'] && 'header' === $auth['add_to'] ) {
+			$auth_headers[ $auth['key'] ] = $auth['value'];
+		}
+
+		return $auth_headers;
+	}
+
+	private static function add_auth_to_endpoint( string|null $endpoint, array $service_config ): string|null {
+		if ( ! isset( $service_config['auth'] ) 
+			|| 'api-key' !== $service_config['auth']['type'] 
+			|| 'queryparams' !== $service_config['auth']['add_to']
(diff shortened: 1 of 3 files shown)


inc/Config/DataSource/HttpDataSource.php
```diff
@@ -98,11 +98,51 @@ protected static function get_service_config_schema(): array {
...
+
+		if ( 'bearer' === $auth['type'] ) {
+			$auth_headers['Authorization'] = 'Bearer ' . $auth['value'];
+		} elseif ( 'basic' === $auth['type'] ) {
+			$auth_headers['Authorization'] = 'Basic ' . base64_encode( $auth['value'] );
+		} elseif ( 'api-key' === $auth['type'] && 'header' === $auth['add_to'] ) {
+			$auth_headers[ $auth['key'] ] = $auth['value'];
+		}
+
+		return $auth_headers;
```
ingeniumed, line 117: It might be worth adding a filter here to allow customizing these headers. I know we only allow 3 types of authorization schemes, but [there are more schemes possible.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Authorization). This might be a good way to support those without expanding the dropdown?

inc/Config/DataSource/HttpDataSource.php
```diff
@@ -98,11 +98,51 @@ protected static function get_service_config_schema(): array {
...
+			|| 'api-key' !== $service_config['auth']['type'] 
+			|| 'queryparams' !== $service_config['auth']['add_to']
+			|| empty( $endpoint )
+		) {
+			return $endpoint;
+		}
+
+		$query_string = $service_config['auth']['key'] . '=' . $service_config['auth']['value'];
+		
+		return $endpoint . ( strpos( $endpoint, '?' ) === false ? '?' : '&' ) . $query_string;
```
ingeniumed, line 131: It might be time to use some built in methods from WordPress or PHP to do URL construction rather than doing it by hand.

inc/Config/DataSource/HttpDataSource.php
```diff
@@ -98,11 +98,51 @@ protected static function get_service_config_schema(): array {
...
+	}
+
 	protected static function map_service_config( array $service_config ): array {
+		$request_headers = $service_config['request_headers'] ?? [];
```
ingeniumed, line 135: What do you think about trying to use the existing request_headers function to add these auth headers in? My thought was that, it should all happen in one place because they are headers for the request at the end of the day.

inc/Config/DataSource/HttpDataSource.php
chriszarate: Agreed

https://developer.wordpress.org/reference/functions/add_query_arg/

maxschmeling:
This is superseded by #495

PR #493, opened 2025-05-02
Remove `Psr\Log`, add "Logs" Query Monitor subpanel

chriszarate:
1. Remove `Psr\Log`. We ran into type hint conflicts when adding this plugin to some internal sites. Those sites used plugins with their own (older and incompatible) version of `Psr\Log`. Since it's likely that other users might hit this same issue, we will remove this library and provide a small implementation of our own.
2. Fixes a bug (typo) that prevented Query Monitor from working.
1. Remove `wpcomvip_log_to_query_monitor` action, since we now log to Query Monitor in our own panel.
1. Rename `wpcomvip_log` action to `remote_data_blocks_log`.
1. Add an additional "Logs" subpanel to our Query Monitor panel. This ensures that all RDB debugging output goes to our panel, instead of being split across core and custom panels:

Screenshot 2025-05-01 at 18 10 18


inc/Logging/QueryMonitor/RdbBlockBindingOutputHtml.php
```diff
@@ -3,15 +3,15 @@
...
 use QueryMonitor as QM;
-use QM_Output_Html_Logger;
 
 if ( ! defined( 'ABSPATH' ) ) {
 	exit;
 }
 
-if ( class_exists( 'QM_Output_Html_Logger' ) ) {
-	class RdbBlockBindingOutputHtml extends QM_Output_Html_Logger {
+if ( class_exists( __NAMESPACE__ . '\RdbLogOutputHtml' ) ) {
```
alecgeatches, line 11: Do we still need to check for this class existing?

alecgeatches: Nevermind, I see that `RdbLogOutputHtml` depends on the existence of `QM_Output_Html_Logger`, which depends on Query Monitor being installed.

alecgeatches:
When testing this, I noticed some repeated warnings on the logger frontend when expanding successful block bindings:

Screenshot 2025-05-02 at 10 50 17 AM

They all seem to be the same warning:

```
Warning: Undefined array key "block_info" in /var/www/html/wp-content/plugins/remote-data-blocks/inc/Logging/QueryMonitor/RdbBlockBindingOutputHtml.php on line 110 Warning: foreach() argument must be of type array|object, null given in /var/www/html/wp-content/plugins/remote-data-blocks/inc/Logging/QueryMonitor/RdbBlockBindingOutputHtml.php on line 110
```

chriszarate:
> They all seem to be the same warning:
> 
> ```
> Warning: Undefined array key "block_info" in /var/www/html/wp-content/plugins/remote-data-blocks/inc/Logging/QueryMonitor/RdbBlockBindingOutputHtml.php on line 110 Warning: foreach() argument must be of type array|object, null given in /var/www/html/wp-content/plugins/remote-data-blocks/inc/Logging/QueryMonitor/RdbBlockBindingOutputHtml.php on line 110
> ```

@alecgeatches Thanks, fixed in https://github.com/Automattic/remote-data-blocks/pull/493/commits/6dec67b0e11594d599531348b1646f4df549dc49

alecgeatches (approved):
👍 Looking good!

ingeniumed (approved):

PR #495, opened 2025-05-02
Add GenericHttpDataSource

chriszarate:
Instead of using `HttpDataSource` directly, we extend it to create `GenericHttpDataSource`. This allows us to specify (and validate) an opinionated service config that does not impose any constraints on consumers of `HttpDataSource`.

After discussion with @shekharnwagh, we agreed to pull `service_config` out of `HttpDataSource` entirely since it is a private implementation detail for our integrations.


ingeniumed (approved):

ingeniumed:
Closes VIPCMS-1226

PR #496, opened 2025-05-02
Add an error block

ingeniumed:
### Description

The purpose of this block variation is to have a block that can be shown when a query errors out. This is a variation of the existing empty block as that ensures no duplication of blocks.

CleanShot 2025-05-02 at 17 22 04

CleanShot 2025-05-02 at 17 49 57

There are going to be 2 blocks now:

- No results
- Error

There is a mode attribute set based on what block it is, that will decide what's rendered.

### DIFF
--- a/inc/Editor/BlockManagement/BlockRegistration.php
+++ b/inc/Editor/BlockManagement/BlockRegistration.php
@@ -53,8 +53,8 @@ class BlockRegistration {
 		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-no-results', [
 			'variations' => [
 				[
-					'name' => 'remote-data-blocks/error-fallback',
-					'title' => 'Error Fallback',
+					'name' => 'remote-data-blocks/error',
+					'title' => 'Error',
 					'description' => 'Display an error message when the remote data fails to load.',
 					'attributes' => [
 						'mode' => 'error',
--- a/inc/Editor/BlockManagement/BlockRegistration.php
+++ b/inc/Editor/BlockManagement/BlockRegistration.php
@@ -50,19 +50,8 @@ class BlockRegistration {
 		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-pagination' );
 
 		// Remote data error fallback block - used to render a message when results aren't available to load.
-		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-no-results', [
-			'variations' => [
-				[
-					'name' => 'remote-data-blocks/error',
-					'title' => 'Error',
-					'description' => 'Display an error message when the remote data fails to load.',
-					'attributes' => [
-						'mode' => 'error',
-					],
-					'isActive' => [ 'mode' ],
-				],
-			],
-		] );
+		// This registers the empty results block, with a variation for when there is an error.
+		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-no-results' );
 
 		// Remote data template - used to render remote data collections.
 		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-template' );
--- a/inc/Editor/BlockManagement/BlockRegistration.php
+++ b/inc/Editor/BlockManagement/BlockRegistration.php
@@ -49,8 +49,23 @@ class BlockRegistration {
 		// Remote data pagination block - used to render pagination links for collections.
 		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-pagination' );
 
-		// Remote data empty result block - used to render a message when no results are found.
-		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-no-results' );
+		// Remote data error fallback block - used to render a message when results aren't available to load.
+		$result = register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-no-results', [
+			'variations' => [
+				[
+					'name' => 'remote-data-blocks/error-fallback',
+					'title' => 'Error Fallback',
+					'description' => 'Display an error message when the remote data fails to load.',
+					'attributes' => [
+						'mode' => 'error',
+					],
+				],
+			],
+		] );
+
+		// if ( empty( $result->variations ) ) {
+		// 	error_log( 'No variations found for remote-data-no-results block' );
+		// }
 
 		// Remote data template - used to render remote data collections.
 		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-template' );
--- a/inc/Editor/BlockManagement/BlockRegistration.php
+++ b/inc/Editor/BlockManagement/BlockRegistration.php
@@ -50,7 +50,7 @@ class BlockRegistration {
 		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-pagination' );
 
 		// Remote data error fallback block - used to render a message when results aren't available to load.
-		$result = register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-no-results', [
+		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-no-results', [
 			'variations' => [
 				[
 					'name' => 'remote-data-blocks/error-fallback',
@@ -59,13 +59,10 @@ class BlockRegistration {
 					'attributes' => [
 						'mode' => 'error',
 					],
+					'isActive' => [ 'mode' ],
 				],
 			],
 		] );
-
-		// if ( empty( $result->variations ) ) {
-		// 	error_log( 'No variations found for remote-data-no-results block' );
-		// }
 
 		// Remote data template - used to render remote data collections.
 		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-template' );
--- a/inc/Editor/DataBinding/BlockBindings.php
+++ b/inc/Editor/DataBinding/BlockBindings.php
@@ -210,17 +210,19 @@ class BlockBindings {
 		// stale results from the block.
 		$query_response = self::execute_queries( $block_context, [] );
 
-		// If there is an error, and it's the error block variation, return true.
+		// If there is an error, and it's the error block variation, the fallback
+		// content should be rendered.
 		if ( is_wp_error( $query_response ) ) {
 			return 'error' === $attributes['mode'];
 		}
 
-		// If there are no results, and it's the empty block variation, return true.
+		// If there are no results, and it's the empty block variation, the fallback
+		// content should be rendered.
 		if ( isset( $query_response['results'] ) && empty( $query_response['results'] ) ) {
 			return 'empty' === $attributes['mode'];
 		}
 
-		// If there are results, it's fine to render the block.
+		// If there are results, the fallback content should not be rendered.
 		return false;
 	}
 
--- a/inc/Editor/DataBinding/BlockBindings.php
+++ b/inc/Editor/DataBinding/BlockBindings.php
@@ -204,20 +204,23 @@ class BlockBindings {
 		}
 	}
 
-	public static function determine_block_state_to_render( WP_Block $block ): string|null {
+	public static function get_empty_or_error_state_for_block( WP_Block $block ): string|null {
 		$block_context = $block->context[ self::$context_name ] ?? [];
 		// Re-execute the query to get the latest results, rather than using the
 		// stale results from the block.
 		$query_response = self::execute_queries( $block_context, [] );
 
+		// If there is an error, return the error state.
 		if ( is_wp_error( $query_response ) ) {
 			return 'error';
 		}
 
+		// If there are no results, return the empty state.
 		if ( isset( $query_response['results'] ) && empty( $query_response['results'] ) ) {
 			return 'empty';
 		}
 
+		// If there are results, it's fine to render the block.
 		return null;
 	}
 
--- a/inc/Editor/DataBinding/BlockBindings.php
+++ b/inc/Editor/DataBinding/BlockBindings.php
@@ -204,7 +204,7 @@ class BlockBindings {
 		}
 	}
 
-	public static function is_error_or_empty_state( array $context, array $attributes ): bool {
+	public static function should_render_fallback_content( array $context, array $attributes ): bool {
 		$block_context = $context[ self::$context_name ] ?? [];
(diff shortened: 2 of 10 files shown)


inc/Editor/BlockManagement/BlockRegistration.php
```diff
@@ -49,8 +49,20 @@ public static function register_helper_blocks(): void {
 		// Remote data pagination block - used to render pagination links for collections.
 		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-pagination' );
 
-		// Remote data empty result block - used to render a message when no results are found.
-		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-no-results' );
+		// Remote data error fallback block - used to render a message when results aren't available to load.
```
chriszarate, line 52: Update comment to indicate that it registers both the block and its variation.

inc/Editor/BlockManagement/BlockRegistration.php
```diff
@@ -49,8 +49,20 @@ public static function register_helper_blocks(): void {
...
+					'name' => 'remote-data-blocks/error',
+					'title' => 'Error',
+					'description' => 'Display an error message when the remote data fails to load.',
+					'attributes' => [
+						'mode' => 'error',
+					],
+					'isActive' => [ 'mode' ],
+				],
+			],
+		] );
```
chriszarate: I tested without these args and works fine.

```suggestion
		register_block_type( REMOTE_DATA_BLOCKS__PLUGIN_DIRECTORY . '/build/blocks/remote-data-no-results' );
```

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -204,18 +204,24 @@ private static function execute_queries( array $block_context, array $source_arg
 		}
 	}
 
-	public static function should_render_empty_result( WP_Block $block ): bool {
+	public static function get_empty_or_error_state_for_block( WP_Block $block ): string|null {
```
chriszarate: I would suggest keeping it as a `bool` return and inspecting block attributes
```suggestion
	public static function should_render_fallback_content( array $attributes ): bool {
```

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -204,18 +204,24 @@ private static function execute_queries( array $block_context, array $source_arg
...
 		$block_context = $block->context[ self::$context_name ] ?? [];
 		// Re-execute the query to get the latest results, rather than using the
 		// stale results from the block.
 		$query_response = self::execute_queries( $block_context, [] );
 
+		// If there is an error, return the error state.
 		if ( is_wp_error( $query_response ) ) {
-			return false;
+			return 'error';
```
chriszarate:
```suggestion
			return 'error' === $attributes['mode'] ?? 'unknown';
```

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -204,18 +204,24 @@ private static function execute_queries( array $block_context, array $source_arg
...
 		}
 
-		// Only give back true if there are no results
-		return isset( $query_response['results'] ) && empty( $query_response['results'] );
+		// If there are no results, return the empty state.
+		if ( isset( $query_response['results'] ) && empty( $query_response['results'] ) ) {
+			return 'empty';
```
chriszarate:
```suggestion
			return 'empty' === $attributes['mode'] ?? 'unknown';
```

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -204,18 +204,24 @@ private static function execute_queries( array $block_context, array $source_arg
...
+		}
+
+		// If there are results, it's fine to render the block.
+		return null;
```
chriszarate:
```suggestion
		return false;
```

inc/Editor/DataBinding/BlockBindings.php
chriszarate: @ingeniumed Can you add some unit tests for this method?

inc/Editor/BlockManagement/BlockRegistration.php
ingeniumed: Good point, done. I hadn't done that as the variation was happening in line here

inc/Editor/BlockManagement/BlockRegistration.php
ingeniumed: It hadn't worked before, but I am guessing the JS registration or the `createBlock` using the variation worked.

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: I had done it this way at first, as I wasn't 100% if the variation would load this way. Didn't revisit this after it started working, so it's simplified now.

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: Done

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: Done

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: Done

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: Yip, added it in. Got used to working on the pieces where it's not possible to add tests and missed out on this one.

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -204,18 +204,24 @@ private static function execute_queries( array $block_context, array $source_arg
...
+	public static function is_error_or_empty_state( array $context, array $attributes ): bool {
+		$block_context = $context[ self::$context_name ] ?? [];
 		// Re-execute the query to get the latest results, rather than using the
 		// stale results from the block.
 		$query_response = self::execute_queries( $block_context, [] );
 
+		// If there is an error, and it's the error block variation, return true.
 		if ( is_wp_error( $query_response ) ) {
-			return false;
+			return 'error' === $attributes['mode'] ?? 'unknown';
```
chriszarate: My mistake in the suggestion. Order of operations matters a lot with the nullish coalescing operator. Either we should group or pass it just the `mode` attribute
```suggestion
			return 'error' === ( $attributes['mode'] ?? 'unknown' );
```

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -204,18 +204,24 @@ private static function execute_queries( array $block_context, array $source_arg
...
+		}
+
+		// If there are no results, and it's the empty block variation, return true.
+		if ( isset( $query_response['results'] ) && empty( $query_response['results'] ) ) {
+			return 'empty' === $attributes['mode'] ?? 'unknown';
```
chriszarate: same here
```suggestion
			return 'empty' === ( $attributes['mode'] ?? 'unknown' );
```

src/blocks/remote-data-no-results/render.php
```diff
@@ -7,10 +7,10 @@
 // $content (string): The block default content.
 // $block (WP_Block): The block instance.
 
-$should_render_empty_result = BlockBindings::should_render_empty_result( $block );
+$state = BlockBindings::is_error_or_empty_state( $block->context, $attributes );
 
-// Skip the rendering if the block's results are not empty.
-if ( ! $should_render_empty_result ) {
+// The state will only be true when the query gives back an error, or no results and the block attribute matched the response.
```
chriszarate: This comment doesn't quite make sense

src/blocks/remote-data-no-results/render.php
```diff
@@ -7,10 +7,10 @@
 // $content (string): The block default content.
 // $block (WP_Block): The block instance.
 
-$should_render_empty_result = BlockBindings::should_render_empty_result( $block );
+$state = BlockBindings::is_error_or_empty_state( $block->context, $attributes );
```
chriszarate: `$state` doesn't seem like an accurate variable name. maybe `$should_render`?

tests/inc/Editor/DataBinding/BlockBindingsTest.php
```diff
@@ -38,6 +39,110 @@ protected function tearDown(): void {
 		Mockery::close();
 	}
 
+	/**
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test_is_error_or_empty_state_with_error_mode(): void {
+			/**
```
chriszarate: fix alignment

inc/Editor/DataBinding/BlockBindings.php
```diff
@@ -204,18 +204,24 @@ private static function execute_queries( array $block_context, array $source_arg
 		}
 	}
 
-	public static function should_render_empty_result( WP_Block $block ): bool {
-		$block_context = $block->context[ self::$context_name ] ?? [];
+	public static function is_error_or_empty_state( array $context, array $attributes ): bool {
```
chriszarate: I think I might prefer something like `should_render_fallback_content` since that is its intended purpose.

tests/inc/Editor/DataBinding/BlockBindingsTest.php
```diff
@@ -38,6 +39,110 @@ protected function tearDown(): void {
...
+				],
+			],
+		], [ 'mode' => 'error' ] ) );
+	}
+
+	/**
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
+	 */
+	public function test_is_error_or_empty_state_with_empty_mode(): void {
```
chriszarate: There's a lot of states to test here that would be good to cover, maybe with data provider

- error mode returns error
- error mode does not return error but returns no results
- error mode does not return error but returns results
- empty mode returns error
- empty mode does not return error but returns no results
- empty mode does not return error but returns results

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: I had originally just done the mode attribute and avoided the unknown so I'll go back to that.

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: Same thing as above.

src/blocks/remote-data-no-results/render.php
ingeniumed: I fixed this already, but didn't push it up. Will push that up with the rest of the fixes

src/blocks/remote-data-no-results/render.php
ingeniumed: `state` seemed the most appropriate when that method was giving back an error/empty state. After having switched it to true or false, I couldn't quite decide on what the updated name should be as technically it is driving the state of the block. That's why I left it as `state`.

Happy to change it to `should_render`, in which case the method would could potentially be `should_render_no_result_state`

inc/Editor/DataBinding/BlockBindings.php
ingeniumed: Happy to go with this method name too, instead of `should_render_no_result_state`

tests/inc/Editor/DataBinding/BlockBindingsTest.php
ingeniumed: I have added in a bunch more tests in a local commit, but I'm going to spend some time cleaning up the code in this file. There's duplicated mocks, and nearly every test has the 

```
	/**
	 * @runInSeparateProcess
	 * @preserveGlobalState disabled
	 */
```

annotation on it. I'd like to see if in my simplification I can drop that.

tests/inc/Editor/DataBinding/BlockBindingsTest.php
```diff
@@ -205,9 +292,6 @@ public function execute( HttpQueryInterface $query, array $input_variables ): ar
 	 * @preserveGlobalState disabled
 	 */
 	public function test_get_value_with_query_input_transformations_and_overrides(): void {
```
ingeniumed, line 294: This, and test_get_value_with_query_input_transformed_by_custom_query_runner are the only two tests that have the query runner still being mocked inside. I couldn't create a helper function that worked properly for both of em, without being specific to just these two functions. Cursor did suggest trying a callback function but that felt overkill. So I've just left this as is and instead made sure to simplify the rest of the methods.

tests/inc/Editor/DataBinding/BlockBindingsTest.php
```diff
@@ -512,4 +499,73 @@ public function test_get_value_with_null_fallback_results_context(): void {
...
+		$mock_qr->addResult( self::MOCK_OUTPUT_FIELD_NAME, $result );
+		return $mock_qr;
+	}
+
+	/**
+	 * Creates a mock query runner with the specified results.
+	 *
+	 * @param array $results The results to return from the query runner.
+	 */
+	private function create_mock_query_runner_with_results( array $results = [] ): MockQueryRunner {
```
ingeniumed, line 519: This is supposed to be a wrapper around the new function I added to bulk add results, albeit here it's empty results.

tests/inc/Mocks/MockQueryRunner.php
```diff
@@ -31,6 +31,13 @@ public function addResult( string $field, mixed $result ): void {
 		] );
 	}
 
+	public function addResults( array $results ): void {
```
ingeniumed, line 34: This was added to cover the case where the results are empty, but it can also be used in the future to cover the case where we want to provide multiple results. I didn't want to add another if statement to the `addResult` function and make that work, as that just felt ugly. This feels a bit more natural as the idea is that you can bulk add results here. It just happens that the wrapper around this in bulkbindings is feeding it an empty result.

tests/inc/Editor/DataBinding/BlockBindingsTest.php
ingeniumed: I've pushed up a bunch of changes to simplify that tests file as there were a lot of mocks that I was able to consolidate into a set of helper functions. 

Thanks cursor for the suggestions as it made easy based on my prompts.

maxschmeling:
I pushed a commit to use different icons:
image

I mainly just care that they're different, not that we use these particular icons.

maxschmeling:
I also removed the `registerBlockVariation` call because it appears to be handled by the `block.json` from what I can tell.

ingeniumed:
@maxschmeling - thanks for updating the icons, I like them. 

The `registerBlockVariation` was needed as without that, I was running into problems within the block editor. It wasn't behaving consistently at times. I had originally had this registration within the PHP code as well, but that I realized wasn't needed.

ingeniumed:
I did some testing again, and looks like now the problem isn't showing up anymore that required the JS code. That's odd, as before it wasn't the case. Happy to keep it in as it's less duplicated code.

tests/inc/Mocks/MockQueryRunner.php
```diff
@@ -31,6 +31,13 @@ public function addResult( string $field, mixed $result ): void {
...
+		array_push( $this->query_results, [
+			'is_collection' => false,
```
chriszarate: Should `is_collection` be `true` here?

chriszarate (approved):
One question, otherwise looks good. Nice work on the tests.

tests/inc/Mocks/MockQueryRunner.php
ingeniumed: Corrected this

Issue #507, opened 2025-05-05
Getting Undefined array key "service_config" error when using Google Sheets as data source


sethrubenstein:
**Describe the bug**

I haven't tested this with another data source, but on the latest release `0.15.0` and the introduction of `GenericHttpDataSource` seems to have broken Google Sheets as a data source. 

When I revert back to `0.14.0` the issue does not persist. 

**To reproduce**

1. Launch WP playground instance
2. Add Google Sheet credentials and connection to Remote Data Block settings
3. Open new post, open browser console, observe error and lack of matching blocks for new Google source.

**Expected behavior**
1. That the new data source blocks be available and that the error is no longer present

**Screenshots or logs (if applicable)**
```
Undefined array key "service_config"	
wp-content/plugins/remote-data-blocks/inc/Integrations/GenericHttp/GenericHttpDataSource.php:109
```

**Version of the plugin**
0.15.0

**Environment**
WP Playground

ingeniumed:
Thanks for flagging this @sethrubenstein. We have fixed this in #508, and released v0.16.0 with said fix. 

Could you test it out, and let us know if the issue has been fixed for you?

sethrubenstein:
Issue now resolved, thanks.

PR #508, opened 2025-05-06
Add missing config schema to generic http data source

ingeniumed:
### Description

The config schema for `GenericHttpDataSource` hasn't been defined, and so it's pulling the config schema from it's parent that doesn't `service_config` which is why it gets sanitized from data sources leveraging this class. I've extended the parent schema, and added `service_config` to it while ensuring that it's nullable.

I've also added a test for it to ensure that it can be caught next time. A follow up to this would be try and add more integration tests to catch this in an end to end test with a real data source.

### Testing

- Pul down trunk and configure any data source. You'll notice the blocks aren't registering and there are errors in the logs related to a bad `service_config` access.
- Pull down this branch and all of that will now go away.

### DIFF
--- a/tests/inc/Integrations/GenericHttp/GenericHttpDataSourceTest.php
+++ b/tests/inc/Integrations/GenericHttp/GenericHttpDataSourceTest.php
@@ -25,7 +25,7 @@ class GenericHttpDataSourceTest extends TestCase {
 			'service_config' => [
 				'__version' => 1,
 				'display_name' => 'Mock Data Source',
-				'endpoint' => 'http://example.com'
+				'endpoint' => 'http://example.com',
 			],
 		];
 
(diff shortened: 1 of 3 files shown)


shekharnwagh (approved):
Thank you for fixing this !

tests/inc/Integrations/GenericHttp/GenericHttpDataSourceTest.php
```diff
@@ -18,4 +19,22 @@ public function testGetServiceMethodReturnsCorrectValue(): void {
...
+			],
+		];
+
+		// By calling to_array, we can ensure that the config schema is correctly defined,
+		// and that values haven't been sanitized due to them missing from the schema.
+		$data_source = GenericHttpDataSource::from_array( $config );
+		$data_source_array = $data_source->to_array();
+
+		$this->assertEquals( 'generic-http', $data_source_array['service'] );
+		$this->assertEquals( 'http://example.com', $data_source_array['service_config']['endpoint'] );
```
chriszarate, line 38: perhaps more authoritative
```suggestion
		$this->assertEquals( $config['service_config'] , $data_source_array['service_config'] );
```

chriszarate (approved):

ingeniumed:
fixes #507

PR #521, opened 2025-05-08
Update the fallback blocks included with the plugin

ingeniumed:
### Description

The purpose of this PR is to tweak the way the fallback blocks work, based on the design feedback. The following has been done:

- Stop auto-inserting the error block along with the RDB template/container. We want users to insert it themselves.
- Update the error, no results and template block's icons
- Add tracking around the usage of the fallback blocks

CleanShot 2025-05-08 at 11 22 06

### Testing

- Create a new post and insert a RDB block
- Ensure the error and no results blocks aren't auto-inserted
- Insert these blocks, and verify their icons are correct
- Verify the block names are correct as well
- Check the telemetry calls when the post is published as well


inc/Telemetry/Telemetry.php
```diff
@@ -102,7 +102,7 @@ public function track_remote_data_blocks_usage( int $post_id, WP_Post $post ): v
 		}
 
 		// Regular expression to match all remote data blocks present in the post content.
-		$reg_exp = '/<!--\s{1}wp:remote-data-blocks\/([^\s]+)\s/';
+		$reg_exp = '/<!--\s{1}wp:remote-data-blocks\/([^\s]+)\s.*?-->/';
```
ingeniumed: This ensures that we capture the attributes set on the block. The downside of this is that, the giant remoteData attribute is also captured. I haven't been able to effectively use regex to have it skip that, so I've left it like this. Happy to get feedback on this, and see if there's another way to do it.

I had one restriction that I kept in mind - we should only make one pass through the matched blocks. This way, performance is always kept in mind.

inc/Telemetry/Telemetry.php
```diff
@@ -113,7 +113,24 @@ public function track_remote_data_blocks_usage( int $post_id, WP_Post $post ): v
...
 			'post_type' => $post->post_type,
 		];
-		foreach ( $matches[1] as $match ) {
+
+		foreach ( $matches[1] as $index => $match ) {
+
+			// Both fallback blocks show up as no-results, so we need to check the mode attribute to figure out which variation it is.
+			if ( 'no-results' === $match ) {
+				// The error mode attribute is present in the Error block variation.
+				$is_error_block = str_contains( $matches[0][ $index ] ?? '', 'error' );
```
ingeniumed: By default, block variations show up with the same name in the post content. It's only the attributes that differentiate it. So, we need to look up the block in the same index in the second matches entry where the attributes are present. This helps us figure out what type of block it is.

inc/Telemetry/Telemetry.php
```diff
@@ -113,7 +113,24 @@ public function track_remote_data_blocks_usage( int $post_id, WP_Post $post ): v
...
+				if ( $is_error_block ) {
+					$track_props['error_fallback_block_count'] = ( $track_props['error_fallback_block_count'] ?? 0 ) + 1;
+				} else {
+					$track_props['no_results_fallback_block_count'] = ( $track_props['no_results_fallback_block_count'] ?? 0 ) + 1;
+				}
+
+				$track_props['remote_data_blocks_fallback_block_count'] = ( $track_props['remote_data_blocks_fallback_block_count'] ?? 0 ) + 1;
```
ingeniumed: Happy to tweak the names here, just felt like this captured the purpose correctly.

src/blocks/remote-data-no-results/index.ts
```diff
@@ -1,11 +1,23 @@
 import { registerBlockType } from '@wordpress/blocks';
+import { border, caution } from '@wordpress/icons';
 
 import metadata from './block.json';
 import { Edit } from './edit';
 import { Save } from './save';
 import './style.scss';
 
+// @ts-expect-error - WordPress registerBlockType type definition does not include variations even though this is the right way to do it.
```
ingeniumed: The type definitions for this method aren't correct which is why typescript throws an error here. We can provide the variations, and that's needed as without that the icon will not dynamically change between border and caution.

A follow up here would be to figure out how we can update those type definitions once the pause is lifted.

src/blocks/remote-data-pagination/block.json
```diff
@@ -4,7 +4,7 @@
   "name": "remote-data-blocks/pagination",
   "version": "0.1.0",
   "usesContext": [ "remote-data-blocks/remoteData" ],
-  "title": "Remote Data Pagination",
+  "title": "Pagination",
```
ingeniumed, line 7: Changed this to match the other name changes we are making

src/blocks/remote-data-no-results/index.ts
chriszarate: We are already patching these types locally, so you can avoid this error by extending this type:
https://github.com/Automattic/remote-data-blocks/blob/88fa49a1714b9c2f5487e1be644bed7a95d5a2c4/types/wordpress__blocks/index.d.ts#L27

src/blocks/remote-data-template/block.json
```diff
@@ -4,7 +4,7 @@
   "name": "remote-data-blocks/template",
   "version": "0.1.0",
   "usesContext": [ "remote-data-blocks/remoteData" ],
-  "title": "Remote Data Template",
+  "title": "Item Template",
```
chriszarate, line 7: @smithjw1 Is "item" the correct noun? What should we generically call representations of remote data entities?

inc/Telemetry/Telemetry.php
chriszarate: It's fine as long as we're not shipping the attributes

inc/Telemetry/Telemetry.php
chriszarate: > So, we need to look up the block in the same index in the second matches entry where the attributes are present

You are using the first match index which is the entire matched string. We are not actually checking the `mode` attribute here, just checking if `'error'` appears anywhere in the attributes. This is very likely to generate false positives.

Before we go too far down the regex rabbit hole, we should consider:

1. How important is this data? Is it needed? What questions does it answer?
2. Should we just delegate to `parse_blocks`?

inc/Telemetry/Telemetry.php
chriszarate: I believe these new props need to be registered in Tracks.

src/blocks/remote-data-template/block.json
smithjw1: When I see it in context, in Gopal's screenshot, it works for me. Naming is hard so I'm open to other suggestions.

chriszarate: Sounds good then! 👍

inc/Telemetry/Telemetry.php
ingeniumed: > How important is this data? Is it needed? What questions does it answer?

To me, it answers the question - is both the error block and the no-results block necessary? Are users using both, or are they satisfied with just one of them? This would help us determine if we should continue supporting both the blocks or re-think our strategy. I had originally included only a total fallback block count, but then realized this might be an important metric to track which is why I included it as well.

As for how to do this, I had targeted the entire mode with error attribute but then removed that in favour of this as these are all remote data blocks. Happy to add that back in. I think parse_blocks will be more expensive than the regex we have here, but let me see and go from there.

inc/Telemetry/Telemetry.php
ingeniumed: Nope, we def aren't. They are ignored unless its the error block and even then it's just to get the mode value and not to ship it.

inc/Telemetry/Telemetry.php
ingeniumed: Yip, they do. But I wanted to make sure they were named correctly before I did that.

inc/Telemetry/Telemetry.php
chriszarate: > I think parse_blocks will be more expensive

Certainly it will be, but I think is ok to run on `save_post`.

ingeniumed: I did some experimenting on a post with some simple art institute blocks and the results are:

0.0007598400115966797
0.0010700225830078125
0.0010840892791748047
0.0008709430694580078
0.0008130073547363281

0.0009195804595947266s - average when using the new regex

0.00019097328186035156
0.00020003318786621094
0.00019693374633789062
0.00019288063049316406
0.00019097328186035156

0.00019435882568359376 - average when using parse_blocks

`parse_blocks` is faster despite us using a very targeted regex. I should have suspected that, this would be the case with using regex. They have been trying to incorporate more and more of the HTML API to replace regex internally, so I wonder if that's what speeding it up. Either ways, the code is a lot simpler imo and feels safer as well so I'll use parse_blocks.

ingeniumed:
CleanShot 2025-05-09 at 11 17 46

I've included the new icon that @jarekmorawski made for the RDB Container. It's an SVG file, similar to the airtable icon

src/blocks/remote-data-no-results/index.ts
ingeniumed: Added this in, hopefully I've done it correctly

inc/Telemetry/Telemetry.php
ingeniumed: Does that name work? I'll go ahead and register it if it is.

ingeniumed:
I reverted the icon update as something that this doesn't account for is, we do want users to be able to provide their own icon when registering the blocks. This way we would be enforcing what icon it should be.

I think this deserves a short discussion, and so I'll punt this to another PR to avoid keeping this open for too long.

inc/Telemetry/Telemetry.php
```diff
@@ -113,8 +116,16 @@ public function track_remote_data_blocks_usage( int $post_id, WP_Post $post ): v
 			'post_status' => $post_status,
 			'post_type' => $post->post_type,
 		];
-		foreach ( $matches[1] as $match ) {
-			$data_source_type = ConfigStore::get_data_source_type( 'remote-data-blocks/' . $match );
+
+		// This is done this way, to do a one pass processing of the blocks.
+		foreach ( $blocks as $block ) {
```
chriszarate, line 136: This does not recurse the tree and expects remote data blocks at the top level. It will not find nested blocks such as a remote data block inside a group block.

maxschmeling: 553da9ad

chriszarate: Two issues still:

1. Sends a track event even if there are no remote data blocks in the post. We probably want to check if `remote_data_blocks_total_count` is > 0 before sending an event.
1. Minor: Does not find remote data blocks nested in other remote data blocks.

maxschmeling: 06276b0a

chriszarate (approved):

inc/Telemetry/Telemetry.php
ingeniumed: Thanks for pushing the fix to this @maxschmeling.

I'd like to just port the method we have in the block data api to a library and re-use that, as that would be handy in the future.

PR #524, opened 2025-05-08
Disable block controls when multiple blocks are selected

chriszarate:
Selecting multiple blocks in the list view can give the impression that any applied settings during that selection will be applied to all selected blocks.

Screenshot 2025-05-08 at 17 57 18

However, there is no affordance for this behavior and implementing it would likely be very confusing and error prone. Why, for example, would you want to update the query inputs for multiple blocks? If there were multiple query inputs, would such an update mutate them all? Or just one?

Accordingly, this PR disables (hides) our block controls when multiple blocks are selected:

Screenshot 2025-05-08 at 17 56 50

Fixes VIPCMS-91.

### DIFF
--- a/src/block-editor/filters/withBlockBinding.tsx
+++ b/src/block-editor/filters/withBlockBinding.tsx
@@ -114,11 +114,6 @@ export const withBlockBinding = createHigherOrderComponent( BlockEdit => {
 			return <BlockEdit { ...props } />;
 		}
 
-		// If multiple blocks are being selected, render it as usual.
-		if ( hasMultiSelection() ) {
-			return <BlockEdit { ...props } />;
-		}
-
 		// Synced pattern overrides are provided via context and the value can be:
 		//
 		// - undefined (block is not in a synced pattern)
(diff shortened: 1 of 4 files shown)


ingeniumed (changes requested):
CleanShot 2025-05-12 at 07 56 11

If I manually create a pattern, and then use the built in pattern for the other RDB blocks I'm noticing the attributes are being reset to their default state. Using the built-in patterns doesn't cause this error.

src/block-editor/filters/withBlockBinding.tsx
```diff
@@ -102,12 +107,18 @@ export const withBlockBinding = createHigherOrderComponent( BlockEdit => {
...
 		const hasAvailableBindings = Boolean( Object.keys( availableBindings ).length );
+		const { hasMultiSelection } = useSelect< BlockEditorStoreSelectors >( blockEditorStore );
 
 		// If the block does not have a remote data context, render it as usual.
 		if ( ! remoteData || ! hasAvailableBindings ) {
 			return <BlockEdit { ...props } />;
 		}
 
+		// If multiple blocks are being selected, render it as usual.
+		if ( hasMultiSelection() ) {
```
ingeniumed: I think it might be how early this is positioned, that's causing this bug. It might be worth it to move into line 151 check? That has the updated attributes.

chriszarate: Thanks, good spot. Fixed with test in https://github.com/Automattic/remote-data-blocks/pull/524/commits/eeab8c89861be29b1fb445875c8020015f6c2851

ingeniumed: Fix works 💯

ingeniumed (approved):

PR #525, opened 2025-05-09
Miscellaneous UI text improvements for settings screen

chriszarate:
- [Miscellaneous UI text improvements](https://github.com/Automattic/remote-data-blocks/commit/48ee9b246a492c9c827413a858b89d4c92d6b30f) for settings screen, much of which was done during a pairing session with @smithjw1 and @maxschmeling. Fixes VIPCMS-59
- [Fix pattern code example](https://github.com/Automattic/remote-data-blocks/commit/b1e233dfedefe4f97ae1266f21f1dcc2505ca539) (props @smithjw1)
- [Add `htaccess` markers](https://github.com/Automattic/remote-data-blocks/commit/67fc19313c97c9ecb76ff33e99b5f172701f2bba) to prevent repeated mutation by WordPress during development

### DIFF
--- a/src/data-sources/airtable/AirtableSettings.tsx
+++ b/src/data-sources/airtable/AirtableSettings.tsx
@@ -52,7 +52,7 @@ export const AirtableSettings = ( {
 		state.access_token ?? '',
 		userId ?? ''
 	);
-	const { fetchingTables, tables, tablesError } = useAirtableApiTables(
+	const { tables, tablesError } = useAirtableApiTables(
 		state.access_token ?? '',
 		state.base?.id ?? ''
 	);
@@ -145,7 +145,7 @@ export const AirtableSettings = ( {
 	const shouldAllowSubmit =
 		bases !== null && tables !== null && Boolean( state.base ) && Boolean( selectedTables?.length );
 
-	let basesHelpText: React.ReactNode = 'Select a base from which to fetch data.';
+	let basesHelpText: React.ReactNode = 'Select an Airtable base.';
 	if ( userId ) {
 		if ( basesError ) {
 			basesHelpText = __(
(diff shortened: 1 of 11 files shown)


src/data-sources/constants.ts
```diff
@@ -35,15 +35,44 @@ export const GOOGLE_SHEETS_API_SCOPES = [
...
+	none: __( 'Ignored', 'remote-data-blocks' ),
+};
+
+/**
+ * Data source component help text
+ */
+export const HTTP_SOURCE_AUTH_TYPE_HELP_TEXT: Record< HttpAuthTypes, string > = {
+	basic: __( 'The credentials will be sent in an Authorization header.', 'remote-data-blocks' ),
+	'api-key': __( 'Construct a custom HTTP header or query parameter.', 'remote-data-blocks' ),
+	bearer: __( 'The bearer token will be sent in an Authorization header.', 'remote-data-blocks' ),
```
maxschmeling, line 63: Minor minor thing I might be wrong about... I *think* "the Authorization header" is the more correct way to say it... but I'm not actually totally sure.

Take it or leave it.

```suggestion
	bearer: __( 'The bearer token will be sent in the Authorization header.', 'remote-data-blocks' ),
```

Mostly because there can't be more than one `Authorization` header... so it's not `an Authorization header` which implies one of multiple... maybe?

maxschmeling (approved):

src/data-sources/constants.ts
chriszarate: True, there should never be more than one `Authorization` header set. But there can be zero, so I guess the question is whether using `the` implies that an `Authorization` header is already present? Let me consult an oracle.

src/data-sources/airtable/AirtableSettings.tsx
shekharnwagh, line 1: The changes around help text when fetching tables break the existing UX -
- When the second step loads (Base and table configuration screen), by default the Tables help text says - `Fetching Tables...` when its not fetching the tables until a base is selected
- When a base is selected and it actually fetches the table and until the table are fetched it says - `No tables found` instead of saying `Fetching Tables...`. Same is repeated when the selected base is changed and tables are refetched.


https://github.com/user-attachments/assets/bd76f15f-c934-455e-a46f-050a80c61f19

chriszarate: Thanks @shekharnwagh, this is resolved by https://github.com/Automattic/remote-data-blocks/pull/525/commits/284b53da30bc5a2c3fab9887bf055864ccaf90c0.

Instead of an initial message, I opted for omitting help text and relying on the disabled field state, which feels intuitive. I needed to update the base component to disable the "how to" text, which doesn't apply to the way we use this component.

shekharnwagh (approved):

src/data-sources/airtable/AirtableSettings.tsx
chriszarate: Forgot to put the video here. Thanks for looking!

https://github.com/user-attachments/assets/14e7f3cf-aede-4b13-9e25-924b9920946f

PR #530, opened 2025-05-12
Curate queries within a block

ingeniumed:
### Description 

This PR is meant to be a continuation of #405. 

The goal is to allow grouping multiple queries within a block, in the block editor. It'll allow query curation based on the data source within the block editor. So as an example, a couple of queries targetting the products in your shopify site could be grouped in the shopify block saving you the time of inserting these individually.

As part of this the schema was changed:

#### Block Registration

`queries`: This is the list of queries to be registered. There's no need to separate them or to mention their linkage.
`display_queries`: The queries that are display queries from the list of queries provided.

A big big shoutout to @brookewp as her work in #405  is what I'm building on 🙏🏾

In addition:

- Multiple block bindings support was added alongside multiple patterns for each display query
- All deprecation warnings resulting from components have been fixed
- Second migration framework targeting the block configuration has been added. Previously, we only had one for handling individual queries.

### Screenshots

How existing blocks that aren't updated look like:

CleanShot 2025-05-30 at 13 49 00

How new blocks that use multiple display queries look like:

CleanShot 2025-05-30 at 13 49 25

The config object for this block:

CleanShot 2025-05-30 at 13 50 08

### Video Demonstration

Note that the first block doesn't have the artist title while the second block does. It's meant to highlight different outputs schemas for the two display queries.

CleanShot 2025-05-30 at 13 58 46


example/rest-api/art-institute/art-institute.php
```diff
@@ -84,6 +84,8 @@ function register_aic_block(): void {
 				],
 			],
 		],
+		'required_query' => 'search_art',
```
ingeniumed: I've gone back and forth with making this a list of queries, and calling it `required_queries` instead, so it's future proofed. But, the question I have is that what could the use case be? One use case that comes to mind:

- The list of items is derived from a search query
- The image for each item can be derived from another query, as the size can be customized

But is it worth the effort to do that now? We would likely need to tweak our consumption of those queries to account for that, and there would be more changes as a result of it? 

So for now, I just left it like this to keep it simple. Happy to get feedback on this and change it.

ingeniumed:
### Bugs

Documenting any bugs I've encountered that aren't fixed yet so it's known.

- [x] Existing posts aren't loading with this branch. The query selector gets shown instead, even though the underlying data is still available via the code editor.
- [x] Only one schema is registered for the block bindings and that's using the first display query that's found.

example/rest-api/art-institute/art-institute.php
chriszarate: > But is it worth the effort to do that now?

IMO it's not worth it. We don't have defined use cases or feedback requesting this. Even if we did, I would first suggest that users just duplicate the "parent" query and make the necessary adjustments to target the "derived" data.

If the resulting code duplication truly becomes unacceptable, we can work to add an abstraction. But as written here, this abstraction is not intuitive and may cause more confusion.

FWIW, I ran into the same issues with https://github.com/Automattic/remote-data-blocks/pull/371, where the devx of composing two queries together is extremely unintuitive.

example/rest-api/art-institute/art-institute.php
```diff
@@ -202,30 +205,27 @@ function register_aic_block(): void {
...
-			'query' => $get_art_query,
-		],
-		'selection_queries' => [
-			[
-				'query' => $search_art_query,
-				'type' => 'search',
-			],
+		'queries' => [
+			// Changing the name of this query to anything but display will break existing blocks content.
+			'display' => $get_art_query,
```
chriszarate: I would really like us to leave these "magic string" query types like "display" and "search" behind. The query should demonstrate its capabilities, which will eventually transcend a singular "type".

inc/Config/Query/HttpQuery.php
```diff
@@ -62,6 +62,38 @@ public function get_cache_ttl( array $input_variables ): null|int {
...
+	 * @param string $required_query The required query for this query.
+	 */
+	public function set_required_query( string $required_query ): void {
+		$this->config['required_query'] = $required_query;
+	}
+
+	/**
+	 * Get the type of this query.
+	 */
+	public function get_type(): ?string {
```
chriszarate: All of these new public methods are a step sideways IMO. They simply formalize the "magic strings" from the config in a different place. How can we move past the concept of a singular "type" and towards an approach where queries declare support for features that the plugin implements (e.g., search, filtering, pagination, sorting, etc.)

inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -48,84 +92,102 @@ public static function register_block( array $block_config = [] ): bool|WP_Error
...
+
+			// Generate the selector for the query.
+			$selectors[] = [
+				'display_name' => self::get_query_name_from_key( $query_key ),
+				'image_url' => $query->get_image_url(),
+				'inputs' => self::map_input_variables( $input_schema ),
+				'name' => self::DISPLAY_QUERY_KEY === $query->get_type() ? ( $has_required_variables ? 'Manual input' : ( $is_collection ? 'Load collection' : 'Load item' ) ) : ucfirst( $query_key ),
+				'query_key' => $query_key,
+				'type' => $has_required_variables ? 'manual-input' : 'load-without-input',
+				'query_group' => $query_key,
```
chriszarate: I don't think I understand the "query group" concept. What is its purpose?

inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -48,84 +92,102 @@ public static function register_block( array $block_config = [] ): bool|WP_Error
...
+			];
+
+			// If the query is a display query, set the display query found flag.
+			if ( self::DISPLAY_QUERY_KEY === $query->get_type() ) {
```
chriszarate: This feels like a core disconnect for me. _Any_ query provided to the block config can be a display query. The user chooses how to curate data; after having done so, we can look at the output schema and decide which queries are compatible as a display query. Very likely it's just one, but maybe the user should choose again.

example/rest-api/art-institute/art-institute.php
ingeniumed: Agreed on all that, I'll mark this as resolved to indicate that this won't be changed.

inc/Config/Query/HttpQuery.php
ingeniumed: > All of these new public methods are a step sideways IMO. They simply formalize the "magic strings" from the config in a different place. 

I agree with you completely. The approach I've settled on came from experimenting with getting rid of this type, while trying to minimize introducing new fields into the query schema. In trying to get rid of these types, I quickly ran into problems with validation. What happens if a search query doesn't include the right search input parameters? It's not a valid search query anymore, but rather a display query or a selection query if its fed into another display query. The same goes for pagination, etc. 

I was also a bit concerned with the migration process, as our current migration framework handles individual queries but not queries as a whole. In order to figure out how selection queries and display queries are related, or to infer some types we need to look at the queries as a whole. So, a second migration framework was needed and I moved forward with the current approach to get that in place.

Now that it is in place, I can tweak the proposed format easily.

> How can we move past the concept of a singular "type" and towards an approach where queries declare support for features that the plugin implements

I agree that the proposed query format could be improved. Once I get block bindings working with multiple display queries, I'm planning to revisit this. 

Are you suggesting taking a page out of `theme.json` playbook here? Something akin to:

```php
'features"=> [ 'search', 'pagination' ]
```

instead of

```php
'type' => 'display',
```

So each query would mark what features it supports, and we can use that to infer types, do the necessary validation, generate the right selectors and even setup the UI accordingly? 

I like that idea.

inc/Editor/BlockManagement/ConfigRegistry.php
ingeniumed: The idea is that:

`query_key`: Represents a query, regardless of it being a selector or a display query
`query_group`: Represents the display query, and its selector queries. So in the case of the Art Institute, the search art query drive the get art query. So they both are group together to indicate that one depends on the other. Only the display queries are rendered in the block's query picker rather than generating the selector queries as well.

Without this, I was having the case where the search query key was being sent to the RDB's execute query endpoint rather than the display query key (as is the case right now since we only have 1 display query). This would allow it to be dynamic.

I'll be honest, I'm not the biggest fan of it. I have plans to revisit this after the block binding supported multiple display queries. My thought was to see if I could make every query a display and/or selector query and have the query execution make the determination if the display query key should be used or if the selector query key should be used. It could be dictated by the block config or how the user has curated the data.

Happy to hear ideas on how this could be replaced.

inc/Editor/BlockManagement/ConfigRegistry.php
ingeniumed: The core problem is:

How do we decide what query is a display query, and what query is a selector query? 

I tried to answer by having the user mark that down in the query config. But, while trying to fix the block bindings I realized the better question to ask is:

Is the query being executed a display query or a selector query?

This question is better as it implies, it can be dynamically adjusted. Is the search query driving the get art query? Great, it's a selector query. Is the search query not driving any other query but just rendering some results in a block? Great, it's a display query. That's roughly what you are getting at if I'm understanding you and I do agree with it.

src/block-editor/format-types/inline-binding/components/InlineBindingSelectFieldPopover.tsx
```diff
@@ -28,6 +29,15 @@ export function InlineBindingSelectFieldPopover( props: InlineBindingSelectField
 	} );
 	const { remoteData, selectedField, type } = props.fieldSelection;
 
+	// For now, we will use the first compatible selector, but this should be improved.
+	// Same as InlineBindingSelection.tsx
+	const compatibleSelector = getBlockConfig( remoteData?.blockName ?? '' )?.selectors.find(
```
ingeniumed, line 34: I've kept this just like it is when the selector is first picked out for the inline binding selection. I do want to fix this, but I also want to get this working end to end. Once the query format has been finalized, I'll see if I can tweak/fix this.

inc/Editor/BlockManagement/ConfigRegistry.php
chriszarate: > How do we decide what query is a display query, and what query is a selector query?

Why do we need to? Why can't we infer from the query's "features" / capabilities which one it can be?

inc/Config/Query/HttpQuery.php
chriszarate: > So each query would mark what features it supports, and we can use that to infer types, do the necessary validation, generate the right selectors and even setup the UI accordingly?

Yes, but ...

```
'features"=> [ 'search', 'pagination' ]
```

We should not ask the query to declare its compatible features explicitly. Instead, we should determine what behavior our plugin expects in order to implement that feature and ask the query to configure itself accordingly. In fact, search and pagination already work this way:

```php
'input_schema' => [
	'search' => [
		'name' => 'Search terms',
		'type' => 'ui:search_input',
	],
	'limit' => [
		'default_value' => 10,
		'name' => 'Pagination limit',
		'type' => 'ui:pagination_per_page',
	],
	'page' => [
		'default_value' => 1,
		'name' => 'Pagination page',
		'type' => 'ui:pagination_page',
	],
],
```

A query that supports search does not need to declare itself as a "search query"; we can just look at the input schema to determine that. That's information we need anyway, so that we know which input variable should receive the search input from the user. The same is true of pagination.

Queries describe how they work via configuration; we interpret how to use them accordingly. We don't want extra config that can "drift" from other parts of the same config.

ingeniumed: I think the disconnect that I have is trying to validate those queries, when they are meant to be a search or pagination query without having that type on hand before hand. What you are saying is that, if they are configured correctly that way then they'll work that way. So we'd be able to infer everything if it is setup correctly. Otherwise, they won't and the user would have to go in and correct that so it can function properly. 

Did I capture that correctly?

inc/Editor/BlockManagement/ConfigRegistry.php
ingeniumed: From the internal discussion, this would be as simple as them declaring this as

`"display_queries" => [ "get_art" ]`

This does make things a bit more simpler, though we aren't inferring whether a query is a selector or a display query based on what's declared.

example/rest-api/art-institute/art-institute.php
ingeniumed: Note: This is the query slug name. The type itself is gone

inc/Config/Query/HttpQuery.php
ingeniumed: This is now been removed and type inference has been added for search/list.

inc/Editor/BlockManagement/ConfigRegistry.php
ingeniumed: This is gone

inc/Editor/BlockManagement/BlockRegistration.php
```diff
@@ -83,24 +84,41 @@ public static function register_block_configuration( array $config ): array {
...
-				'type' => $mapping['type'],
-			];
+		$available_bindings_for_queries = [];
+		$display_queries_to_selectors = $config['display_queries_to_selectors'] ?? [];
+
+		$patterns = $config['patterns'] ?? [];
+
+		// Using array_keys here triggers a psalm error, so it's set to $_ instead.
+		// Supressing the psalm error is a not a good idea, so instead this is the better solution.
+		// ToDo: Fix the psalm error, and see if array_keys could be used here again.
```
ingeniumed, line 95: I haven't been able to solve a psalm error that complains about the types being invalidated and the key being unused, when `array_keys` was used for the loop. So I gave up, and instead just changed the name of the unused selectors to be such that psalm ignores it. If anyone knows this error, feel free to suggest a fix.

inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -48,84 +90,116 @@ public static function register_block( array $block_config = [] ): bool|WP_Error
...
+				return $inferred_type;
+			}
 
-			foreach ( array_keys( $to_query->get_input_schema() ) as $to ) {
-				if ( ! isset( $from_output_schema['type'][ $to ] ) ) {
-					return self::create_error( $block_title, sprintf( 'Cannot map key "%1$s" from %2$s query. The display query for this block requires a "%1$s" key as an input, but it is not present in the output schema for the %2$s query. Try adding a "%1$s" mapping to the output schema for the %2$s query.', esc_html( $to ), $from_query_type ) );
-				}
+			// If the output schema's type is not an array, then skip selector generation as that'll not work.
+			// This has been done because some output schemas have the type set to string.
+			if ( ! is_array( $output_schema['type'] ) ) {
```
ingeniumed: In order to figure out what selector queries work for what display queries, the output schema of the selector query has to have a type that can be iterated through. That's why this check is present.

I noticed in the tests, there were some cases where the type was set to string and that was it. That prompted me to add this in. If there is a valid case that I should consider here, let me know. I'll update line 160 accordingly.

@chriszarate  - feel like you might be able to chime in here.

inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -48,84 +90,116 @@ public static function register_block( array $block_config = [] ): bool|WP_Error
...
 
-			$config['queries'][ $from_query::class ] = $from_query;
+				continue;
+			}
 
-			$from_input_schema = $from_query->get_input_schema();
-			$from_output_schema = $from_query->get_output_schema();
+			// Infer the type of the query, for selector generation and to validate the non-display queries.
+			$inferred_type = self::infer_query_type( $input_schema, $output_schema );
+			if ( is_wp_error( $inferred_type ) ) {
```
ingeniumed: This inverses the previous way of throwing an error for invalid query based on the type. Now, based on the inputs the type is inferred. So if the inputs aren't setup correctly, then the type won't be inferred and the error would be thrown.

inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -195,4 +303,27 @@ function ( string $slug, array $input_var ): array {
 			array_values( $input_schema )
 		);
 	}
+
+	private static function get_query_name_from_key( string $key ): string {
+		// Replace any non-alphanumeric characters with spaces and convert to title case
+		return ucwords( preg_replace( '/[^a-zA-Z0-9]/', ' ', $key ) );
```
ingeniumed, line 285: This is meant to ensure that the name makes sense when it's used on the UI.

tests/integration/RDBTestCase.php
```diff
@@ -1,6 +1,5 @@
 <?php declare(strict_types = 1);
 
-use WP_UnitTestCase;
```
ingeniumed, line 3: Noticed this warning while running the tests, so removed it.

src/pattern-editor/components/PatternEditorSettingsPanel.tsx
```diff
@@ -48,6 +48,8 @@ export function PatternEditorSettingsPanel() {
 					options={ [ { label: __( 'Select a block' ), value: '' }, ...options ] }
 					onChange={ updateBlockTypes }
 					value={ blockType }
+					__next40pxDefaultSize
```
ingeniumed, line 51: I've been fixing all the deprecation warnings in the console, so this is part of that

src/data-sources/components/DataSourceForm.tsx
```diff
@@ -370,6 +370,7 @@ const DataSourceFormBlocks = ( {
 			}
 		>
 			<ToggleControl
+				__nextHasNoMarginBottom
```
ingeniumed, line 373: I've been fixing all the deprecation warnings in the console, so this is part of that

src/blocks/remote-data-container/components/placeholders/ItemSelectQueryType.tsx
```diff
@@ -1,22 +1,29 @@
...
 export function ItemSelectQueryType( props: ItemSelectQueryTypeProps ) {
-	const {
-		blockConfig: { name: blockName, selectors },
-		onSelect,
-	} = props;
+	const { blockName, selectors, onSelect } = props;
 
 	return (
-		<ButtonGroup className="remote-data-blocks-button-group">
+		<ToggleGroupControl
```
ingeniumed: I've been fixing all the deprecation warnings in the console, so this is part of that

src/blocks/remote-data-container/components/panels/OverridesPanel.tsx
```diff
@@ -53,6 +53,7 @@ export function OverridesPanel( props: OverridesPanelProps ) {
 					key={ override.name }
 					label={ override.display_name || override.name }
 					onChange={ enabled => updateOverrides( override.name, enabled ) }
+					__nextHasNoMarginBottom
```
ingeniumed, line 56: I've been fixing all the deprecation warnings in the console, so this is part of that

src/blocks/remote-data-container/components/panels/DataPanel.tsx
```diff
@@ -44,7 +44,12 @@ export function DataPanel( props: DataPanelProps ) {
 
 	return (
 		<PanelBody title={ __( 'Remote data management', 'remote-data-blocks' ) }>
-			<ButtonGroup>
+			<ToggleGroupControl
```
ingeniumed, line 47: I've been fixing all the deprecation warnings in the console, so this is part of that

ingeniumed:
Note: I'll update the docs in a follow up PR just given the size of this PR.

ingeniumed:
I've cherry picked the deprecation warning fixes into #577

tests/inc/Mocks/MockQuery.php
```diff
@@ -30,6 +30,10 @@ public function preprocess_response( mixed $response_data, array $input_variable
 		return $response_data;
 	}
 
+	public function set_input_schema( array $input_schema ): void {
```
chriszarate: Unused?

inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -23,15 +23,57 @@ class ConfigRegistry {
...
 	public const LIST_QUERY_KEY = 'list';
 	public const SEARCH_QUERY_KEY = 'search';
+	public const QUERIES_KEY = 'queries';
 
 	public static function init( ?LoggerInterface $logger = null ): void {
 		self::$logger = $logger ?? new Logger();
 		ConfigStore::init( self::$logger );
 	}
 
+	private static function migrate_block_config( array $block_config = [] ): array|WP_Error {
```
chriszarate: Seems like we should have actual type validation of block config in a class that extends `ArraySerializable`

inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -23,15 +23,57 @@ class ConfigRegistry {
...
+		// Nothing to migrate, return the block config as is.
+		if ( isset( $block_config[ self::QUERIES_KEY ] ) ) {
+			return $block_config;
+		}
+
+		// if render_query is not set, error out.
+		if ( ! isset( $block_config['render_query'] ) ) {
```
chriszarate: I'm confused, I thought we were moving away from `render_query` towards `queries`.

chriszarate: Oh I see, this is if `queries` is not set.

inc/Editor/BlockManagement/ConfigRegistry.php
chriszarate: I did this in 82b30540

inc/Editor/BlockManagement/ConfigRegistry.php
```diff
@@ -195,4 +264,27 @@ function ( string $slug, array $input_var ): array {
 			array_values( $input_schema )
 		);
 	}
+
+	private static function get_query_name_from_key( string $key ): string {
+		// Replace any non-alphanumeric characters with spaces and convert to title case
+		return ucwords( preg_replace( '/[^a-zA-Z0-9]/', ' ', $key ) );
+	}
+
+	private static function infer_query_type( array $input_schema, array $output_schema ): string|WP_Error {
```
chriszarate: I still don't think it's a good idea to implement a singular "first-one-wins" type

inc/Editor/BlockManagement/ConfigRegistry.php
ingeniumed: Indeed, this is the second migration system targets the entire block config instead of just a single query.

inc/Editor/BlockManagement/ConfigRegistry.php
ingeniumed: Thanks for making that change!

inc/Editor/BlockManagement/ConfigRegistry.php
ingeniumed: What would be a good alternative that you'd prefer, instead for situations like this?

chriszarate: Multiple types

ingeniumed: The upside of the way the types are used is that, we can make this change later on as well. The user provided configs don't need to change, instead it would be the selectors that we generate that would change. So we could add support for that, if we wanted.

That's why it's been done in this way - the user provided types are gone. I'd lean towards leaving it like this, and adding support down the line for multiple types. But we ensure that, the user doesn't need to make any changes for that which I believe this PR does cover. If you do spot an area that prevents that from happening, I can tweak it to change that?

tests/inc/Mocks/MockQuery.php
ingeniumed: Removed in [2110b6](https://github.com/Automattic/remote-data-blocks/pull/530/commits/ff81040c4851f70ef2ae41b127e11d80db2110b6)

chriszarate (changes requested):
Thank you for the great work here! Ultimately, though, I think this selection flow is backwards. In this PR, the user first chooses how they want to _display_ the data (e.g., a query for a single item or a collection of items). Then, based on that choice, they select the data to be displayed using compatible "selection queries." 

Our [design exploration](https://wp.me/pfvsEV-zb), though, is the inverse:

image

The user first chooses how they want to select data: "I want to choose from a list of best-selling products" or "I want to search for one of more items." Once their selection is made, ideally, we automatically know which "display query" to use via inference. If not, we can present the user with a secondary choice.

This is related to my aversion to `display_queries` as a top-level configuration option. I think we can achieve the goals of the design exploration with a single `queries` associative array:

1. Traverse the list of queries and infer their types—search, list, and other future types. Each of these types maps to a selection UI in the block editor. Queries that have a supported type become a top-level option in the placeholder (e.g., "Best Sellers" in the image above). These are our "selection queries."
2. For each selection query, determine which other queries are a compatible display query—that is, their input and output schemas match.
  - If there is only one, we can use it automatically when the user selects data.
  - If there is more than one compatible display query, we display a placeholder with a secondary choice.

There is no need for the user to identify "display queries," mostly because the appropriate display query depends on which "selection query" the user chooses at runtime. And we can infer that information.

chriszarate:
@maxschmeling and I chatted about this for awhile. He interpreted the design exploration a bit differently, which helped me realize that the distinction between "display" and "selection" queries is blurry and ultimately a cause of a lot of confusion for us as plugin developers—and for plugin users as well.

I really strongly feel that we shouldn't perpetuate this distinction in user-facing config or documentation. For example, the "On Sale" query in the screenshot above functions as both a display query (fetch the on-sale products and display them) _and_ a selection query (allow the user to curate filters that are applied to the query):

screenshot_2025-06-05_at_13 31 35_720

Many queries can act in both roles. This helped me realize that `display_queries` as written in this PR is ultimately just controlling the placeholder UI, and we should embrace that. In addition, there are other options that we will likely want to expose. Here is a proposed example:

```php
register_remote_data_block( [
  'name' => 'Product',
  'queries' => [
    'best_sellers' => $best_sellers_query,
    'on_sale' => $on_sale_query,
    'by_id' => $product_by_id_query,
  ],
  'placeholder' => [
    [
      'name' => 'Best Sellers',
      'query_key' => 'best_sellers',
      'icon' => 'store',
    ],
    [
      'name' => 'On Sale',
      'query_key' => 'on_sale',
      'icon' => 'sale-tag',
    ],
  ],
  /* ... */
] );
```

This `placeholder` option takes the place of `display_queries` and controls the order, appearance, and behavior of the placeholder UI choices. (I'm sure there is a better name than `placeholder`, which does not intuitively convey its purpose.)

What's more, this `placeholder` option is completely optional. If omitted, each query in `queries` will be represented in the placeholder UI according to its inferred type. Therefore, the `placeholder` option just gives implementors the ability to (but not obligation) to control the selection flow.

So, ignore what I said about inversion. I think we just need to rename and enhance `display_queries` to become placeholder config.

inc/Editor/BlockManagement/ConfigRegistry.php
ingeniumed: Note: I've punted this to a future PR when we actually leverage this.

ingeniumed:
I've re-done the entire config system based on the feedback provided by @chriszarate. The new config format looks as follows:

```php
	register_remote_data_block( [
		'title' => 'Art Institute of Chicago',
		'icon' => 'art',
		'queries' => [
			// Changing the name of this query to anything but display will break existing blocks content.
			'display' => $get_art_query,
			'search_art' => $search_art_query,
		],
		'placeholders' => [
			[
				'name' => 'Get Art',
				'query_key' => 'display',
			],
		],
	] );
```

I've kept `placeholders` as I couldn't come up with a better name and frankly that works. The placeholders config is entirely option, but if it is provided the `name` and `query_key` are compulsory. If the `placeholders` config isn't provided then every query is a display query and a selector query for any other display query. 

I've also updated the example blocks to use the new config format. I went back and forth for this, as that would reduce the amount of changes in this PR. But, it's better to just isolate all the changes in this PR then. Lemme know if it's better to move this to another PR as I'm happy to do that.

Items I've punted to a future PR that I won't tackle in this PR:

- Adding icon support for the placeholders config. Right now, this is omitted entirely as it'll be optional anyways.
- Adding support for inferring multiple types for a selector. Right now, we have support for one selector only anyways.
- Allowing the picking of the display query or selector query in inline-bindings. Right now just the first display query, and the first compatible selector is picked.
- Fixing the psalm error when `array_keys` is used in some places.

None of these are critical to the merging of this PR and feature imo.

src/blocks/remote-data-container/components/QueryComponent.tsx
```diff
@@ -0,0 +1,188 @@
...
+	);
+}
+
+export function QueryComponent( props: QueryComponentProps ) {
+	const blockProps = useBlockProps( { className: CONTAINER_CLASS_NAME } );
+
+	return (
+		<>
+			<div { ...blockProps }>
+				<EditErrorBoundary blockTitle={ getBlockTitle( props.blockName ) }>
```
ingeniumed, line 178: Is having an ErrorBoundary here, and in the Edit component the right way to handle errors from each component? At the top level, the error from inside QueryComponent wasn't handled, but it was when it was split into two like this. Realistically, does Edit even need an error boundary now? I kept it to be safe but I feel like it could be skipped.

inc/Validation/ConfigSchemas.php
```diff
@@ -86,27 +85,21 @@ private static function generate_remote_data_block_config_schema(): array {
...
 				)
 			),
-			'render_query' => Types::object( [
-				'query' => Types::one_of(
-					Types::instance_of( QueryInterface::class ),
-					Types::serialized_config_for( HttpQuery::class ),
-				),
-			] ),
-			'selection_queries' => Types::nullable(
+			'placeholders' => Types::nullable(
```
chriszarate, line 88: Nitpick: I might lean towards `placeholder` (singular) since it refers to the UI component, not the buttons / actions it contains.

types/localized-block-data.d.ts
```diff
@@ -17,29 +18,33 @@ interface InputVariable {
...
-		inner_blocks?: string;
-	};
-	selectors: {
-		image_url?: string;
-		inputs: InputVariable[];
-		name: string;
-		query_key: string;
-		type: string;
-	}[];
+	patterns: Record< string, string >;
```
chriszarate, line 46: Why did this type widen?

ingeniumed: In order to allow for multiple schemas to be mapped to the block bindings for each query, the patterns had to be differentiated as well. So I store the `query_key` in the keywords, and then map that name to the pattern. Since the `query_key` is dynamic, the types had to be widened.

chriszarate: Since we are allowing multiple schemas to be mapped to a single block (via distinct "display" query keys), I think the block bindings need to store the query key in their source args. That way we know if they are compatible by inspecting the binding (instead of via manually managed keywords). We can enhance the existing `hasBlockBinding` util for this purpose.

ingeniumed: Let me explain what I did a bit more, because we might be talking about two different areas:

- Patterns: This is meant to register different patterns for each display query. Since each query could have a different schema, it's important to differentiate the patterns. So each pattern has the display query stored as the keyword when registering. Then, the patterns are mapped using that query key in `REMOTE_DATA_BLOCKS`. So when trying to show the pattern selection screen, the correct pattern is picked out from `REMOTE_DATA_BLOCKS` and then selected from the pattern store in the block editor. I have a gif up top showing two art institute queries with this setup that would help make that clear.

- Block bindings: This is meant to register different block bindings per query, based on their schema. This is purely for building `REMOTE_DATA_BLOCKS` and shares the code with the pattern registration. In the UI code, this map is useful for performing CRUD operations on the binding based on the query the block is using. This would be the UI sidebar where we pick the fields (can change based on the query's schema) for example.

Hopefully I haven't misunderstood what you mean. The main goal was to translate the user provided block/query config into information that the client could use to actually build out the new placeholder UI feature.

chriszarate: > Patterns: This is meant to register different patterns for each display query. 

Understood here, all that makes sense. But instead of using keywords to maintain a correspondence between patterns and display queries, we should store the query key in the block binding source args. That way the correspondence is data-driven and between the things that actually have the relationship (bindings and query keys). We can use / improve a fn like `hasBlockBinding` to determine if a pattern corresponds to a display query by inspecting the bindings inside of the pattern.

It will also help us highlight (in the UI) situations where bindings "wander" or the display query is changed. In those situations, in the future, we can let the user know that their binding is broken.

> Block bindings: This is meant to register different block bindings per query, based on their schema. This is purely for building REMOTE_DATA_BLOCKS and shares the code with the pattern registration. In the UI code, this map is useful for performing CRUD operations on the binding based on the query the block is using.

I think I'm lost here. There is still only one block binding source that is registered. Its behavior is controlled by the ancestor block context (`remoteData` attribute) and the binding source args.

ingeniumed: I like the idea, as it'll add more inference across the code rather than having the server send out the config to use. 

That said, I don't think this change should be done in this PR. I'd push to punt that to another one. This PR has been open for a bit, it's big enough in its scope and works well enough that I'd like to push off any more changes unless they are really needed. IMO, the proposed changes aren't needed in this PR (note: not saying not at all, just in this PR).

They are part of the future changes stemming from a large number of areas that have been revealed to work the way they do because we assumed 1 query key called `display` would be used. While, I have attempted to change that as much as possible or to the extent possible while keeping the PR from blowing there's a lot still to be done. It's not new features, just QOL improvements imo.

inc/Validation/ConfigSchemas.php
ingeniumed: On one hand, I get where you are coming from since the UI itself is just 1 block. But on the other hand, it feels weird to call it a singular placeholder when several buttons are showing and each of those buttons corresponds to an entirely different query. With the support for multiple display queries, and multiple schemas it's likely even the UI would be different once the query is picked.

ingeniumed:
Ready for review once more

PR #532, opened 2025-05-12
Run E2E and integration tests at minimum version

chriszarate:
Run E2E and integration tests at minimum version instead of latest versions.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -38,8 +38,16 @@ jobs:
           npx playwright install chromium --with-deps
 
       - name: Start up WordPress environment
+        env:
+          WP_ENV_CORE: WordPress/WordPress#6.7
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -38,9 +38,17 @@ jobs:
           npx playwright install chromium --with-deps
 
       - name: Start up WordPress environment
+        env:
```
ingeniumed, line 41: This doesn't have to happen in this PR, but we should consider running against 3 versions:

- Lowest version
- Latest released version
- Trunk

This way we can stay ahead of any potential bugs on all the versions we support, and will support.

ingeniumed (approved):

.github/workflows/e2e-tests.yml
chriszarate: Agreed, just a bummer because these tests are so slow

PR #535, opened 2025-05-13
Round 1 of docs updates

chriszarate:
Round 1 of docs updates designed to fix factual errors, remove context that is unhelpful and/or repetitive and provide more examples that can be of use to AI agents.


chriszarate:
Merging these docs updates to keep iteration going

PR #537, opened 2025-05-14
Update wp-env plugins URLs to be latest, rather than outdated ones

pkevan:
This allow local environments to stay up to date with the latest version of a WP.org plugin.

The change forces a latest version to be used via a seamless 301 redirect on downloads.wp.org, and avoids the need to keep changing this file.


chriszarate (approved):
Nice, thanks!

PR #543, opened 2025-05-15
Rename "field shortcodes" to "inline bindings"

chriszarate:
As discussed in Slack and confirmed with @maxschmeling, rename "field shortcodes" to "inline bindings." This is purely a token replacement with no code changes.

For now, I have not updated the names of the corresponding Tracks events.


maxschmeling (approved):

maxschmeling:
I didn't pull this down and search for instances of field shortcode in the names, but I'm assuming you did. This looks good.

PR #557, opened 2025-05-20
Remove whitespace

pkevan:
Otherwise, the output generate looks like this:

Screenshot 2025-05-20 at 16 59 17


chriszarate (approved):
I think the regex also needs adjustment. The tab before `$data_source =` is getting mangled for some reason. But this fix can be shipped

PR #558, opened 2025-05-22
Data binding: make sure `field_name` is a string

### LINKED ISSUE #556: Remote HTML block: fails when incorrectly configured
## Describe the bug

Block causes fatal error when incorrectly configured.

## Expected behavior

As described in the block itself, when the block isn't configured, nothing should happen.

## Actual behavior

When you insert a `Remote HTML` block into a post, and save it, noting should happen, and you can still edit the post created.

## Reproduction

Steps to reproduce the bug.

1. Create a new post, add some content and then save (so it creates a post ID)
2. Add a `Remote HTML` block to the post
3. Observe the message that the block will be ignored, as it's not configured
4. Save post
5. Refresh the page, and note that a fatal error occurs, and you can no longer edit the post.

### Screenshots or logs

Error displayed:
```
Uncaught Error: RemoteDataBlocks\Editor\DataBinding\BlockBindings::get_block_fallback_content(): Argument #1 ($field_name) must be of type string, null given, called in /var/www/html/wp-content/plugins/remote-data-blocks/inc/Editor/DataBinding/BlockBindings.php on line 293
in /var/www/html/wp-content/plugins/remote-data-blocks/inc/Editor/DataBinding/BlockBindings.php on line 347
```

### Environment

Provide details about your environment:

- Remote Data Blocks plugin version: 1.0
- WordPress version: 6.8
- Local environment, running trunk

pkevan:
Otherwise a fatal error occurs when nothing is configured.

Fixes #556.

### DIFF
--- a/inc/Editor/DataBinding/BlockBindings.php
+++ b/inc/Editor/DataBinding/BlockBindings.php
@@ -280,7 +280,7 @@ class BlockBindings {
 
 		// Extract field information from the binding source args.
 		$field_label = $source_args['label'] ?? null;
-		$field_name = $source_args['field'] ?? null;
+		$field_name = $source_args['field'] ?? '';
 		$field_type = $source_args['type'] ?? 'field';
 		$result_index = $source_args['index'] ?? 0;
 
@@ -300,7 +300,7 @@ class BlockBindings {
 			'remote_data_block_name' => $block_name,
 		];
 
-		if ( null === $field_name ) {
+		if ( empty( $field_name ) ) {
 			self::log_error( $log_context, new WP_Error(
 				'remote_data_blocks_binding_get_value_error',
 				'Missing field mapping for block binding',


chriszarate:
@pkevan This doesn't quite track for me. Where was the fatal being thrown?

```sh
> php -a
Interactive shell

php > $foo = [];
php > $bar = $foo['foo'] ?? '';
php > var_dump( empty( $bar ) );
bool(true)
```

chriszarate:
Ah, I suppose it's when it's passed to `BlockBinding::get_block_fallback_content()`. We could adjust the type hint there, but this works too.

chriszarate (approved):

pkevan:
> Ah, I suppose it's when it's passed to `BlockBinding::get_block_fallback_content()`. We could adjust the type hint there, but this works too.

Apologies - missed that part in the description!

PR #560, opened 2025-05-23
Introduce "block key" identifiers

chriszarate:
This PR addresses a few related issues:

1. Our current approach to creating deterministic "query IDs" is flawed. Since some properties of a query's config can be unserializable (e.g., closures), they are simply dropped from the hash input, resulting in collisions. These IDs are only used to help pagination URL query parameters to find their target block, so we can move to a new strategy with little risk.
2. Our block binding callback function is called by Core for each bound block (e.g., paragraph, image, heading) inside a remote data block, even if it is targeting the same query and query input. While we have network-level caching, the number of redundant operations can take a toll. This was surfaced in our recent performance testing.

This PR introduces a new deterministic ID for remote data blocks based on their block name and query input, which I am tentatively calling a "block key." Remote data blocks with the same "block key" are functionally identical (with respect to our resolution of their remote data):

- They use the same query and input variables for that query.
- They will display the exact same content.

Therefore, we can use this key to address the two issues above:

1. Move from flawed "query ID" to "block key" as the identifier used for pagination URL query parameters.
2. Use the "block key" as an in-memory cache key to improve performance.

This requires some minor changes to the `BlockBindings` class, but is covered by existing tests.


ingeniumed (approved):
I'm good with the PR. My only concern is with the name of the identifier.

I think some variant on `blockID` or `remoteBlockID` might be better than using `blockKey`. Since we use `queryKey` for differentiating the queries, and it's value is set by us for now (until #530 is merged in)   that'd be better.

chriszarate:
> I think some variant on blockID or remoteBlockID might be better than using blockKey. Since we use queryKey for differentiating the queries

Agreed, naming is hard. I specifically didn't choose `*ID` since it's not a unique ID, and IDs in WordPress generally are unique. I guess it's uniquely identifying the config, so it could be `configId`?

chriszarate:
I moved to `configId` in https://github.com/Automattic/remote-data-blocks/pull/560/commits/92e3e8d5bb5f9d58bbffc4113aa2a1b3c2a66ec2

ingeniumed:
I like `configId`, that works 👍🏾

PR #562, opened 2025-05-26
Improve quickstart

chriszarate:
Improve quickstart by [aligning with user intentions](https://github.com/Automattic/remote-data-blocks/blob/d9ac70c8c72bde3d13ff4f7b2077e1a66714b035/docs/quickstart.md) and supplying sample content that makes exploration easier:

Screenshot 2025-05-26 at 15 41 49

This might take a few iterations to get right, since blueprints aren't always easy to test.

Also:

- Moves `releasing.md` into `CONTRIBUTING.md`
- Removes link to `README.md` from [public docs landing page](https://remotedatablocks.com/docs/). It is redundant with the information presented there.


ingeniumed (approved):

pkevan (approved):
Thanks for starting this @chriszarate!

Agree on the difficulty in testing this without merging, so probably makes sense to use this as a first step in improving quickstart.

PR #564, opened 2025-05-30
Fix Block Bindings source label was overridden warning

ingeniumed:
### Description

Currently, there is a warning being spat out by the plugin every time a post is loaded in the editor that looks as follows:

CleanShot 2025-05-30 at 13 30 27

This is because our label is different between the PHP and TS code. So I've united this to be `Remote Data Blocks` just like it is on the PHP code.

### DIFF
--- a/src/block-editor/binding-sources/remote-data-binding.ts
+++ b/src/block-editor/binding-sources/remote-data-binding.ts
@@ -2,7 +2,7 @@ import { registerBlockBindingsSource } from '@wordpress/blocks';
 
 registerBlockBindingsSource( {
 	name: 'remote-data/binding',
-	label: 'Remote Data Binding',
+	label: 'Remote Data Blocks',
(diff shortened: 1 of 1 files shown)


src/block-editor/binding-sources/remote-data-binding.ts
```diff
@@ -2,7 +2,7 @@ import { registerBlockBindingsSource } from '@wordpress/blocks';
 
 registerBlockBindingsSource( {
 	name: 'remote-data/binding',
-	label: 'Remote Data Binding',
+	label: 'Remote Data Blocks',
```
ingeniumed, line 5: Don't mind changing the PHP code instead of Remote Data Binding, but feel like Remote Data Blocks will make more sense to someone than Remote Data Bindings as they know the plugin.

chriszarate (approved):
Nice catch, I thought this was because of our merged attributes.

PR #568, opened 2025-06-02
Allow auth-less Shopify Mock Shop

chriszarate:
A side-effect improvement from my work on better docs and prompts. We can support [Shopify's Mock Shop](https://mock.shop) more directly.

https://github.com/user-attachments/assets/19738129-a914-41f3-9cd7-a95755adf0c7

### DIFF
--- a/inc/Integrations/Shopify/ShopifyDataSource.php
+++ b/inc/Integrations/Shopify/ShopifyDataSource.php
@@ -15,7 +15,7 @@ class ShopifyDataSource extends GenericHttpDataSource {
 	protected static function get_service_config_schema(): array {
 		return Types::object( [
 			'__version' => Types::integer(),
-			'access_token' => Types::skip_sanitize( Types::string() ),
+			'access_token' => Types::nullable( Types::skip_sanitize( Types::string() ) ),
 			'display_name' => Types::string(),
 			'enable_blocks' => Types::nullable( Types::boolean() ),
 			'store_name' => Types::string(),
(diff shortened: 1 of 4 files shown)


inc/Integrations/Shopify/ShopifyDataSource.php
```diff
@@ -15,21 +15,32 @@ class ShopifyDataSource extends GenericHttpDataSource {
...
 			'store_name' => Types::string(),
 		] );
 	}
 
 	protected static function map_service_config( array $service_config ): array {
+		// Extract the store name from the full URL, if provided.
+		$store_name = preg_replace( '#^https?://#', '', $service_config['store_name'] );
+		$store_name = preg_replace( '#\.myshopify\.com.*#', '', $store_name );
+
+		$endpoint = sprintf( 'https://%s.myshopify.com/api/2024-04/graphql.json', $store_name );
```
ingeniumed, line 30: Is there a way we could abstract the logic for this mock store on the backend and frontend, for re-usability across data sources? I was thinking if it could be done that way, then the same functions could be called on both the backend and frontend to handle this case. Right now, I could see each data source that does require this needing to implement this.

At the moment, it seems like Shopify is the only with a mock store but maybe we might be interested in providing one for the other data sources. We could then leverage this abstraction and perhaps offer it as an option to pick when setting up your data sources just so you can test out the plugin.

chriszarate: I would love for all API calls to transact over the REST API. That would enable a lot of code removal. However, it would be a bigger undertaking. I believe there is an issue filed for this work.

shekharnwagh (approved):

PR #569, opened 2025-06-02
Improve examples, introduce templates, and default to array config

chriszarate:
This reorganizes example code and updates docs to be more AI-friendly:

1. Examples that can be run directly without any setup steps or credentials (e.g., blocks that use auth-less APIs) live in `example/`. They can be required by uncommenting code in `remote-data-blocks.php` and be ready to use immediately.
2. Examples that require setup steps or credentials are converted to templates and live in `example/templates/`. These templates are heavily annotated and contain `{{ Placeholders }}`, making them much more AI-friendly.
3. Documentation is generally updated to use "bare array config" instead of `::from_array()` methods. This is a friendlier default and where we want our dev-ex to be.
4. Miscellaneous documentation updates.

No code changes.


pkevan (approved):
👍 only spotted on spelling error (I think), so committed it in [5c37f55](https://github.com/Automattic/remote-data-blocks/pull/569/commits/5c37f55e18c9df953cd55c12861eb6fdc4b5fa3c)

chriszarate:
> 👍 only spotted on spelling error (I think), so committed it in [5c37f55](https://github.com/Automattic/remote-data-blocks/pull/569/commits/5c37f55e18c9df953cd55c12861eb6fdc4b5fa3c)

@pkevan "optiona" is the Latin plural 😆 . Thanks for the fix!

PR #571, opened 2025-06-03
Pass `$raw_response_data` to `generate` functions

chriszarate:
Pass `$raw_response_data` to `generate` functions in order to give them access to the query's input variables, response metadata, and other future contextual or run-time data that is not available in scope when the `generate` function is defined.

### DIFF
--- a/docs/extending/query-output-schema.md
+++ b/docs/extending/query-output-schema.md
@@ -7,7 +7,9 @@ Note that the output schema may require updates whenever the shape or schema of 
 ## Properties
 
 - `format` (optional): A callable function that formats the output variable value.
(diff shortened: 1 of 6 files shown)


maxschmeling (approved):
I think `raw_response_data` is a bit strange for the variable where you get `input_variables`, but it's consistent with the naming we have already so it works for me.

chriszarate:
> I think raw_response_data is a bit strange for the variable where you get input_variables, but it's consistent with the naming we have already so it works for me.

Just the name? Suggestions welcome. The query runner flow is a little shaky and maybe deserves attention down the line, but I'm hesitant to move things around right now.

PR #573, opened 2025-06-04
Add optional query display name

chriszarate:
Queries do not currently accept a display name, but I believe a display name will be useful in the future. Asking for it now (optionally) might pay off in the future.

### DIFF
--- a/inc/Validation/ConfigSchemas.php
+++ b/inc/Validation/ConfigSchemas.php
@@ -156,6 +156,7 @@ final class ConfigSchemas {
 
 	private static function generate_http_query_config_schema(): array {
 		return Types::object( [
+			'display_name' => Types::nullable( Types::string() ),
 			'cache_ttl' => Types::nullable( Types::one_of( Types::callable(), Types::integer(), Types::null() ) ),
 			'data_source' => Types::one_of(
 				Types::instance_of( HttpDataSource::class ),
--- a/example/blocks/art-block/art-block.php
+++ b/example/blocks/art-block/art-block.php
@@ -20,6 +20,7 @@ function register_art_remote_data_block(): void {
 	];
 
 	$get_art_query = [
+		'display_name' => 'Get artwork by ID',
 		'data_source' => $aic_data_source,
 		// Provide a callable (closure) to dynamically generate the endpoint using
 		// the base endpoint from the data source and the input variables.
@@ -80,6 +81,7 @@ function register_art_remote_data_block(): void {
 	];
 
 	$search_art_query = [
+		'display_name' => 'Search artworks',
 		'data_source' => $aic_data_source,
 		// Provide a callable (closure) to dynamically generate the endpoint using
 		// the base endpoint from the data source and the input variables.
--- a/example/blocks/github-markdown-block/github-markdown-block.php
+++ b/example/blocks/github-markdown-block/github-markdown-block.php
@@ -39,6 +39,7 @@ function register_github_markdown_remote_data_block(): void {
 	$file_extension = '.md';
 
 	$get_file_as_html_query = [
+		'display_name' => 'Get GitHub Markdown file as HTML',
 		'data_source' => $github_data_source,
 		// Provide a callable (closure) to dynamically generate the endpoint using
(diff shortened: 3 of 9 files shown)


inc/Validation/ConfigSchemas.php
```diff
@@ -156,6 +156,7 @@ private static function generate_generic_http_data_source_config_schema(): array
 
 	private static function generate_http_query_config_schema(): array {
 		return Types::object( [
+			'display_name' => Types::nullable( Types::string() ),
```
ingeniumed, line 159: Thoughts around making it mandatory? That way we don't need to worry about fallbacks, and it does force users to think about the purpose of the query.

This'll make some of the code I added in #530 redundant which is great as we shouldn't be generating that name ideally.

ingeniumed (approved):
Never been so eager to review a PR such as this one

pkevan:
I don't know if this has been thought about yet, but how are we planning to use internationalization within the plugin? 

I guess display name is an easy one to start with, although if the input is via a user then maybe that isn't important to them, but it might be for end users if it's display somewhere user facing.

inc/Validation/ConfigSchemas.php
chriszarate: It's a breaking change. That's the main reason I made it nullable.

chriszarate:
> I don't know if this has been thought about yet, but how are we planning to use internationalization within the plugin?

We've thought about it a little:

1. For plugin code, we make an effort to localize our own strings with a consistent text domain. It undoubtedly needs attention and improvements.
2. By design, we ask plugins users to define their configuration at runtime (e.g., `register_remote_data_block( /* big configuration blob */ )`, so they are free to localize strings within that config, including display names and other user-facing content.
3. For data returned from remote data sources, I don't know if we have a strong opinion. Internationalized APIs might be controlled by parameters that could be adjusted at runtime (e.g., [via a filter](https://github.com/Automattic/remote-data-blocks/blob/5c56ad1bd77b2573210be61ca104da4276af4ed5/docs/extending/hooks.md#remote_data_blocks_query_input_variables)) depending on the user's location or language preference. Or plugin users could use `generate` to perform localization themselves. I don't know if there is a strong need for it.

inc/Validation/ConfigSchemas.php
ingeniumed: That's a fair point, and I think I have a solution from it. We can leverage the migration framework that'll be introduced in #530 to generate the display name in the same way that the PR does already. That way, the nullable could be dropped while not breaking anything already existing.

This can happen in a follow up PR

pkevan:
> > I don't know if this has been thought about yet, but how are we planning to use internationalization within the plugin?
> 
> We've thought about it a little:
> 
> 1. For plugin code, we make an effort to localize our own strings with a consistent text domain. It undoubtedly needs attention and improvements.
> 2. By design, we ask plugins users to define their configuration at runtime (e.g., `register_remote_data_block( /* big configuration blob */ )`, so they are free to localize strings within that config, including display names and other user-facing content.
> 3. For data returned from remote data sources, I don't know if we have a strong opinion. Internationalized APIs might be controlled by parameters that could be adjusted at runtime (e.g., [via a filter](https://github.com/Automattic/remote-data-blocks/blob/5c56ad1bd77b2573210be61ca104da4276af4ed5/docs/extending/hooks.md#remote_data_blocks_query_input_variables)) depending on the user's location or language preference. Or plugin users could use `generate` to perform localization themselves. I don't know if there is a strong need for it.

Thanks! That makes sense - I wonder if we should provide an example of this? 

I'm more thinking around the use-case whereby the plugin is configured across many sites (maybe not via the plugin itself but some mu-plugin code), and these sites have different languages - we should help them in these scenarios somehow.

chriszarate:
> I wonder if we should provide an example of this?
>
> I'm more thinking around the use-case whereby the plugin is configured across many sites (maybe not via the plugin itself but some mu-plugin code), and these sites have different languages - we should help them in these scenarios somehow.

I think 1 and 2 should solve for this, with some additional effort. I created VIPCMS-1316 to track this

PR #574, opened 2025-06-04
Switch to the public playground url

ingeniumed:
### Description

Now that the public playground instance has been updated, and is the official one that we are supporting it's time to switch to that. This PR does exactly that.


pkevan (approved):

docs/quickstart.md
```diff
@@ -2,11 +2,11 @@
 
 ## I want to explore the plugin without configuring anything.
 
-[Launch the plugin in WordPress Playground](https://wordpress-playground.atomicsites.blog/?blueprint-url=https://raw.githubusercontent.com/Automattic/remote-data-blocks/trunk/blueprint.json) and explore. An example remote data block ("Conference Event") has been registered.
+[Launch the plugin in WordPress Playground](https://playground.wordpress.net/?blueprint-url=https://raw.githubusercontent.com/Automattic/remote-data-blocks/trunk/blueprint.json) and explore. An example remote data block ("Conference Event") has been registered.
```
pkevan, line 5: I wonder if this should take the user into a post page with content being displayed, rather than the RDB config screen?

maxschmeling: I think a better "next steps" on the RDB page is probably ideal. Maybe something that is aware that it's launching from a new blueprint. But we can improve that type of thing as we have time down the road.

chriszarate: > I wonder if this should take the user into a post page with content being displayed, rather than the RDB config screen?

@pkevan @maxschmeling It already does that via [this PR](https://github.com/Automattic/remote-data-blocks/pull/562). Can that content be improved? It's a simple wxr file hosted in `example/assets/`

pkevan: Probably worth merging that PR if it improves it, then iterate?

chriszarate: @pkevan I merged it last week.

PR #579, opened 2025-06-04
v1.0.1

maxschmeling:


shekharnwagh:
@maxschmeling We have enough [commits](https://github.com/Automattic/remote-data-blocks/compare/v1.0.0...trunk) in this to justify this being minor release instead of a patch release. What do you think ?

maxschmeling:
> @maxschmeling We have enough [commits](https://github.com/Automattic/remote-data-blocks/compare/v1.0.0...trunk) in this to justify this being minor release instead of a patch release. What do you think ?

There are quite a few commits, but there isn't much in the way of new functionality. Could maybe be a minor release, but I think it's fine as a patch.

PR #580, opened 2025-06-05
Add example weather block

chriszarate:
Provide an example weather block (generated by `claude-4-sonnet`). Notably, this example makes an initial API request to translate a city name to lat/long.

Screenshot 2025-06-05 at 13 45 28

### DIFF
--- a/remote-data-blocks.php
+++ b/remote-data-blocks.php
@@ -69,4 +69,5 @@ do_action( 'remote_data_blocks_loaded' );
 // require_once __DIR__ . '/example/blocks/art-block/art-block.php';
 // require_once __DIR__ . '/example/blocks/github-markdown-block/github-markdown-block.php';
 // require_once __DIR__ . '/example/blocks/shopify-mock-store-block/shopify-mock-store-block.php';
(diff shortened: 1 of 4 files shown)


remote-data-blocks.php
```diff
@@ -69,4 +69,5 @@
 // require_once __DIR__ . '/example/blocks/art-block/art-block.php';
 // require_once __DIR__ . '/example/blocks/github-markdown-block/github-markdown-block.php';
 // require_once __DIR__ . '/example/blocks/shopify-mock-store-block/shopify-mock-store-block.php';
+// require_once __DIR__ . '/example/blocks/weather-block/weather-block.php';
```
ingeniumed, line 72: This new way of enabling and disabling blocks is a nice improvement over the previous way. That said, I wonder if we could improve it further by offering a toggle within the settings screen to enable/disable this? We could list out the available example blocks, and potentially link to some documentation about these blocks as well. 

Stretch goal: That would open the door to potentially having special VIP only blocks that could be controlled via the integrations screen too.

This can happen in another PR, but I can see this be a good QOL improvement.

ingeniumed (approved):

PR #585, opened 2025-06-06
Updating links to patterns

stoph:
The Patterns section was removed from Core Concepts in 6dff245 
Patterns are now described in extending/block-patterns.md


chriszarate (approved):
👍 Thanks!

PR #586, opened 2025-06-06
Improve documentation around supported data sources and next steps

chriszarate:


docs/extending/data-source.md
```diff
@@ -2,9 +2,36 @@
...
+
+A data source that has been defined in the plugin settings screen can be loaded by its UUID using the `HttpDataSource::from_uuid()` static method. The UUID is provided via the actions (three-dot) menu. This approach allows you to write code to define [queries](query.md) and [register blocks](block-registration.md) to complete your integration.
+
+```php
+$data_source = HttpDataSource::from_uuid( '{{ Data source UUID }}' );
+
+/* Additional code to use the data source in queries and block registration */
+```
+
+Unsupported data sources, as well as data sources that require customization not offered in the UI, must be defined in code.
```
smithjw1, line 32: I'm not sure "unsupported" is the best word here. It sounds like you can't do it. Or that we will help you troubleshoot your wonky Google auth issues since that's a "supported" data source.

I realize naming is hard.

I've generally used "Named," which is also problematic, because your HTTP data source has a name, too. But it avoids the issues above.

"Included" or "Provided" might be helpful. Does "Standard" and "Custom" work any better?

If we can't find something we like more, "Supported" isn't a miss, but maybe we can land on something better.

chriszarate: - built-in
- managed
- ui-managed
- assisted
- native
- automatic

smithjw1: I like "Built In" and "Custom Built"

chriszarate: Updated in https://github.com/Automattic/remote-data-blocks/pull/586/commits/f49c7c3230058b21b2bb6d10dbdfbfd0929c3fb6

shekharnwagh (approved):

PR #593, opened 2025-06-10
Add MCP server

chriszarate:
Adds an MCP server via [WP Feature API](https://github.com/Automattic/wp-feature-api) + [WordPress MCP](https://github.com/Automattic/wordpress-mcp). This allows us to implement custom functionality and expose it to agentic tools. Here is GitHub Copilot Chat in VS Code using MCP to list the data sources in my local environment, with UUIDs and code snippets:

https://github.com/user-attachments/assets/13540069-350a-4d55-af0f-7b632b842f67

This was an experiment to make sure MCP was possible, but I additional strong use cases don't come immediately to mind. What would you like to see integrated with MCP?

## Try it

1. Restart your local environment to pick up the new plugins defined in `.wp-env.json`.
2. Go to Settings > MCP > Authentication and create a JWT token.
3. Start the MCP server inspector:

```sh
npx @modelcontextprotocol/inspector \
  -e WP_API_URL=http://localhost:8888/ \
  -e JWT_TOKEN=your-jwt-token-here \
  npx @automattic/mcp-wordpress-remote@latest
```

4. Hit "Connect".
5. 4. Navigate to "Tools" and "List tools".
6. Select the `wp_feature_list-remote-data-block-data-sources` tool, and run.


maxschmeling:
Closing this out as an old experiment so the open PR queue reflects active work. If MCP becomes a priority, we should restart from current trunk and the current WP MCP/Feature API landscape with a clear use case.

PR #594, opened 2025-06-11
Add an extra filter to fix the memory caching bug in Google Sheets

ingeniumed:
### Description

Found a bug in RDB using Google Sheets as a data source - you can't correctly select more than 1 entry from the DataViews modal. It'll always replicate the entry. It's due to [the get_request_details](https://github.com/Automattic/remote-data-blocks/blob/trunk/inc/Config/QueryRunner/QueryRunner.php#L228) method returning the same object for all the entries. That causes the in memory cache key to be the same and thereby return the same result. I haven't been able to replicate this in any other data sources, because they include some different detail in the url, etc. For Google Sheets we just get all the data and then filter out the entries we don't want.

Instead of trying to fix the request details method, I took advantage of the filter that we provide to enhance it instead. This way, if there are other data sources that run into this problem they could take on a similar approach. If we do see that this happens often, we can then add in a fix for the future.

### Testing

- Setup a Google Sheets Data Source
- Ensure that when you select more than 1 entry, they are not duplicated

### DIFF
--- a/inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php
+++ b/inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php
@@ -11,6 +11,7 @@ use WP_Error;
 class GoogleSheetsIntegration {
 	public static function init(): void {
 		add_action( 'init', [ __CLASS__, 'register_blocks' ], 10, 0 );
+		add_filter( 'remote_data_blocks_request_details', [ __CLASS__, 'enhance_request_details' ], 10, 3 );
 	}
 
 	public static function register_blocks(): void {
@@ -185,5 +186,29 @@ class GoogleSheetsIntegration {
 		}
 
 		return $snippets;
+	}
+
+	/**
+	 * Due to the fact that we are using the same query for both the get and list queries, and
+	 * only filtering out the results based on the input variables, the in-memory cache will not
+	 * work as expected. This enhances the request details to include the input variables, so that
+	 * the in-memory cache will be able to differenciate each request.
+	 *
+	 * @param array<string, mixed> $request_details The request details.
+	 * @param string $query The query being executed.
+	 * @param array<string, mixed> $input_variables The input variables for the current request.
+	 * @return array<string, array{
+	 *   method: string,
+	 *   options: array<string, mixed>,
+	 *   origin: string,
+	 *   uri: string,
+	 * }>
+	 */
+	public static function enhance_request_details( array $request_details, string $query, array $input_variables ): array {
+		if ( isset( $request_details['origin'] ) && 'https://sheets.googleapis.com' === $request_details['origin'] && ! empty( $input_variables ) ) {
+			$request_details['input_variables'] = $input_variables;
+		}
+
+		return $request_details;
 	}
 }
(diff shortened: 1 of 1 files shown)


inc/Integrations/Google/Sheets/GoogleSheetsIntegration.php
```diff
@@ -186,4 +187,28 @@ public static function get_block_registration_snippets( array $data_source_confi
...
+	 * @param string $query The query being executed.
+	 * @param array<string, mixed> $input_variables The input variables for the current request.
+	 * @return array<string, array{
+	 *   method: string,
+	 *   options: array<string, mixed>,
+	 *   origin: string,
+	 *   uri: string,
+	 * }>
+	 */
+	public static function enhance_request_details( array $request_details, string $query, array $input_variables ): array {
```
ingeniumed: I've taken this approach on purpose. I didn't want to alter the request_details method as it's technically working as expected in this case. The actual request doesn't really have any enhancements and we really shouldn't have data source specific logic in there. 

Doing it this way means we let the integration handle it. Happy to have a discussion on this.

pkevan:
Is this bug present in both in memory and object caching setup? I couldn't see from the bug description as to where this wasn't working.

shekharnwagh (approved):

ingeniumed:
> Is this bug present in both in memory and object caching setup? I couldn't see from the bug description as to where this wasn't working.

It's present in our in-memory caching setup. We use the request details to do an in-memory cache so that it would be benefit repeated requests to the exact same data source. Since Google Sheets queries don't include the inputs in their request, and instead process the response after using those inputs the in-memory cache key ends up being the same. This filter adds the input variables to the request details thereby forcing the cache key to be different.

chriszarate:
@ingeniumed I think this highlights a footgun that we should remove. Passing `$input_variables` to `preprocess_response` is a mistake, because it allows you to implement non-deterministic processing that ends up poisoning the in-memory cache.

maxschmeling:
> @ingeniumed I think this highlights a footgun that we should remove. Passing `$input_variables` to `preprocess_response` is a mistake, because it allows you to implement non-deterministic processing that ends up poisoning the in-memory cache.

To be fair, `preprocess_response` can return non-deterministic results either way. Passing input into it probably increases the chances, but it does provide for more flexibility. Definite footgun, but I'm not 100% convinced yet that it should be removed.

Should we log an issue and have discussion about that elsewhere?

chriszarate:
> To be fair, preprocess_response can return non-deterministic results either way.

Only with random input that you implement yourself or that is provided by the API response. Everything else is static with respect to the current request. I can log an issue.

chriszarate:
VIPCMS-1347

ingeniumed:
To me, the question would be - how should the data source that uses preprocess_response to filter out the response fulfil it's required request? That's how Google Sheets [works right now](https://github.com/Automattic/remote-data-blocks/blob/trunk/inc/Integrations/Google/Sheets/GoogleSheetsDataSource.php#L93). We could say in this specific case, Google Sheets API does allow you to use cell ranges to be able to pick out specific values. It's a bit tedious to use that, compared to just getting the values and filtering it out but it's there. 

But, that's not always going to be the case with an API. There are lots of APIs that are poorly designed but still need to be used and in those cases it's handy to have the input_variables on hand.

I think the real question is - how could we prevent the in-memory cache from being poisoned? Could we restrict what items we use from request-details for generating the cache key? Could we restrict the changes allowed in the filter?

PR #606, opened 2025-06-23
updating documentation to reference IC

smithjw1:
We don't want to include complete documentation on enabling the plugin on VIP, but we do want to guide customers away from a direct installation.

This draft is because the Integration Center isn't live yet in the Dashboard. This PR can be merged as soon as it is.


yolih (approved):
LGTM!
I think that's the best Docs link to share in this situation too.

PR #612, opened 2025-07-11
Avoid fatals when invalid row selected

### LINKED ISSUE #611: Google sheets integration can break editor via user input
## Describe the bug

When using the Google sheets integration, it's relatively straightforward to break it, and thereby lock a user out of the editor.

## Expected behavior

Any user input should not allow errors to be created and the editor should remain usable.

## Reproduction

Steps to reproduce the bug.

1. Create an integration with Google sheets
2. Input integration into post editor with several rows selected
3. View valid output within the editor
4. Adjust one of the row IDs to be either `0` or completely blank
5. Save updated data
6. See fatal error
7. Can no longer edit post when returning to the post, or refreshing the page

pkevan:
Fixes #611.

### DIFF
--- a/inc/Integrations/Google/Sheets/GoogleSheetsDataSource.php
+++ b/inc/Integrations/Google/Sheets/GoogleSheetsDataSource.php
@@ -91,13 +91,13 @@ class GoogleSheetsDataSource extends GenericHttpDataSource {
 	}
 
 	public static function preprocess_get_response( array $response_data, ?string $row_id = null ): array {
-		$selected_row = null;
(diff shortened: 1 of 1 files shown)


shekharnwagh (approved):
It fixes the issue as it no longer crashes the editor. Though if you set one of the `rowId` as 0, it updates in editor as empty, but on frontend it still shows the previously selected row data. Maybe that's an acceptable tradeoff to fall back to last known non-empty value, as that's what we do with the remote data stored in post content.

PR #632, opened 2025-08-06
Add/allow explicit block name

### LINKED ISSUE #631: Allow block name to be different than block title
When registering a new block, the block name is generated based on the block title/label. This means that if we ever decide to rename the block, this would break all instances of it.

It would be great if we could explicitly set the block name on block registration, so we have the ability to change the block title in the future without breaking existing instances of the block.

As you know, naming is one of the most complex tasks in computer sciences, so having to stick with the name chosen in the early stages of a block development can be very limiting.

leogermani:
This PR allows the block name to be set upon block registration, so it can be different and independent of the block title.

## Testing

Make sure existing blocks being registered without an explicit block name are still working the same way they were.

Register a new block and inform a different `name` in its configuration. 
* Confirm that the block works as expected
* confirm that `remote-data-blocks/` prefix is added to the block name and that special characters in the names are sanitized and no additional slashes are added.

PHP tests were generated by AI, sorry if they are too verbose.

Fixes #631

### DIFF
--- a/inc/Editor/BlockManagement/ConfigRegistry.php
+++ b/inc/Editor/BlockManagement/ConfigRegistry.php
@@ -41,10 +41,14 @@ class ConfigRegistry {
 			return $validated;
 		}
 
-		// Check if the block has already been registered.
 		$block_title = $block_config['title'];
-		$block_name = ConfigStore::get_block_name( $block_title );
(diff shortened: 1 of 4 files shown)


maxschmeling:
This looks great. AI is so helpful at writing tests 😄  Thanks!

maxschmeling (approved):

PR #633, opened 2025-08-12
fix: update how the Remote Data Blocks pattern category is added in the plugin

laurelfulford:
This PR fixes the formatting for how the Block Pattern Category "Remote Data Blocks" was being added. 

Prior to this change, any patterns created for the Remote Data Blocks were showing up under 'Uncategorized'.

## Steps to test:
1. Start on a test site with at least one Remote Data Block with a pattern that can be selected.
2. Open the Block Inserter in the editor, click the Patterns tab and note there's no "Remote Data" category. All your Remote Data patterns are appearing under "Uncategorized".
3. Apply the PR and confirm the patterns are now appearing under "Remote Data Blocks".

There were originally both a "Remote Data" and a "Remote Data Blocks" pattern categories listed; I consolidated them down to "Remote Data Blocks" but can split them back into two categories again!

### DIFF
--- a/inc/Editor/BlockManagement/ConfigRegistry.php
+++ b/inc/Editor/BlockManagement/ConfigRegistry.php
@@ -153,7 +153,7 @@ class ConfigRegistry {
 		// Create the pattern properties, allowing overrides via pattern options.
 		$pattern_properties = [
 			'blockTypes' => [ $block_name ],
-			'categories' => [ 'Remote Data' ],
+			'categories' => [ 'remote-data-blocks' ],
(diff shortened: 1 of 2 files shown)


ingeniumed (approved):
This is a nice change that'll ensure that our users can easily look up the patterns tied to our plugin. Thanks for making it!

A follow up task for us that I'm noting down:

Our patterns don't have a preview so it looks like this in the sidebar: 

CleanShot 2025-08-13 at 13 03 49

It'd be good for us to take care of that so it doesn't look like that.

Issue #634, opened 2025-08-12
Remote data block appears unresponsive during data fetching after item selection


rbcorrales:
When using the remote data block in the editor, there's a confusing user experience during data fetching:

### Steps to reproduce
1. User clicks the "Choose" button to open a modal with data items.
2. User selects an item from the modal.
3. The modal closes, but there's a delay of a few seconds while it's still loading before the pattern selection step appears.
4. During this delay, there's no visual indicator that something is happening.
5. During this window, the "Choose" button remains clickable, allowing users to open the modal again.
6. If the background process finishes while the modal is open, it abruptly closes the modal.

### Expected behavior
- Clear visual indication that data is being fetched.
- The interface should prevent duplicate actions during loading.

### Actual behavior
- The user might think the interface is unresponsive.
- The user may click "Choose" multiple times.
- Sudden modal closures create poor UX when background processes complete.

leogermani:
Fixed by #635

PR #635, opened 2025-08-12
Add loading overlay for initial data selection

rbcorrales:
Adds a loading overlay that appears when users select data in the remote data block, providing visual feedback during the fetch process.

Addresses https://github.com/Automattic/remote-data-blocks/issues/634.

## Changes
- Reused existing overlay rendering logic and refactored it into a helper function.
- Noticed that the existing overlay allows click interactions, so added a param to conditionally apply `pointer-events: 'auto'` for blocking interactions during placeholder loading and prevent button interactions.

image

## Testing

1. Insert a remote data block in the editor.
2. Click the "Choose" button to open the data selection modal.
3. Select an item from the modal and confirm/save.
4. Verify that the modal closes and a translucent loading overlay appears over the placeholder.
5. Verify that clicking anywhere on the placeholder area during loading has no effect (overlay should block clicks).
6. Verify that the loading overlay disappears when the pattern selection buttons appear or blocks are inserted.
7. Refresh the editor page and make sure that there are remote data blocks added with data.
8. During load, verify that the spinner and overlay keep appearing, but you can still interact with the components behind it.

### DIFF
--- a/src/blocks/remote-data-container/edit.tsx
+++ b/src/blocks/remote-data-container/edit.tsx
@@ -101,9 +101,30 @@ function RemoteDataBlockEdit( props: BlockEditProps< RemoteDataBlockAttributes >
 		refreshRemoteData();
 	}
 
+	function renderLoadingOverlay( isClickable = false ): JSX.Element {
+		return (
+			<div
+				className="remote-data-blocks-loading-overlay"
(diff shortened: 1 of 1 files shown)


ingeniumed (approved):

maxschmeling:
Good improvement. Thanks for the contribution!

PR #644, opened 2025-08-28
feat: improve modal selection; update buttons

thomasguillot:
This PR adds a little bit more consistency with paddings and spacing within the modal. I've removed the unnecessary total selected since the footer checkbox already provided the same information.

The buttons have been updated to be more inline with Gutenberg Placeholder block/component. 

__Before__
before placeholder

before modal

__After__
after placeholder

after modal


maxschmeling (approved):
Thanks for the contribution @thomasguillot !

PR #645, opened 2025-09-01
Update contributors list in readme.txt

pkevan:
Fixes import issues for plugin directory.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -1,5 +1,5 @@
 === Remote Data Blocks ===
-Contributors: chriszarate, maxschmeling, mhsdef and others.
+Contributors: czarate, maxschmeling, hew.
 Tested up to: 6.8
 Stable tag: 1.1.0
 License: GPLv2 or later


shekharnwagh (approved):

readme.txt
```diff
@@ -1,5 +1,5 @@
 === Remote Data Blocks ===
-Contributors: chriszarate, maxschmeling, mhsdef and others.
+Contributors: czarate, maxschmeling, hew.
```
ingeniumed, line 2: Shouldn't this include the other members of the team as well?

acicovic: Not the PR author, but I think the purpose of this PR is just to fix the list as it exists today. This is a project with many contributors, so the question is how many of them get added here, and which are the criteria of addition. I think that having a common convention about this across our plugins could be useful.

pkevan: > Shouldn't this include the other members of the team as well?

It could do, but like @acicovic says, it was to fix an existing problem (.org usernames were invalid). Most probably we want to set some criteria for inclusion into this block (50+ commits or notable contributions?).

maxschmeling: Yeah, fixing the *issue* is good. Would like to determine the "process" for this. Anyone have context on how plugins typically do it? Should it be the *active* contributors? All time contributors with a certain criteria like @pkevan suggested?

maxschmeling: https://github.com/WordPress/gutenberg/blob/2646370b28ffaf1a7048f688cbe13491ed6e0379/readme.txt

This is the gutenberg contributors list
```
Contributors: matveb, joen, karmatosed
```

pkevan: > Yeah, fixing the _issue_ is good. Would like to determine the "process" for this. Anyone have context on how plugins typically do it? Should it be the _active_ contributors? All time contributors with a certain criteria like @pkevan suggested?

Can be however we want, but usually it just sticks the same like Gutenberg since no-one really has a process.

PR #646, opened 2025-09-04
Fix User-Agent header being overwritten by default

maxschmeling:
The `User-Agent` middleware will always set the default `User-Agent` value, but if the user provides a value that they want to use, we shouldn't overwrite that.

Also, this adds the actual plugin version to the default `User-Agent` string instead of `1.0`.

### DIFF
--- a/tests/inc/HttpClient/HttpClientTest.php
+++ b/tests/inc/HttpClient/HttpClientTest.php
@@ -580,22 +580,5 @@ class HttpClientTest extends TestCase {
 		// Should preserve custom User-Agent header
 		$this->assertTrue( $result->hasHeader( 'User-Agent' ) );
 		$this->assertSame( 'CustomApp/2.0', $result->getHeaderLine( 'User-Agent' ) );
-		$this->assertNotSame( 'WordPress Remote Data Blocks/1a test.0', $result->getHeaderLine( 'User-Agent' ) );
-	}
-
-	public function testProvideDefaultUserAgentWithEmptyUserAgent(): void {
-		// Test that provideDefaultUserAgent sets default when User-Agent is empty
-		$http_client = HttpClient::instance();
-
-		// Create a request with empty User-Agent header
-		$request = new \GuzzleHttp\Psr7\Request( 'GET', '/test', [ 'User-Agent' => '' ] );
-
-		// Apply the function
-		$result = $http_client->provide_default_user_agent( $request );
-
-		// Should set default User-Agent (empty string is considered as having a header)
-		$this->assertTrue( $result->hasHeader( 'User-Agent' ) );
-		$this->assertSame( '', $result->getHeaderLine( 'User-Agent' ) );
-		$this->assertNotSame( 'WordPress Remote Data Blocks/1a test.0', $result->getHeaderLine( 'User-Agent' ) );
 	}
 }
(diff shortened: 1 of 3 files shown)


tests/inc/HttpClient/HttpClientTest.php
```diff
@@ -551,4 +551,51 @@ public function testRepeatedErrorsWithSameCacheKeyResultInCacheHit(): void {
...
+		// Create a request with custom User-Agent header
+		$request = new \GuzzleHttp\Psr7\Request( 'GET', '/test', [ 'User-Agent' => 'CustomApp/2.0' ] );
+
+		// Apply the function
+		$result = $http_client->provide_default_user_agent( $request );
+
+		// Should preserve custom User-Agent header
+		$this->assertTrue( $result->hasHeader( 'User-Agent' ) );
+		$this->assertSame( 'CustomApp/2.0', $result->getHeaderLine( 'User-Agent' ) );
+		$this->assertNotSame( 'WordPress Remote Data Blocks/1a test.0', $result->getHeaderLine( 'User-Agent' ) );
```
chriszarate: assertNotSame is a bit silly

tests/inc/HttpClient/HttpClientTest.php
```diff
@@ -551,4 +551,51 @@ public function testRepeatedErrorsWithSameCacheKeyResultInCacheHit(): void {
...
+		// Create a request with empty User-Agent header
+		$request = new \GuzzleHttp\Psr7\Request( 'GET', '/test', [ 'User-Agent' => '' ] );
+
+		// Apply the function
+		$result = $http_client->provide_default_user_agent( $request );
+
+		// Should set default User-Agent (empty string is considered as having a header)
+		$this->assertTrue( $result->hasHeader( 'User-Agent' ) );
+		$this->assertSame( '', $result->getHeaderLine( 'User-Agent' ) );
+		$this->assertNotSame( 'WordPress Remote Data Blocks/1a test.0', $result->getHeaderLine( 'User-Agent' ) );
```
chriszarate: silly x2

tests/inc/HttpClient/HttpClientTest.php
```diff
@@ -551,4 +551,51 @@ public function testRepeatedErrorsWithSameCacheKeyResultInCacheHit(): void {
...
+		$this->assertSame( 'CustomApp/2.0', $result->getHeaderLine( 'User-Agent' ) );
+		$this->assertNotSame( 'WordPress Remote Data Blocks/1a test.0', $result->getHeaderLine( 'User-Agent' ) );
+	}
+
+	public function testProvideDefaultUserAgentWithEmptyUserAgent(): void {
+		// Test that provideDefaultUserAgent sets default when User-Agent is empty
+		$http_client = HttpClient::instance();
+
+		// Create a request with empty User-Agent header
+		$request = new \GuzzleHttp\Psr7\Request( 'GET', '/test', [ 'User-Agent' => '' ] );
```
chriszarate: This isn't how we run our requests, so it's not great coverage in that respect. But I guess it does unit test the public method.

maxschmeling: I struggled to get a real test to actually work and didn't want to keep spending time and decided just to test the function.

tests/inc/HttpClient/HttpClientTest.php
maxschmeling: Ope, this was left over from earlier test iteration that I removed but didn't get this line. Removed in ab9389ff

tests/inc/HttpClient/HttpClientTest.php
maxschmeling: ab9389ff

chriszarate (approved):

PR #654, opened 2025-09-24
Implement `getValues()` and supply preview index via custom store

chriszarate:
- Implement `getValues()` to resolve binding values from our context. This avoids the need to "merge attributes" in the `withBlockBinding` HOC.
- Supply preview index via custom Redux store, since that cannot be resolved from context or the block's binding.


maxschmeling:
This breaks the bindings for the pattern selection like we had seen before. I thought the implementation I had wasn't having that issue anymore last I had checked, but I will have to test again and see if that's true. I tested on `trunk` just to confirm it was  working there, and it is.

CleanShot 2025-09-25 at 01 30 46@2x

maxschmeling (approved):
Works well. I was trying to think of a better name than `previewValue` since it's there with `previewId` but it feels like maybe the best name for that property.

PR #661, opened 2025-10-01
Update permission callback to check user capability

pkevan:
Make sure the user has permission to edit the corresponding post.


inc/REST/RemoteDataController.php
```diff
@@ -79,6 +79,6 @@ public static function execute_queries( WP_REST_Request $request ): array|WP_Err
 	}
 
 	public static function permission_callback(): bool {
-		return true;
+		return current_user_can( 'edit_post' );
```
chriszarate: @pkevan Do you think we need to pass the current post ID here? This would require adding it to the payload.

chriszarate: This doesn't improve the security stance, but might account for custom caps that only apply to certain post types or posts.

pkevan: > @pkevan Do you think we need to pass the current post ID here? This would require adding it to the payload.

🤔 how were you thinking of supplying this?

chriszarate: By adding it to the payload here:
https://github.com/Automattic/remote-data-blocks/blob/23e571292a639043ea71e03bc142c10aa9b0420b/src/blocks/remote-data-container/hooks/useRemoteData.ts#L193

chriszarate: Can use this selector to get it from the store:
https://developer.wordpress.org/block-editor/reference-guides/data/data-core-editor/#getcurrentpostid

pkevan:
How does this look @chriszarate?

src/blocks/remote-data-container/hooks/useRemoteData.ts
```diff
@@ -194,6 +199,7 @@ export function useRemoteData( {
 			block_name: blockName,
 			query_key: queryKey,
 			query_inputs: inputs,
+			post_id: postId ?? null,
```
chriszarate: Very minor but appreciate keeping things in alpha order when that is the current state
```suggestion
			post_id: postId ?? null,
			query_key: queryKey,
			query_inputs: inputs,
```

types/remote-data.d.ts
```diff
@@ -97,6 +97,7 @@ interface RemoteDataApiRequest {
 	block_name: string;
 	query_inputs: RemoteDataQueryInput[];
 	query_key: string;
+	post_id: ?( number | string );
```
chriszarate: This is the published type in the docs, but the package types indicate `number | undefined`, which we modify to `number | null`.

```suggestion
	post_id: number | null;
	query_inputs: RemoteDataQueryInput[];
	query_key: string;
```

pkevan: hmm - confusing, must have done it based on the docs.

pkevan: ah, it fails: https://github.com/Automattic/remote-data-blocks/actions/runs/18285326911/job/52058739549

chriszarate: My fault, in that case change to `number | string | null`

inc/REST/RemoteDataController.php
```diff
@@ -78,7 +78,8 @@ public static function execute_queries( WP_REST_Request $request ): array|WP_Err
 		);
 	}
 
-	public static function permission_callback(): bool {
-		return true;
+	public static function permission_callback( WP_REST_Request $request ): bool {
+		$post_id = $request->get_param( 'post_id' );
```
ingeniumed: Should the `post_id` value be santized, before it's sent off to `current_user_can`?

pkevan: Maybe, although it would have to be a generic filter

maxschmeling:
```suggestion
		$post_id = $request->get_param( 'post_id' );
		
		if ( ! is_numeric( $post_id ) || intval( $post_id ) < 0 ) {
			return new WP_Error( 'invalid_post_id', 'Invalid post ID', array( 'status' => 400 ) );
		}
```

maxschmeling (changes requested):

inc/REST/RemoteDataController.php
pkevan: Tidied this up slightly, wasn't sure if `400` or `404` was appropriate, but there is varying implementations of this!

chriszarate (approved):

maxschmeling (approved):

PR #688, opened 2025-11-25
Fix duplicate link in readme.txt

pkevan:
Removed duplicate link to launch in WordPress Playground.


shekharnwagh (approved):
Thank you for fixing this. I don't have strong preferences, but I think the badge might be more noticeable than a regular link. What do you think?

pkevan:
> Thank you for fixing this. I don't have strong preferences, but I think the badge might be more noticeable than a regular link. What do you think?

Ah @shekharnwagh, I should have mentioned that this is only for wp.org plugin directory, see: https://wordpress.org/plugins/remote-data-blocks/#description. This isn't used on github.

shekharnwagh:
> > Thank you for fixing this. I don't have strong preferences, but I think the badge might be more noticeable than a regular link. What do you think?
> 
> Ah @shekharnwagh, I should have mentioned that this is only for wp.org plugin directory, see: https://wordpress.org/plugins/remote-data-blocks/#description. This isn't used on github.

Oh, didn't notice the `.txt` extension. I kinda assumed its `README.md` file.

PR #712, opened 2026-01-29
Minor enhancements for better UI

Utsav-Ladani:
## Summary

I am build a plugin for the **Newspack** which extends RDB's capabilities and provide domain centric user experience. I would like to propose following enhancements to plugin for building perfect UI/UX.

| Enhancements | Before | After
|--------|--------|--------|
| Hide label on frontend if value is empty | Screenshot 2026-01-29 at 11 58 03 AM | Screenshot 2026-01-29 at 11 58 49 AM |
| ~Remove~ Improve escaping to allow usage of SVGs in RDB blocks | Screenshot 2026-01-29 at 12 07 02 PM | Screenshot 2026-01-29 at 12 04 31 PM |
| Allow to fetch all sheet columns instead of just A to Z | N/A | N/A |


Utsav-Ladani:
CC: @kleinmatic

Utsav-Ladani:
Instead of removing escaping, improve it to render SVG related HTML tags and attributes.

CC: @pkevan @maxschmeling

pkevan:
Thanks for the PR @Utsav-Ladani - they look to be 3 distinct changes, so probably better to have them as such, rather than being together.

On the SVG element - that really isn't going to be supported in the plugin itself (following core WP's approach), so we'd need to re-think that part. Potentially we could add the output as a filter through `wp_kses`, which is defaulted to `post` which would allow a site to override at that point.

Utsav-Ladani:
Thanks @pkevan for the clarification. I have raised two separate PRs for each of the tiny changes:

- #720
- #721

For SVG elements, I found that RDB's render functions are sanitizing themselves as well as all inner blocks, which is causing the issue. Modifying `wp_kses` function's behavior using filter will change it for all places and may conflict with other plugins.

What do you think about moving the escaping part to the `get_value()` method? This is because data fetched from remote sources may be unsafe and only needs to be sanitized while inner blocks are already sanitized by their own render function.

Utsav-Ladani:
Thanks @pkevan for approving above two PRs.

I have pushed a different approach to sanitize labels and values earlier while fetching them instead of sanitizing whole output. Let me know your thoughts about this approach.

Utsav-Ladani:
Closing this as separate PRs are raised for each enhancement

PR #800, opened 2026-08-18
Configure request headers used in cache keys

maxschmeling:
## Summary
- add query-level `cache_key_request_headers` configuration for HTTP queries
- always retain `Authorization` and `Cache-Control`, merging additional query header names case-insensitively
- make `get_cache_key_request_headers()` part of `HttpQueryInterface` and have `QueryRunner` consume the query-owned policy directly
- carry the resolved list in one internal request header used consistently by cache lookup, storage, and logging
- strip the internal metadata before outbound dispatch and before storing the request in the cache entry
- configure Shopify’s built-in queries to include the storefront access-token header
- document the cache-isolation security risk, query-only configuration, and the direct-interface implementation requirement

## Security
Remote Data Blocks does not include every request header in cache keys. If a custom header changes authentication, authorization, tenancy, or returned data, every query that uses it must list that header in `cache_key_request_headers`; otherwise requests with different security contexts can share a cached response.

Headers configured on a data source are not added automatically. The documentation calls this out prominently, including the cross-request and cross-user risk with persistent object caching. The built-in `Authorization` and `Cache-Control` defaults cannot be removed.

## Compatibility
Direct implementations of `HttpQueryInterface` must implement `get_cache_key_request_headers()`. Returning `[]` uses only the built-in defaults. The extending guide includes an upgrade example.

## Testing
- `npm run format:check`
- `npm run lint:php`
- `npm run lint:js`
- `npm run check-types`
- `npm test` (596 PHP tests / 1,374 assertions; 76 JavaScript tests)
- `npm run build`
- `git diff --check`


inc/Config/DataSource/HttpDataSource.php
```diff
@@ -12,7 +13,11 @@
  *
  * Implements the HttpDataSourceInterface to define a generic HTTP data source.
  */
-class HttpDataSource extends ArraySerializable implements HttpDataSourceInterface {
+class HttpDataSource extends ArraySerializable implements HttpDataSourceInterface, CacheKeyRequestHeadersInterface {
+	public function get_cache_key_request_headers(): array {
```
chriszarate: I'm not sure I agree that a data source should be able to add request headers to the cache key. Data sources can define (base) request headers, but control is ultimately held by the query.

More importantly, caching is a behavior tied directly to queries and not to data sources. Cache TTL is another example of behavior that is only configurable at the query level.

inc/Config/CacheKeyRequestHeadersInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php declare(strict_types = 1);
+
+namespace RemoteDataBlocks\Config;
+
+interface CacheKeyRequestHeadersInterface {
```
chriszarate: If control is consolidated on queries, this interface can merge into `HttpQueryInterface`

inc/HttpClient/RdbCacheStrategy.php
chriszarate: Unmerged headers are being passed to the parent constructor here. The `varyHeaders` are currently only used in the overridden `getCackeKey` method, but it carries future risk of cache poisoning if that changes.

inc/Config/DataSource/HttpDataSource.php
maxschmeling: It felt to me like a safety net to make sure you can't leave it off of a query. But I'm ok with removing this for now and reconsidering it down the road if requested.

inc/Config/CacheKeyRequestHeadersInterface.php
maxschmeling: Merged `get_cache_key_request_headers()` into `HttpQueryInterface` and deleted the standalone interface. `QueryRunner` now consumes the method directly. The extending guide also documents the required method for direct interface implementations and that returning `[]` keeps the built-in defaults.

inc/HttpClient/RdbCacheStrategy.php
maxschmeling: Removed the separate `KeyValueHttpHeader` list from the parent constructor. The parent now receives only storage and TTL, leaving the per-request merged query list as the single cache-key source.

inc/Config/DataSource/HttpDataSource.php
maxschmeling: Removed the data-source-level configuration entirely, including the schema/getter, Generic HTTP UI and mapping, and data-source documentation. Cache-key header policy is now query-only. Shopify declares its token header explicitly on each built-in query.

inc/HttpClient/RdbCacheMiddleware.php
```diff
@@ -2,7 +2,31 @@
 
 namespace RemoteDataBlocks\HttpClient;
 
+use Psr\Http\Message\RequestInterface;
+
 class RdbCacheMiddleware extends \Kevinrob\GuzzleCache\CacheMiddleware {
+	public const CACHE_KEY_REQUEST_HEADERS_HEADER = 'X-Remote-Data-Blocks-Cache-Key-Headers';
+	public const CACHE_KEY_REQUEST_HEADERS_OPTION = 'remote_data_blocks_cache_key_request_headers';
+
+	public function __invoke( callable $handler ): callable {
```
chriszarate, line 8: I lose the plot a bit here. The underlying `GreedyCacheStrategy` uses a request header to set cache TTL and suppresses it on the actual request and cache object. See: https://github.com/Kevinrob/guzzle-cache-middleware/blob/master/src/Strategy/GreedyCacheStrategy.php#L23

Why can't we do that and avoid this confusing juggling? Perhaps `getCacheObject` being `protected` is a stumbling block, but we can just provide our own implementation of `GreedyCacheStrategy`—it's not that big. Replicating that code is much simpler than these changes IMO.

maxschmeling: Done in https://github.com/Automattic/remote-data-blocks/pull/800/commits/dafd9197dc07f1604b7cde6f53545ac09fad6c8e

I don't really love this type of flow, but it's simple and consistent.

inc/Config/Query/HttpQuery.php
```diff
@@ -38,6 +39,18 @@ public function execute_batch( array $array_of_input_variables ): array|WP_Error
...
+	/**
+	 * Get the request header names whose values should be included in cache keys.
+	 *
+	 * @return array<string> Request header names included in cache keys.
+	 */
+	public function get_cache_key_request_headers(): array {
+		return CacheKeyRequestHeaders::merge(
+			CacheKeyRequestHeaders::DEFAULT_HEADERS,
+			$this->config['cache_key_request_headers'] ?? []
+		);
```
chriszarate: Merging already happens later in `QueryRunner`, so this is redundant. It also gives the appearance that the defaults can be overridden.


```suggestion
		return $this->config['cache_key_request_headers'] ?? [];
```

inc/HttpClient/CacheKeyRequestHeaders.php
```diff
@@ -0,0 +1,32 @@
+<?php declare(strict_types = 1);
+
+namespace RemoteDataBlocks\HttpClient;
+
+final class CacheKeyRequestHeaders {
+	public const DEFAULT_HEADERS = [ 'Authorization', 'Cache-Control' ];
```
chriszarate: The docs state:

> The built-in defaults cannot be removed.

Given that stance, this `const` should be private and `merge` should be updated to always include the defaults. That encapsulates the contract in this class instead of requiring external callers to fulfill it.

inc/HttpClient/CacheKeyRequestHeaders.php
```diff
@@ -0,0 +1,32 @@
...
+				if ( isset( $seen_headers[ $normalized_header ] ) ) {
+					continue;
+				}
+
+				$seen_headers[ $normalized_header ] = true;
+				$merged_headers[] = $header;
+			}
+		}
+
+		return $merged_headers;
```
chriszarate:
```suggestion
	 * @param array<string> $header_names Request header names.
	 * @return array<string> Merged request header names.
	 */
	public static function merge( array $header_names ): array {
		$merged_headers = self::DEFAULT_HEADERS; // copied on assignment
		$seen_headers = [];

		foreach ( $header_names as $header ) {
			$normalized_header = strtolower( $header );
			if ( isset( $seen_headers[ $normalized_header ] ) ) {
				continue;
			}

			$seen_headers[ $normalized_header ] = true;
			$merged_headers[] = $header;
		}

		return $merged_headers;
```

inc/HttpClient/RdbCacheMiddleware.php
```diff
@@ -2,7 +2,19 @@
 
 namespace RemoteDataBlocks\HttpClient;
 
+use Psr\Http\Message\RequestInterface;
+
 class RdbCacheMiddleware extends \Kevinrob\GuzzleCache\CacheMiddleware {
+	public const CACHE_KEY_REQUEST_HEADERS_HEADER = 'X-Remote-Data-Blocks-Cache-Key-Headers';
+
+	public function __invoke( callable $handler ): callable {
```
chriszarate, line 8: I'm not sure I understand what is accomplished by this `__invoke` override. I believe this header is already removed by the `RdbCacheStrategy#getCacheObject`

inc/HttpClient/RdbCacheStrategy.php
```diff
@@ -19,31 +19,30 @@ class RdbCacheStrategy extends GreedyCacheStrategy {
...
 	private const FALLBACK_CACHE_TTL_IN_SECONDS = 300; // 5 minutes for success responses
 
 	public function __construct( ?CacheStorageInterface $storage = null ) {
-		// Filter this if customization is needed.
-		$vary_headers = new KeyValueHttpHeader( self::CACHE_INVALIDATING_REQUEST_HEADERS );
-
 		parent::__construct(
 			$storage ?? new WordPressObjectCacheStorage( self::WP_OBJECT_CACHE_GROUP ),
-			self::FALLBACK_CACHE_TTL_IN_SECONDS,
-			$vary_headers
```
chriszarate, line 33: Cache poisoning is still possible in the future. The vary headers passed to the constructor  (`[]`) and the ones used in `get_object_cache_key_from_request` are different. That is why I suggested porting the code from `GreedyCacheStrategy` and no longer extending it. I don't think we can honor its contract.

inc/HttpClient/RdbCacheMiddleware.php
```diff
@@ -2,7 +2,19 @@
 
 namespace RemoteDataBlocks\HttpClient;
 
+use Psr\Http\Message\RequestInterface;
+
 class RdbCacheMiddleware extends \Kevinrob\GuzzleCache\CacheMiddleware {
+	public const CACHE_KEY_REQUEST_HEADERS_HEADER = 'X-Remote-Data-Blocks-Cache-Key-Headers';
```
chriszarate: `RdbCacheStrategy` is probably a better home for this. That's where `CACHE_TTL_REQUEST_HEADER` lives also

inc/HttpClient/RdbCacheMiddleware.php
maxschmeling: Left over duplicate. I've removed it.

inc/HttpClient/CacheKeyRequestHeaders.php
maxschmeling: Updated in a slightly different way https://github.com/Automattic/remote-data-blocks/pull/800/commits/44f1b259ceb7e9aa1a564b83c5015659bb1ca2ca

inc/HttpClient/CacheKeyRequestHeaders.php
maxschmeling: I've made the `const` private and updated the `merge` in a slightly different way. It is shorter, with the main difference that it doesn't maintain the header casing, which I think is fine since if there are differently cased entries it would only keep one anyways.

https://github.com/Automattic/remote-data-blocks/pull/800/commits/44f1b259ceb7e9aa1a564b83c5015659bb1ca2ca

inc/HttpClient/RdbCacheMiddleware.php
maxschmeling: Moved

inc/HttpClient/RdbCacheMiddleware.php
chriszarate: The `__invoke` is still present

inc/HttpClient/RdbLogMiddleware.php
```diff
@@ -27,8 +27,8 @@ public function __invoke( callable $handler ): callable {
 		return function ( RequestInterface $request, array &$options ) use ( $handler ): PromiseInterface {
 			return $handler( $request, $options )
 				->then(
-					$this->handle_success( $request, $options ),
-					$this->handle_failure( $request, $options )
+					$this->handle_success( $request ),
```
chriszarate: The edits in this file serve no purpose. Options might be logged in the future

maxschmeling: https://github.com/Automattic/remote-data-blocks/pull/800/commits/52bc96e60834622f5c622006518c76939c911cee

inc/HttpClient/RdbCacheStrategy.php
maxschmeling: Yeah, you're right, it's fine now but not safe for the future.

Copied the implementation here https://github.com/Automattic/remote-data-blocks/pull/800/commits/1cde10a06529e482de9fc334e3db05e4461ed2ca

Small tweak here: https://github.com/Automattic/remote-data-blocks/pull/800/commits/9ae57316ad368f1e2c043d9edd59f06bb1ee153b

Testing and checking now

inc/HttpClient/RdbCacheMiddleware.php
chriszarate: To be clear, I don't think this `__invoke` has any effect. Can you test this?

maxschmeling: It doesn't. I removed it but didn't get that in the commit I guess. Pushed in 78741db7

inc/Config/Query/CacheKeyRequestHeadersAwareInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php declare(strict_types = 1);
+
+namespace RemoteDataBlocks\Config\Query;
+
+/**
+ * Implemented by HTTP queries that add request headers to their cache keys.
+ */
+interface CacheKeyRequestHeadersAwareInterface {
```
chriszarate, line 8: Curious why this is a separate interface instead of part of `HttpQueryInterface`. Are we worried about backward compatibility? I don't think I see the issue.

maxschmeling: Yes, it's a backwards compatibility thing. I'm sure it's not that big of a deal and probably wouldn't cause anybody any issues, but Codex added it and it seemed appropriate enough to leave.

chriszarate: Ok, I guess if we added it to `HttpQueryInterface` and an integration implemented `HttpQueryInterface` without extending `HttpQuery`, it could raise a fatal. Breaking API change.

inc/HttpClient/RdbCacheMiddleware.php
chriszarate: @maxschmeling it's baaaaaaack

maxschmeling: This removal actually does cause a test failure.

`get_cache_object()` removes the header from the request stored in the cache entry, but it runs only after the downstream HTTP handler has returned. This override wraps that handler so the cache strategy can still use the metadata while the transport receives a sanitized request. Without it, the metadata header is sent to the remote API, as covered by `testCacheKeyRequestHeaderMetadataIsNotSentToRequestHandler`.

maxschmeling: I also realized this means the TTL value isn't being stripped either. So I've added a test and fix for that as well.

https://github.com/Automattic/remote-data-blocks/pull/800/commits/fea0af2e42dbc8d139a41ca34d3b579db6aa7cfe

chriszarate: > Without it, the metadata header is sent to the remote API

The same is probably true of `X-Remote-Data-Blocks-Cache-TTL`. Is that bad?

chriszarate: I'm not sure it is a bad side effect. It's just a request header. But we should pick a policy and be consistent. I'd vote to leave it personally.

maxschmeling: Yeah, I already removed it in the commit I commented above.

I definitely don't think we should be adding headers to requests for internal functions and letting them flow through to the endpoint. Unless you have strong opinions about it, I'm going to go with the approach of stripping them.

chriszarate (approved):

